PR #321, opened 2018-05-24
Fixes the problem where on refresh, the current highlighted nav tab w…

dstopp:
This addresses  the problem where, on refresh, loses the current selected tab 'focus' and puts it back to 'Search', even tho the contents of the page rendered are not 'Search'.


rfrenchseti:
Merged into new_import branch.

PR #342, opened 2018-07-02
fixes clicking on planet wasn't opening corresponding targets group

basilleaf:
Hello I noticed the targets were no longer opening when a planet is clicked so here is a fix for that.

### DIFF
--- a/static_media/js/widgets.js
+++ b/static_media/js/widgets.js
@@ -243,8 +243,6 @@ var o_widgets = {
 
     // this is called after a widge is drawn
     customWidgetBehaviors: function(slug) {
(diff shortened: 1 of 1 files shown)


rfrenchseti:
Thanks! I hadn't noticed that. We have a major re-deploy happening in a few weeks so it will get added then.

rfrenchseti:
Doesn't this change break the other switch cases? They use mult_id as well but now it won't be defined.

basilleaf:
oops! ok i updated the PR I think

rfrenchseti:
Integrated with new_import branch. Also swapped meaning of -/+ symbols, which were backwards.

PR #371, opened 2018-07-12
Api guide fix

basilleaf:
this is more or less a copyedit of the [API guide page](https://tools.pds-rings.seti.org/opus/api/) : 

- commented out 2 examples for having broken links, but I think they were rather obscure examples so not sure it's worth filing an issue for them.

- changes any html examples to json instead because it looks like in html endpoints the link urls are not getting the base path, so they aren't displaying in the browser, but pretty sure no one (including opus) uses those endpoints (except maybe me in some past keynote slide!)


rfrenchseti:
Incorporated into new_import branch

PR #398, opened 2018-07-24
Bug 394 - RF attempt

rfrenchseti:


rfrenchseti:
Postponing until new gallery view.

PR #448, opened 2018-08-28
bug#308 + removed refs to unused widgets2

dstopp:
also removed dead code and updated loops to use $.each


rfrenchseti:
Doesn't fix the q-types for string widgets.

rfrenchseti:
This will be dealt with in the new UI code.

PR #504, opened 2018-11-28
Test branch

juzen2003:
test_api & test_db_data folders added.
test_api: contains files for api testing against app/live-sites.
test_db_data: for now it doesn't have any test, but would add db integrity test, like testing sentinel values in the future here.
All tests related to each app should still stay in its app folder. (standard)


rfrenchseti (changes requested):
**manage.py:**

* What is the purpose of the "app" argument to manage.py? To me testing the "app" means testing the entire OPUS application, but it seems here to mean testing only the API. Since people will normally use "manage.py test" to run tests, it seems like something more specific should be used instead at this top level.

**TEST_API_README.md:**

* Instead of putting the pip install here, djangorestframework should be added to the top-level requirements files.
* A lot of the other notes seem to be personal notes - you might want to move these somewhere other than the repo.

**api_for_test_cases:**

* I don't understand the name of this directory. Should it be "test_cases_for_api" instead?
* More importantly, I'm confused about the whole "api_for_test_cases" package. Why does this exist at all instead of just putting this code in each of the test_*.py files? It seems like each module in this directory is only used in one test case file. Is that correct?

**opus/application/test_api/api_for_test_cases/api_return_formats.py:**

* Since we're using Python 3.6+, there's a much cooler way to put strings together. Instead of saying

> "metadata/%s." %(opus_id)

instead say:

> f"metadata/{opus_id}."

**opus/application/test_api/api_for_test_cases/api_vims_image_downlinks.py:**

* Do you really have to say:

> from . import api_return_formats

instead of just

> import api_return_formats

**test_db_integrity.py:**

* Go ahead and copy the db integrity tests from search/tests.py here so we have something to test.

rfrenchseti (changes requested):
One more thing that should be done - In the code that checks the result status for various formats, you should try both the "api" and the "__api" versions, assuming they both exist. We may want to support different return possibilities for internal and external. For example internal might be able to return HTML, while external might be able to return only JSON.

juzen2003:
**manage.py:**

Specific usage added:
* against internal database:
    * python manage.py test -s -v 2 api-internal-db
* against production site:
    * python manage.py test -s -v 2 api-livetest-pro
* against dev site:
    * python manage.py test -s -v 2 api-livetest-dev
* if other api.* commands is inputed, above three options would get printed out without running anything.

**TEST_API_README.md:**

* Only available commands are there now
* requests==2.20.1 and djangorestframework==3.9.0 are added in requirements*.txt since they are required for api tests

**api_for_test_cases:**

* Original intention for these files is that later on we can keep re-using the same api urls & payloads for other tests, or creating a subclass of ApiFormats for a more specific test from api_return_formats.py, just like how I did for VIMS test.
* Now they are combined into their related test_*.py, and ApiForVimsDownlinks is a subclass of ApiFormats which serves the same purpose.
 
**opus/application/test_api/api_for_test_cases/api_return_formats.py:**

* All string interpolations are changed to this cooler way.

**opus/application/test_api/api_for_test_cases/api_vims_image_downlinks.py:**

* Originally if I don't do:
`from . import api_return_formats`
instead just do:
`import api_return_formats`
It would complained about no such module for api_return_formats even if they are under the same directory in the same package. Now after re-org, this import is removed.

**test_db_integrity.py:**

* db integrity tests from search/tests.py is copied over for testing.

**Notes for __api & api:**

* When running test against internal db, test would run with both __api & api urls.
* When running against live sites, it would only run with api urls since we only release api urls in document. I did test __api urls against live sites, and the return format and test results are the same as api_urls. Both __api & api urls would render the same result because in urls.py for different apps in django, the regex of url pattern for __api & api are the same except __, their corresponding views (function) inside views.py are the same as well.

rfrenchseti:
I made a few minor code cleanups:

* Instead of re.match("api.*", command) I used the much faster and easier command.startswith("api").
* There's no point in saying print(f"{usage}") instead of use print(usage)
* I changed test_return_formats.py to disable logging through WARN instead of CRITICAL so we still see database configuration errors (like the missing test_dictionary).
* Instead of saying "not v1_count > 0" you can just say "not v1_count" since v1_count can't be negative.
* I removed the (now redundant) database integrity checks from search/tests.py and I removed the setup/teardown from test_db_integrity.py since these tests don't use the user_searches or cache tables.

rfrenchseti (approved):

rfrenchseti:
Also combined lines:

from rest_framework.test import APIClient, CoreAPIClient, RequestsClient

PR #520, opened 2018-12-12
Unit branch

juzen2003:
# Modification:
# apps/results/views.py
* Update get_page function to display rounded off values in mini detailed page.
* manual and manage.py test have been run to make sure nothing is broken.

### DIFF
--- a/opus/application/apps/results/views.py
+++ b/opus/application/apps/results/views.py
@@ -1002,7 +1002,8 @@ def get_page(request, use_collections=None, collections_page=None, page=None,
     # if pi_form_type has format, we format the results
     for idx, form_type in enumerate(form_lists):
         for entry in results:
-            if form_type and entry[idx]:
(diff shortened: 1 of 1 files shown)


rfrenchseti:
I made a couple of very minor changes:
- Renamed form_lists to form_type_formats and form_type to form_type_format for better clarity.
- Once again used abs() around the value so we properly handle negative numbers.
- Used () in if statement to avoid use of \ as line terminator (I think it looks cleaner)
- The comparison against THRESHOLD_FOR_EXPONENTIAL etc was getting redundant so I made a new function in tools/app_utils.py called format_metadata_number that does all this and called it from each place. This also allowed me to consolidate exception checking in one place.

rfrenchseti:
Also added formatting to _get_metadata_by_slugs so the "Metadata for Current Columns" section of the Details page is also properly formatted.

Issue #303, opened 2018-04-10
Don't allow selection of more than one surface geo target


rfrenchseti:
If you select the Surface Geometry tab and click on "target name" and then select more than one target, only the first target shows up as a new menu with surface geometry options.

rfrenchseti:
We will never support more than one surface geometry target. Thus the selections should be radio buttons, not checkboxes.

rfrenchseti:
Reopening this because it has a problem when you refresh your browser window. For all sub-menus other than the one with the valid selection, it gives the message "Select a valid choice. ["XXX"] is not one of the available choices."

Honestly I thought I had fixed this once before in the surface geometry target widget. I think the trick is to not set a choice when creating the form widget if the given search option isn't one of the available choices.

juzen2003:
So the issue is related to how django form verifies the data when there is one exists. The issue is there at the production site too on Intended Target Name. By changing MultipleChoiceField or ChoiceField to CharField and directly passing in choices in widget would fix the issue.

rfrenchseti:
Fixed

PR #525, opened 2018-12-14
Test branch

juzen2003:
# Modifications related to Issue #303:
# search/forms.py
  * Use RadioSelect widget when we are selecting surfacegeometry target so that only one surfacegeometry target can be selected at a time.
# widgets.js
  * Attached event handler to radio button selection for surfacegeometry target 
    * Modify the logic a little bit so that only one value is passed in for surfacegeometrytargetname when running api call on /opus/__api/meta/result_count.json (previously multiple target names can be assigned to the slug and it would fail to query the database since only one is allowed)
  * Add another on change event handler to close all opened surfacegeometry widgets when user changes to another surfacegeo target (previously we have to manually close all related surfacegeometry widgets)
# tools/app_utils.py
  * Add a check to make sure val is not None before passing it into abs() so that we don't get error sometimes when val is None

### DIFF
--- a/opus/application/apps/search/forms.py
+++ b/opus/application/apps/search/forms.py
@@ -150,12 +150,20 @@ class SearchForm(forms.Form):
                 else:
                     choices = [(mult.label, mult.label) for mult in model.objects.filter(display='Y').order_by('disp_order')]
 
-                self.fields[slug] = forms.MultipleChoiceField(
-                        # label = ParamInfo.objects.get(slug=slug).label,
-                        label = '',
-                        choices = choices,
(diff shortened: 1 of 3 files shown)


dstopp (changes requested):
in widgets.js; line 97 - more appropriate would be to use a push to add elements to the end of an array.
ie., values.push(value);

juzen2003:
Change is made based on Debby's review. Using .push to append values to the end of the array instead of indexing by the array length every time to have a faster operation.

dstopp (approved):

Issue #524, opened 2018-12-14
Clean up Django HTTP caching


rfrenchseti:
We need to investigate how Django does HTTP-level caching (of API results, CSS files, and JS files). Currently some API functions are marked as @never_cache (like /api/images) for no obvious reason. We need to standardize which API calls we allow to be cached as well as how long we want things to be cached in the browser.

rfrenchseti:
This has all been dealt with. JS and CSS files have version numbers that make sure caching doesn't happen across releases. Pretty much every API call is marked @never_cache so that back end updates take effect immediately, and so that everything makes it into the Apache log files.

PR #531, opened 2018-12-19
Add a version to all javascript src in script tag, the version is the latest git commit id

juzen2003:
# Modifications related to #524 
# /ui/templates/footer.html
  * use the latest git commit id inside the src attribute of script tag for .js files to make sure users are downloading the new .js files whenever files are updated.
# /tools/app_utils.py
  * add a function to get the latest git commit id
# ui/views.py
  * footer.html is included in base.html and base.html is render by class based view, add a never cache to the class based view. 
  * Update settings.LATEST_GIT_COMMIT_ID if it's an empty string (initial value), that way it would only get updated once when the program is run
# settings.py
  * Initialize LATEST_GIT_COMMIT_ID as empty string here.


rfrenchseti (changes requested):
- I think it would be a good idea to make a log entry (probably "warning" level) when get_latest_git_commit fails to get the commit ID.
- Let's call the global variable something like OPUS_JS_VERSION instead of LATEST_GIT_COMMIT_ID. That way in an environment that doesn't even support git we could just set the variable in settings to a version number and the variable name would still make sense.
- I don't understand why you're using "@method_decorator". That is used to make a normal decorator work on a class method. But you aren't using it on a class method, and we don't use Class-Based View anyway.

juzen2003:
* A log.warning is added whenever we can't get the latest commit id and if that happened, it would generate a random number for OPUS_JS_VERSION
* Variable changed to OPUS_JS_VERSION, and it would get a random number if git commit id is not available
* We used class-based view, main_site.as_view() to render base.html (which has footer.html included), to decorate the class-based view with never_cache, @method_decorator is used. "dispatch()" would get called after "as_view()". I was referring to the django document, is it a proper way to do it? (https://docs.djangoproject.com/en/2.1/topics/class-based-views/intro/)

juzen2003:
# /ui/templates/header.html
  * Add the version for main.css in link tag as well to force reloading css file whenever there is a new update.
  * Rename OPUS_JS_VERSION to OPUS_FILE_VERSION in multiple files.

Issue #253, opened 2018-02-12
All input fields need better validation and feedback to the user


rfrenchseti:
For example, entering "A" for a float or something bad for a time.

From Michael Aye <Michael.Aye@lasp.colorado.edu>
It’s okay, it works now, I had an illegal day number entered, 000.
Maybe the system should defend faster/better against that, it takes like half a minute to return after entering 2017-000T00:00:00 as t0.

rfrenchseti:
There are two places validation is done - in the Django Form when the user hits Enter (see search/forms.py) and in the slug decoder (see urlToSearchParams in search/views.py). The Form validation isn't doing anything particularly useful right now. The slug decoder throws an internal ValueError when the float or special decoder fails, and the resulting selection criterion is just ignored during the search.

rfrenchseti:
Implementation ideas:

- In order to handle this properly, we may have to get rid of the use of Django forms and implement our own directly in JavaScript.
- All data validation should be done in Python, since it has to be done there for other purposes anyway, and we don't want to duplicate code in Python and JavaScript.
- Whenever a user "exits" an input field, there should be an AJAX call to Django that validates the input. If the input is valid, it gets added to the hash. Otherwise, the Django will return a specific error message saying why the value is invalid, which will be displayed to the user in some obvious way.
- When a value is invalid it will not be added to the hash, so illegal values will never be sent to the backend.
- When a permalink comes in, each slug will be validated in the same way to prevent the permalink from continuing to have invalid values in it.

rfrenchseti:
New plan:
- There is a new API /__api/normalizeinput.json?slug1=val1&slug2=val2. For each slug, it validates the input value. If it is a valid value, it normalizes it (for example, puts date/time into the format YYYY-MM-DDTHH:MM:SS.sss) and returns it. If the value is not valid, it returns null.
- When the user types a character in a numerical input field, OPUS should call this back-end API on the current value. If the value is invalid, the input field should be bordered in red. If the value is valid, the input field should be bordered normally, or perhaps in green.
- When the user hits ENTER or changes focus away from the input field, a final call to normalizeinput should be made. If the value is invalid, a more dramatic way of indicating that the field is bad should be made (Thick red border? Shaded red background?). If the value is valid, it should be replaced with the normalized version.
- Under no circumstances should an actual search be performed or the hash updated when there is an invalid value in one of the search boxes. In fact we probably want to disable the "Search" tab entirely in this case.
[...]

PR #535, opened 2018-12-21
Input fields validation

juzen2003:
# Initial Implementation of input validation related to #253 
**Here is the behavior of input validation for range inputs:**
- When a user opened a range input widget and focused one of inputs:
  - Validation started, it would keep calling normalized api whenever there is an input change
  - Border color would change corresponding to the json return of api call, reqno is used to handle the async return of ajax calls.
  - When user finished up typing and the input value is valid with green border, either focused out of ALL range inputs or hit enter key while staying in a range input would start the search. Before starting the search:
    - A final call of normalized api would be run:
      - If any input is not valid, that input field would have red background, and no search would be performed
      - Else search would be performed, and all border change back to original gray color

- When a red background input existed:
    - If a valid value is inputted, the red background would be gone
    - or if it got focused in, the red background would be gone and only left with red border

**Notes**
- Not quite sure if current border / background colors looks ok for everyone, those can be updated easily by changing the variable at the top of search.js
- Event handlers for input.RANGE are added (input focusin, keypress) and re-written (change focusout).


rfrenchseti:
There's some very strange stuff going on here:

- Start with a brand new OPUS
- Activate Observation Duration
- Click on the min field and type "1"
- Click on the max field and type "2"
- Click on the "X" to kill Observation Duration. The hash is:

`http://127.0.0.1:8000/#/view=search&browse=gallery&colls_browse=gallery&page=1&gallery_data_viewer=true&limit=100&order=time1&cols=opusid,instrument,planet,target,time1,observationduration&widgets=planet,target&widgets2=&detail=`

as expected.

- Activate Observation Duration again
- Click on the min box
- The normalize call received is:

```
[20/Dec/2018 19:25:27] DEBUG [tools.app_utils:217] API 35 /opus/__api/normalizeinput.json {
    "browse": "gallery",
    "colls_browse": "gallery",
    "cols": "opusid,instrument,planet,target,time1,observationduration",
    "detail": "",
    "gallery_data_viewer": "true",
    "limit": "100",
    "observationduration1": "",
    "observationduration2": "2",
    "order": "time1",
    "page": "1",
    "reqno": "15",
    "view": "search",
    "widgets": "observationduration1,planet,target",
    "widgets2": ""
}
```

Notice that "observationduration2" already has a value. It has remembered the value from last time, even though it isn't in the hash!

There are some other weird cases where when you kill a widget, the values still stay in the hash.

rfrenchseti:
The current behavior is that any time you leave focus on any widget, it starts a search. I think the change to require you to leave ALL range widgets or hit enter is going to confuse people. It actually confused me for awhile until I read your comment above. I couldn't figure out why sometimes it had a red border and other times a red background, even though in both cases I had focus elsewhere.

As soon as focus leaves a range widget or you hit enter, the background needs to change to red if necessary, and a search needs to be invoked if everything is green.

rfrenchseti:
Actually you can get into a weird state where typing into a field keeps sending the same normalize request to the back end over and over - without the new value the user has typed. Type something like "1,2" (which is a valid number in our system). The API call just says "1," so something in the JS is parsing it badly. Then backspace over the "1,2" and start typing something else. You will keep getting the same "1," call to the API.

rfrenchseti:
Also, there's a big thing missing we had talked about. When you leave focus and it does the final normalize call, it needs to actually update the value in the input widget so the user sees it change. For example, if I type "1_2_3" it should change to "123" when I leave the focus.

rfrenchseti:
I noticing a couple of problems with the normalize API now that I can easily type values in it. I'll fix it and these are here for my reference:

- You can type a large number like "1e1234" and it will return "inf" as a valid value
- You can type "inf" or "nan" and it will actually parse it as a valid number

EDIT: I've fixed this now and checked the changes into master along with new tests.

rfrenchseti:
For future testing, I suggest you turn on API logging in OPUS. Then pull up each type of field:

- Numerical range
- Date/Time range
- Spacecraft Clock Count (Cassini, Voyager, Galileo)
- Cassini Orbit Number

Try typing in invalid and valid values, backspacing, changing from valid to invalid, etc. Numbers are OK to have "_", ",", and space in them.

Watch the API calls and see what the return values say. If the return value doesn't match what you're typing, there's a problem. I'm frequently seeing it get into a state where I can type whatever I like, and it just sends some old value over and over.

rfrenchseti (changes requested):
See various comments

dstopp (changes requested):
- Please move all css-related items into main.css; create appropriate classes and then turn the classes on as off as required.
- I would prefer you to use $.getJSON for your json requests.  I am not clear you you needed to create the function 'ajaxAPICall' for your promise, looks like if you call $.getJSON you can call direct.
- I believe you can replace your for loop w/$.each.
- I need to ponder your addition of on 'keypress' - my gut says not good plan, onblur, onchange, better but i cannot come up w/the appropriate suggestion just at the moment.
- Please move any new global vars that are specific to search.js into o_search namespace, naming them appropriately so there is some context.
- I cannot stress enough appropriate naming of functions and variables.  If 'performSearchOnRangeInput'
 is validating a field, perhaps a better name might be 'validateMyField' <filling in 'MyField' w/the appropriate thing>... maybe even validateRangeInput. 
- It's not noted anywhere, but i'm trying to have JS coding standards be camel case for functions and vars; 4 space tab.  Note that there is really no real consistency at the moment, it's more of a goal if possible.

rfrenchseti:
Search/views.py: There doesn’t need to be () around the if. Reqno is asssigned to twice. Do you really want the default for reqno to be ‘’? Wouldn’t None be better? It’s a number so having a default of a string seems bad.

rfrenchseti:
You and Debby should discuss whether to use ‘ or “ for strings in JS code. I think it’s currently using “. Yes I know that’s not the same as the Django...

I think you’re getting rid of it, but I don’t see the point of temp_hash. I think only the real hash matters, and it shouldn’t be updated until there’s a valid field contents and focus has moved or enter pressed. 

I haven’t tried it but make sure pasting triggers validation.

juzen2003:
### Modifications
* Normalized API is called whenever range input is changing, user focus out (hit enter), or close a widget.
* No search would be performed if any of range inputs has invalid value.
* Whenever user left focus or hit enter on any range input, if there is an invalid value, the input would have red background and the search tab would be disable. User couldn't add any more widgets until they fixed the current inputs.
* Normalized values would be displayed in range inputs if they were valid.
* js formats:
  * css related items are put in main.css and turned on/off by using addClass/removeClass
  * camel case & 4 sapce tab are used for newly added code
  * for loop is replaced with $.each
  * 'keypress' is removed
  * global variables are moved into related namespace
  * $.ajax is replaced with $.getJSON, except the one in opus.js for chaining two ajax calls

### Notes
* When user is in focus of a range input with an invalid value, and click a new tab in search to leave focus, the new widget might/might not open up before search tab is disabled, but no search would be performed. Please refer to the step 14 in the below test scenario. Do we want to add a normalized api call to check all data before opening a new widget?
* When pasting url with invalid values, the green label of result count would not show up (because result count api call would return 404 or 500 with invalid values). But after fixing values in all inputs, search would be performed and everything would be back to normal.


### Test Scenario
* With all range inputs, open multiple widgets and close some of them. Mix with valid and in-valid values
1. Open up "Observation Duration":
    * Type "1 2" in min (normalized api return json is correct when input is changing)
      * Hit enter, return 8477
        * Same return as of: "api/meta/result_count.json?observationduration1=1+2"
    * Type "5,100" in max (normalized api return json is correct when input is changing)
      * Hit enter, return 7895
        * Same return as of: "api/meta/result_count.json?observationduration1=1+2&observationduration2=5,100"
  2. Open up "Observation Time":
    * Type "2017-02-4" in max (normalized api return json is correct when input is changing)
      * Change focus to min, return 6971
        * Same as return of: "api/meta/result_count.json?observationduration1=1+2&observationduration2=5,100&timesec2=2017-02-4"
    * Type "1999-01-01" in min
      * Hit enter/leave focus, return 6540
        * Same as return of: "api/meta/result_count.json?observationduration1=1+2&observationduration2=5,100&timesec2=2017-02-4&timesec1=1999-01-01"
  3. On "Observation Time":
    * Change min to something invalid like "1999x"
      * Hit enter, min would have red background, and search tab would be disabled, and no search would be performed.
      * Close "Observation Time", return 7895 (search is performed and search tab is enabled)
        * Same as return of: "api/meta/result_count.json?observationduration1=1+2&observationduration2=5,100"
  4. Repeat step 2
  5. Open up "Mission":
    * Click "Cassini", return 6483
      * Same as return of: "api/meta/result_count.json?observationduration1=1+2&observationduration2=5,100&timesec2=2017-02-4&timesec1=1999-01-01&mission=Cassini"
  6. Click "Cassini Mission Constraints" and open up "Saturn Orbit Number (By Range)":
    * Type "3" in min
      * leave focus, return 2506
        * Same as return of: "api/meta/result_count.json?observationduration1=1+2&observationduration2=5,100&timesec2=2017-02-4&timesec1=1999-01-01&mission=Cassini&CASSINIrevnoint1=3"
    * Type "0_4" in max
      * leave focus, return 2118
        * Same as return of: "api/meta/result_count.json?observationduration1=1+2&observationduration2=5,100&timesec2=2017-02-4&timesec1=1999-01-01&mission=Cassini&CASSINIrevnoint1=3&CASSINIrevnoint2=0_4"
  7. Open up "Spacecraft Clock Count":
    * Type "1485958000" in min
      * leave focus, return 2112
        * Same as return of: "api/meta/result_count.json?observationduration1=1+2&observationduration2=5,100&timesec2=2017-02-4&timesec1=1999-01-01&mission=Cassini&CASSINIrevnoint1=3&CASSINIrevnoint2=0_4&CASSINIspacecraftclockcount1=1485958000"
    * Type "1489511000" in max
      * leave focus, return 2092
        * Same as return of: "api/meta/result_count.json?observationduration1=1+2&observationduration2=5,100&timesec2=2017-02-4&timesec1=1999-01-01&mission=Cassini&CASSINIrevnoint1=3&CASSINIrevnoint2=0_4&CASSINIspacecraftclockcount1=1485958000&CASSINIspacecraftclockcount2=1489511000"
  8. On "Spacecraft Clock Count":
    * Change min to something invalid like "1485958000.000xx"
      * Hit "x" to close "Saturn Orbit Number (By Range)", min would have red background, and search tab would be disabled, and no search would be performed.
      * Change min to "1485958000", return 2108 (search is performed and search tab is enabled)
        * Same as return of: "api/meta/result_count.json?observationduration1=1+2&observationduration2=5,100&timesec2=2017-02-4&timesec1=1999-01-01&mission=Cassini&CASSINIspacecraftclockcount1=1485958000&CASSINIspacecraftclockcount2=1489511000"
  9. Repeat step 6
  10. Close all widgets, only "Planet" and "Intended Target Name" are left open
  11. Open up "Mission":
    * Click "Voyager", return 1698
      * Same as return of: "api/meta/result_count.json?mission=Voyager"
  12. Click "Voyager Mission Constraints" and open up "Spacecraft Clock Count":
    * Type "10000" in min
      * Click "Earth Received Time", return 906
        * Same as return of: "api/meta/result_count.json?mission=Voyager&VOYAGERspacecraftclockcount1=10000"
  13. Close "Earth Received Time"
  14. On "Spacecraft Clock Count":
    * Type "10000x" in min
      * Click "Earth Received Time", min of "Spacecraft Clock Count" would have red background, and search tab would be disabled, and no search would be performed.
      * "Earth Received Time" might or might not pop out due to ajax async return before search tab is disabled, but no search would be performed
    * Change min of "Spacecraft Clock Count" to 9000 and hit enter, return 1668
      * Same as return of: "api/meta/result_count.json?mission=Voyager&&VOYAGERspacecraftclockcount1=9000"

* Pasting URL:
1. Open a tab, pasting the URL with correct value:
    * http://localhost:8000/#/observationduration1=12&observationduration2=5100&timesec1=1999-01-01T00:00:00.000&timesec2=2017-02-04T00:00:00.000&mission=Cassini&CASSINIrevnoint1=003&CASSINIrevnoint2=004&CASSINIspacecraftclockcount1=1485958000&CASSINIspacecraftclockcount2=1489511000&view=search&browse=gallery&colls_browse=gallery&page=1&gallery_data_viewer=true&limit=100&order=time1&cols=opusid,instrument,planet,target,time1,observationduration&widgets=CASSINIspacecraftclockcount1,CASSINIrevnoint1,timesec1,mission,observationduration1,planet,target&widgets2=&detail=
      * All widgets with normalized value is shown
  2. Open a tab, pasting the URL with some invalid value:
    * http://localhost:8000/#/observationduration1=1x2&observationduration2=5q100&timesec1=1999x-01T00:00:00.000&timesec2=2017-0s-04T00:00:00.000&mission=Cassini&CASSINIrevnoint1=0x03&CASSINIrevnoint2=004&CASSINIspacecraftclockcount1=1485,,958000&CASSINIspacecraftclockcount2=1489511000&view=search&browse=gallery&colls_browse=gallery&page=1&gallery_data_viewer=true&limit=100&order=time1&cols=opusid,instrument,planet,target,time1,observationduration&widgets=CASSINIspacecraftclockcount1,CASSINIrevnoint1,timesec1,mission,observationduration1,planet,target&widgets2=&detail=
      * search tab would be disabled and no search would be performed until user corrected all invalid inputs or closed widgets with invalid inputs

rfrenchseti (commented):
- What is the purpose of greying out the search menu on the left when an input is invalid? Especially since you can’t guarantee it will be disabled in time?
- I think it would be reasonable to remove the green result count when an input is invalid, since the count isn’t actually true. Maybe replace it with “—-“ or something similar.

rfrenchseti (changes requested):
- What happened to the idea of having a separate last request number for each search slug (input field)? I thought that we had decided that was necessary to avoid a race condition.
NEVER MIND - I was confused by something else and see what you did now. 
- for the CSS, class names should be functional (“searchInputError”) not descriptive (“red_background”). The whole point is you can change the color and it still makes sense. 
- api_normallze_input still has the weird double assignment.
- I think the green and red borders are a little too subtle for the average user to notice.

rfrenchseti (changes requested):
It occurs to me we aren’t being very useful to people who are red/green color blind. Perhaps a different way to indicate bad input, like making the border a little thicker in one case or the background slightly grey when it’s bad or something like that?

dstopp:
- Unless there is a specific reason you are using $.ajax for your json requests, please standardize to $.getJSON.  I was unable to find a reason to not use $.getJSON but would be interested to hear if you have one.

- The names of the text colors in CSS may be better named as "Warning", "Error", etc.  There are examples in bootstrap 4 to look at. IMO naming CSS objects "red", "green", etc, can cause you trouble downstream when you decide that your warnings really need to be "orange" ... then you might have to go rename all instances of the use of that CSS... 

- in "updateHash' - please return the hash value regardless of the flag, such the the flag option just adjusts the window.location if need be.

- line 118 in searchBehaviors - slug should not be global, i think...?  if so, please add comment.

- line 141 in searchBehaviors - please add comments, as i have no clue what is going on or why.  I'm sure renaming the CSS classes will help lots w/this bit of code.

juzen2003:
* For search tab grey out, it was trying to avoid user kept adding new tabs before they correct the current input. The grey out would come at the same time as we parse the final normalized api return data, which is before we updated the hash and performed search. It would always happened before search. I was referring to the note on the issue: ("Under no circumstances should an actual search be performed or the hash updated when there is an invalid value in one of the search boxes. In fact we probably want to disable the "Search" tab entirely in this case."). I can remove this one if we don't want the grey out.
* For api_normallze_input double assignment: are you talking about the following case? this is the only place I saw double assignment. When user leave focus by clicking "x" to close another widget. Final normalized API call is attached to range input change and widget close click event. Sometimes the return normalized data from on change event would include the slug from the closing widget if the ajax return is faster than the one from widget close click event, but this would not affect the final search because updateHash is called for both events to update url, and it's based on opus.selections. opus.selections would be up-to-date whenever the 2nd updateHash is called.
* For using $.ajax, it's just me more used to seeing a return of $.ajax call. $getJSON is a shorthand one, will use it to replace all ajax in new code.
* Will change the result count display when inputs are not valid
* Will rename CSS class name and make the border thicker for input changes
* Will modify line 141 & 118 in search.js
* Will modify updateHash so that it would always return hash and update the url based on flag

rfrenchseti:
I think the CSS classes should be called something like "search_input_valid", "search_input_invalid", "search_input_invalid_no_focus".

rfrenchseti:
- I'm ambivalent about the grey out. I could see a good argument for it, so let's leave it for now.
- For the double assignment I'm just talking about the code that says "reqno = reqno = ..."
- Right now, you send the complete current hash to normalize as the user is typing, right? This causes a fair amount of extra work for the back end. Would it be easy to send only the part of the hash relevant to the current search slug?
- As long as we're changing the Result Count look, we might as well make it look nice. How about putting a big red X instead of the number or something like that?

juzen2003:
Changes required from yesterday are added. As discussed, I'll move on to string hint. Will come back to this one later if more optimization or fixes needed.
# Here are current updates:
* CSS class is renamed based on recommendation and border is thicker for input changes now. There is also a grey background when current input is not valid while input is still changing.
* browse result count display would be a red "X" and green hints would be all "?" when user left focus or hit enter and there is an invalid values in range inputs
* short hash is used for normalized api call 
* updateHash would always return hash, and updating actual url would be based on flag
* double assignment in normalized api is removed.
* line 141 & 118 is modify, comment is added and CSS class is renamed.

rfrenchseti:
- We had discussed having the Result Count number be a "?" instead of "X" to be consistent with the hints but that may have gotten lost in the discussion.
- I think having a red X on a green background pretty much guarantees that someone red/green color blind can't see it. Let's just make the question mark in the result count the normal color that the number would be.
- As soon as you start typing in an input field, the border gets thicker, which is great, but unfortunately since the "empty input widget" border is thinner, the new thicker border makes the whole input widget move and resize slightly. It would be better if the original empty border was the same thickness, or perhaps it just had a little extra padding, so the widget didn't change size when you start typing.
- As you are typing, the normalize API gets all of the slugs currently open to normalize, not just the slug the user is currently typing in.

Issue #316, opened 2018-04-23
Auto-fill of search boxes


rfrenchseti:
We could support auto-fill of search boxes like Google does. Type in the first few letters of a text field and be presented with a drop-down of choices that starts with those letters.

rfrenchseti:
I've created a new API to do partial string searches. You give it the slug name you want to search on in the `<slug>.json` part of the URL. You give it all the current search parameters like normal after the ?, including the partial value typed by the user for `<slug>`. It does a search on all the constraints *except* `<slug>` and then returns the distinct strings for `<slug>` available in that search result, sorted in alphabetical order.

 It looks like this:
`http://127.0.0.1:8000/opus/__api/stringsearchchoices/productid.json?volumeid=COISS_2002&productid=1460960`
and returns JSON like this:
`{"choices": ["1_N<b>1460960</b>653.122", "1_N<b>1460960</b>868.118", "1_N<b>1460960</b>908.120", "1_N<b>1460960</b>944.118", "1_N<b>1460960</b>992.120"]}`

Note that the substring that matches the user's selection is marked in bold, so this should be rendered as HTML when displayed. There is also a limit= field that can set a limit on the number of results (some string search fields have 100,000s of unique strings). The default is 100:

`http://127.0.0.1:8000/opus/__api/stringsearchchoices/productid.json?volumeid=COISS_2002&productid=14609&limit=10`
gives:
`{"choices": ["1_N<b>14609</b>60653.122", "1_N<b>14609</b>60868.118", "1_N<b>14609</b>60908.120", "1_N<b>14609</b>60944.118", "1_N<b>14609</b>60992.120", "1_N<b>14609</b>61026.118", "1_N<b>14609</b>61061.118", "1_N<b>14609</b>61193.118", "1_N<b>14609</b>62279.118", "1_N<b>14609</b>62327.120"]}`

rfrenchseti:
Actually the return is now like:

`{"choices": ["1_N<b>14609</b>60653.122", "1_N<b>14609</b>60868.118", "1_N<b>14609</b>60908.120", "1_N<b>14609</b>60944.118", "1_N<b>14609</b>60992.120", "1_N<b>14609</b>61026.118", "1_N<b>14609</b>61061.118", "1_N<b>14609</b>61193.118", "1_N<b>14609</b>62279.118", "1_N<b>14609</b>62327.120"], "full_search": false}`

[...]

PR #554, opened 2019-01-07
String hints

juzen2003:
# Implementation related to #316 
### Modifications
* When a string is open, and user started to type 1 or more character, stringsearchchoices api would be called and dynamically updated the hinted values
* User could do the following options:
  1. continue typing and then left focus or hit enter (hint list would disappear), search would be performed.
  2. use mouse to hover over the hint list and right click one of options (hint list would disappear) and search would be performed.
  3. use down key to hover over the hint list and hit enter on one option (hint list would disappear) and search would be performed.
* hints are limited and displayed up to 25 items, and hints window height is set to fix size (200px) and scrollable vertically, we can modify this if needed.
* We didn't dynamically update the string input value when user was hovering over the list, we only update displayed value when user selected one. (This is the same behavior as Google) We could change this part to dynamically display focused value in the list if needed.

### Test Scenario
* Open multiple string inputs and type or select different values
  1. Open up "Product ID":
    * Type "125" (stringsearchchoices api is called to dynamically update hint list)
      * Hit enter, return 762
        * Same return as of: "api/meta/result_count.json?productid=125"
    * Change to "1258"
      * leave focus, return 5
        * Same return as of: "api/meta/result_count.json?productid=1258"
  2. Open up "Volume ID":
    * Type "CO"
      * Use mouse to hover over to hint list and right click on "COISS_2008" (input value would change to "COISS_2008"), return 1
        * Same return as of: "api/meta/result_count.json?productid=1258&volumeid=COISS_2008
    * Hit delete to "CO"
      * Use down key to focus "COVIMS_0006", hit enter, return 4
        * Same return as of: "api/meta/result_count.json?productid=1258&volumeid=COVIMS_0006
  3. Focus back to "Product ID":
    * Change to "125", leave focus, return 8
      * Same return as of: "api/meta/result_count.json?productid=125&volumeid=COVIMS_0006
    * On q-type, select "ends", return 0
      * Same return as of: "api/meta/result_count.json?productid=125&volumeid=COVIMS_0006&qtype-productid=ends
    * Focus back to input and delete to "12", select "1/1484536612_1 from hint list", return 2
      * Same return as of: "api/meta/result_count.json?productid=1/1484536612_1&volumeid=COVIMS_0006&qtype-productid=ends


rfrenchseti:
String hinting isn't working reliably for me. There appears to be some kind of timing error with the back end. I sometimes see the search results for a search I did one keystroke ago, and I very often don't see a result at all.

Open Data Set ID.
Type N. No hinting.
Type A. No hinting.
Type /. No hinting.
Backspace over the /. Get hinting for "NA/" with "NA/" in bold.
Backspace over the A. Get hinting for "NA" with "NA" in bold.

I have a same problem with Volume ID.

PR #560, opened 2019-01-08
Input validation and string hints

juzen2003:
# Merge branches related to #535 and #316 
* String hints:
  * Footer and header messages are added to the drop down list of string input
* Range input validation:
  * Only use current focus range input slug to call normalized api when every input event happened. A final call of normalized api with all slugs would be performed when change event happened.
  * Add cleaner codes to updating hinting labels


dstopp:
- highlight of items in pulldown should not be underlined
-

juzen2003:
* Underline of items in dropdown list is removed.
* Thicker border would not change the size of widget.

dstopp:
- when nothing is selected from pulldown, the typed in characters should be dealt with - currently, they are leftover remnants, not sure that is desired behaviour
- using the remnant characters to continue the search did not work and resulted in two separate requests to the DB for data
- emptying the input field resulted in a call to the DB looking for empty string.  That probably shouldn't happen

rfrenchseti:
As later discussed, Debby's point number 1 is not correct. I'm not sure what she means by point #2, because it works fine for me to leave a field and then come back. Point #3 is probably an actual bug.

rfrenchseti:
Browse Results normally has commas in large numbers. If you lose focus on a bad range input, and kill the bad widget, the result comes back from being a ? but it no longer has commas in it.

You don't see this if you just fix the bad input usually because then it actually runs a new search.

In fact there's this weird glitch when you first lose focus, the commas go away before the spinner starts in the result count

rfrenchseti:
This is for @rfrenchseti: There is a timing bug in the string hint API. Do a search that requires a long time. While the spinners are going, start typing into a string field. You won't get any hints, since the original search isn't done yet. But it appears this negative result is also cached, so later when the search results are actually available, typing those same characters doesn't give you hints then either.

juzen2003:
* Fix the browse result to display with "," when it's a large number.
* Move autocomplete to getWidget instead of attaching it to input.STRING on input event. No longer need to init the source for autocomplete when focus event happened on input.STRING at the very beginning.
* Would not call normalized api when input.RANGE field is an empty string on input event.

rfrenchseti:
If you have a string widget open, and a range widget, and put an invalid value in the range widget, then when you type in the string widget it still calls the string hints API even though the rest of the search is invalid, and it calls it with the illegal range value in the hash, causing an exception.

rfrenchseti:
Another bug in input validation:
- Open two range widgets, like Right Ascension and Declination. Click inside one of them. Type "1x". Click inside the other. Type "1x". Click back in the first one. The field will stay red instead of switching to yellow/grey.

rfrenchseti:
It also seems weird that when you have an illegal value in a range widget, and update a string widget, the "Result Count" at the top (which is already ?) flashes and then goes back to ?.

rfrenchseti:
I'm finally getting around to actually looking at the code.

- search.js lines 15 and 228: `let regexForShortHash = /(.*)&view/;`
I'm actually really surprised this works all the time, because slugs can be in any order. For example, you can have search slugs at the end of a URL (although the UI doesn't create these, you can do them by hand). Such as:

https://tools.pds-rings.seti.org/#/view=search&browse=gallery&colls_browse=gallery&page=1&gallery_data_viewer=true&limit=100&order=time1&cols=opusid,instrument,planet,target,time1,observationduration&widgets=observationduration1,planet,target&widgets2=&detail=&planet=Earth&observationduration1=.05&observationduration2=.5

In this case, cutting everything after the view= is clearly the wrong thing to do. However, the part that surprised me is that OPUS appears to rewrite the URL when it's first entered and put all the search slugs at the beginning! So I guess this really does work, as long as we never change the part of OPUS that rewrites the URL. It might be useful to add a comment to this effect, that we rely on the URL order. Also the comment "short hash" doesn't actually tell the reader anything useful. What is a short hash?

- The argument `data` to `validateRangeInput` doesn't mean a lot to me, and you have to search through two other functions to know it comes from normalizeinput.json. Can you call it something like "normalizedInputData" instead?

- I don't know JS or jQuery well, but isn't `value` here already set to `data[eachSlug]`? So the if doesn't need to recalculate it. You do use `value` a few lines later in the else clause.
```
        $.each(data, function(eachSlug, value) {
            let currentInput = $(`input[name="${eachSlug}"]`);
            if(data[eachSlug] === null) {
```

- When converting range endpoints and mult hints to ?, and then switching them back, we end up calling `getHinting` on each slug right away. At this point, the search is still the OLD search, not the new one. This means that the hints get updated with the old values, then the search gets run, then the new values are shown. You can actually watch this happen when the search is long enough. Do you actually have to call `getHinting` immediately, or can you just let the normal search do its thing? You'd probably have to know if the hash has changed enough to trigger a new search, and if not do the current behavior.

- I think `o_search.slugReqno` should be called `o_search.slugNormalizeReqno` to be clear.

- `performSearch` seems like a bad name for this function because it doesn't, actually...perform a search! All it does is call normalizeinput.json and do stuff with the results. Or does it? Maybe the search is hidden inside `updateHash`, in which case a comment would be helpful. I'm not very familiar with that code. Also same comment about `data` as above.

- search.js line 86, there's no point in making this an "else" since the previous if returns.

- search.js line 156 uses the word "hover" but I think you actually mean "change focus".

- I'm having a hard time parsing the language of some of the comments, like "when there is no change event happened while user focus out" or "Use only the current slug to call normalized api while input event happened". Comments are free (and will disappear when we minimize) so might as well make them more verbose, good English, and useful to future developers. Can you generally go through and clean up the new comments? The general rule should be that the code explains itself, but comments explain the strategy or "big plan" or things that are just not obvious to anyone. If you need help with the English grammar, let me know.

- search.js line 187:
```
            // if input field is empty, do not perform api call
            // if input value didn't change from last successful search (normalized value), do not change border
```
These two comments don't seem to agree with each other even though they refer to the same if.

- search.js line 197:
```
                // if a newer input is there, re-call api with new input
                if(data["reqno"] < o_search.slugReqno[slug]) {
                    return;
                }
```
This comment doesn't actually describe what the if does.

- There are inconsistent blank lines at search.js lines 246, 257, 282, and 301.

rfrenchseti:
Open a range widget. Put the focus in the min box. Type a valid value. Really quickly hit tab and type an invalid letter. The second box will turn solid red even though it has the focus.
You have to do it fast enough that it hasn't started doing the original search query I think

juzen2003:
# Modifications
* search.js lines 15 and 228:
  * Rename regexForShortHash to regexForHashWithSearchParams
  * Opus would rewrite the URL when a URL is pasted, all the search related slugs would be moved ahead of "&view" (please refer to hash.js getSelectionsFromHash and updateHash functions)
  * Add comments about the above effect in all places that regexForHashWithSearchParams is used
* validateRangeInput:
  * Change argument "data" of validateRangeInput to "normalizedInputData"
  * You are right about the "value", change the "data[eachSlug]" to "value" to avoid recalculating
* getHinting:
  * Now it would let the new search update the hints
  * If input is changed from an invalid value to a valid value but URL is not changed, it would call getHinting on each slug
* o_search.slugReqno:
  * Change "slugReqno" to "slugNormalizeReqno"
* performSearch:
  * The function would do the following:
    * Call the final normalized api, and validate the return json from that api call
      * If the result is invalid, it would not do anything. UI effects are displayed (red background and "?" in hints are processed in validateRangeInput which is in the middle of performSearch function).
      * else, the result is good, update the URL (updateHash). By updating the URL, backend would query the database based on search params in the URL and display the result.
  * Above is the reason I called it "performSearch", change it to "parseFinalNormalizedInputDataAndUpdateHash"
  * Change "data" to "normalizedInputData"
* search.js line 86:
  * Remove the "else". Originally I was thinking to put "if" and "else" so that they are more readable. (like it's mentioned in above performSearch)
* search.js line 156:
  * Update the comments
* Clean up comments:
  * Sorry, I was having bad comments on these codes. Would write more verbose comments in good english
  * Clean up and re-write all comments
* search.js line 187:
  * Update comments
* search.js line 197:
  * Update comments
* Remove inconsistent blank lines at search.js lines 246, 257, 282, and 301
* Bug fixed:
  * Open one range widget say "Observation Duration". Focus in min and type "12" (or something valid)
  * Hit tab to focus in max and type "1x" (or something invalid)
    * Instead of having red background, now the max would have yellow border. 
    * Search is based on value in min. Value in max would be excluded since user is still in focus on max

Issue #575, opened 2019-01-17
String hints don't move on screen when widget scrolled


rfrenchseti:
- Type into a string field and get the hint dropdown box
- Use the mouse scroll wheel to scroll the window with the search widgets
- The search widgets will move, but the dropdown box will stay where it was

Issue #574, opened 2019-01-17
String hints sometimes stay around after enter pressed


rfrenchseti:
- Type into a string field
- Right after entering the last character, hit ENTER very quickly
- When the string hints come back, a drop down box will appear even though the user has hit ENTER

Issue #573, opened 2019-01-17
Normalizeinput can overwrite user's actual value


rfrenchseti:
- Enter a valid value in a range field
- Leave focus to another range field so that a search is triggered
- Start to type a value in that field (valid or not doesn't matter) quickly enough that the first all-normalizeinput hasn't yet returned
- When the all-normalizeinput returns, it will write the normalized version of the original value in the second field, even though that field has since changed. Assuming the second field was original empty, it will erase the new input.

This is easiest to see with an API delay, but I can reproduce it easily without just by using tab and typing quickly.

I think you have to fix this by saving the value of each input field when all-normalizeinput is first called, and then checking to see if the inputs have since changed when all-normalizeinput returns. If a field has changed, don't overwrite the value.

PR #578, opened 2019-01-17
String hints issues

juzen2003:
# Modifications:
* Issue #573 
  * Now if the focused range input value is changed between calling final normalized api and waiting for return of api call, that input will not be treated as one of the search parameters because user is still typing a new value. When user finishes typing the value in focused range input, it will be treated as search parameter if user leaves focus or hits enter.
  * Test Scenario:
    * Open "Observation Duration", and type "12" in min and "5100" in max
    * Open "Right Ascension", and type "3" in min
    * Quickly (or add api delay) leaves focus from "Right Ascension" min to "Observation Duration" min, and change "Observation Duration" min to "14"
    * "Observation Duration" min will have green border and search will be run with "5100" in "Observation Duration" max and "3" in "Right Ascension" min. "Observation Duration" min will be excluded because user is in focus there and trying to type a new value.
    * Change focus from "Observation Duration" min to max, and don't type anything.
    * Search will be run with "5100" in "Observation Duration" max "14" in "Observation Duration" min and "3" in "Right Ascension" min because user didn't change value in "Observation Duration" max.
* Issue #574 
  * Add a parameter "displayDropDownList" to deal with the source of the dropdown list (coming from an ajax return)
  * Test Scenario:
    * Open "Product ID" and type "1", select any of items in dropdown lists to perform the search
    * Hold backspace all the way until the value is "1/" and hit enter real quick (or add api delay)
    * Search will be run when enter is hit, and there will be no dropdown list shown
* Issue #575 
  * Now if the widget column window is scrolling, the string hints list will disappear
  * Test Scenario:
    * Open "Product ID" and type "1", string hints dropdown list will show up
    * Scroll the widget column window, and the dropdown list will disappear (even with only 1 scroll, it will disappear)

### DIFF
--- a/opus/application/static_media/js/search.js
+++ b/opus/application/static_media/js/search.js
@@ -278,6 +278,10 @@ var o_search = {
             o_search.parseFinalNormalizedInputDataAndUpdateHash(slug, url);
         });
 
+        // $("input.STRING").on("scroll", "#mCSB_1_container", function() {
+        //     console.log("scrolling");
+        // });
+
         // filling in a range or string search field = update the hash
         // range behaviors and string behaviors for search widgets - input box
         $('#search').on('change', 'input.STRING', function(event) {
--- a/opus/application/static_media/js/search.js
+++ b/opus/application/static_media/js/search.js
@@ -41,21 +41,27 @@ var o_search = {
                     currentInput.val(opus.selections[eachSlug]);
                     opus.allInputsValid = false;
                 }
+                if(currentInput.hasClass("input_currently_focused")) {
+                    delete opus.selections[eachSlug];
+                }
             } else {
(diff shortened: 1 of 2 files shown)


juzen2003:
* Update on Issue #573
  * Now if the focused range input value is changed between calling final normalized api and waiting for return of api call, that input will still be treated as one of the search parameters but the displayed value would not be overwritten.
  * Test Scenario:
    * Open "Observation Duration", and type "12" in min and "5100" in max
    * Open "Right Ascension", and type "3" in min
    * Quickly (or add api delay) leaves focus from "Right Ascension" min to "Observation Duration" min, and change "Observation Duration" min to "14"
    * "Observation Duration" min will display as "14" with green border. Search will be run with "12" in "Observation Duration" min, "5100" in "Observation Duration" max, and "3" in "Right Ascension" min.

PR #592, opened 2019-02-12
Font awesome

fyellin:
Add --glyphs to use fontawesome glyphs instead of letters as markers.  
For now, I think the results are rather ugly. but Jinja2 makes it reasonably easy to fix if we want.

Rob, if you want to play around, and see if there are better glyphs to use, look at lines 91-96 of templates/log_analysis.html.  This is where the letters and glyphs are set.


rfrenchseti:
There was a spurious ; on line 95 of log_analysis.html that was preventing the download icon from showing.

rfrenchseti:
I think there must be a bug with the gallery icon. Consider this session:

82.32.169.140 (0:30:38)‌ / 2018 Dec 31 16:39:28 (0:05:42) 

There are multiple viewings of the gallery, but there's no icon. Is it possible you're only displaying the icon with view table?

rfrenchseti:
I like the search, gallery, and zip download icons. I think they are intuitive.

The columns icon makes me think they are looking at a table, which we don't have any way to distinguish going forward. But I can't find a better icon.

Likewise I'm not crazy about the X-circle for old slugs, but I can't find a better icon.

I think the icons would look better with just slightly more horizontal spacing.

fyellin:
>
> Line 95

Fixed

On Tue, Feb 12, 2019 at 4:01 PM Robert French <notifications@github.com>
wrote:

> I think there must be a bug with the gallery icon. Consider this session:
>
The column glyph and the gallery glyph were swapped.  It was telling you
that the user had viewed the gallery, but incorrectly!

fyellin:
> I like the search, gallery, and zip download icons. I think they are
> intuitive.
>
Okay.  Do you want me to get rid of the letters and only use glyphs?  I was
never actually intending that --glyphs be an option.  Just a way for you to
compare outputs.

> The columns icon makes me think they are looking at a table, which we
> don't have any way to distinguish going forward. But I can't find a better
> icon.
>
Yeah.  I know.

> Likewise I'm not crazy about the X-circle for old slugs, but I can't find
> a better icon.
>
book-dead?  biohazard?

> I think the icons would look better with just slightly more horizontal
> spacing.
>
If I do this, I'm also going to make the left-hand column a little bit
wider.  Maybe 33% instead of 25%.

I'm not sure I know the html magic to give a little bit more horizontal
spacing.

fyellin:
p.s.  I'm really glad that I switched to jinja2.  It let's you write code
that more closely resembles real Python.  It also lets you pull out common
code into a single place, and even just makes the code easier to read by
being able to refactor it into smaller routines.

It's only disadvantage seems to be that it generates more whitespace, and
there doesn't seem to be an easy way of telling it to remove it.  My

PR #595, opened 2019-02-15
New ui dave

juzen2003:
# Update
* remove white background rail of perfectScrollbar and make it transparent in table & gallery view
* make table header sticky by using .dataTable as the container for table view y-scrolling
* remove little gap between table and y-scrollbar in table view by assigning bootstrap spacing notation in html
* rename & add perfectScroll in browse/collections/detail pages
* disable x-scrolling in gallery view
* udpate code & rename cdn fallback files with versions in it
* pease ignore the change in requirements-python3-windows.txt, this is just to remove the complaint of security warning in github.


juzen2003:
# Update
* update detail page perfectScrollbar after all metadata_v2 api loads are done
  * this will make perfectScrollbar to show up a bit late but it will remove that weird length changing behavior of the scrollbar.
* make collection gallery perfectScrollbar visible in collection page
* remove codes that tries to set position of x-scrollbar

PR #597, opened 2019-02-21
New ui dave

juzen2003:
**This is a dummy pull request with Dave's updates**
# Update:
* Fix ps behavior in detail page
* Add a navbar on top of search sidebar with "Reset Search" and "Reset Everything" options
  * "Reset Search" will reset all search params
  * "Reset Everything" will reset all search params and empty shopping cart
  * Dark background of navbar is removed because it looks a bit weird, please let me know if we want to change to two buttons at the bottom
* Fix ps behavior in search sidebar and widgets, ps will only show up when scrolling is needed
  * Remove unwanted white spaces at the bottom of search sidebar and widgets
  * Fix the sidebar width so that the width will not increase when a .menu_help message shows up
    * Not sure if this is the desired behavior, and we can always remove the fixed width by updating css styling
* Update ps in table view so that it won't go through the navbar or sort-order-container

# Next item in progress:
* Start to check at sortable cards in search widgets


dstopp:
-I'd like to do something a little different w/the nav on the search sidebar, it doesn't feel right.  Still thinking.
-Feel free to remove unused code <vs. comment out> when you are done with it
-Bootstrap4 has a very rich library of classes, so the preference would be to use those vs. adding them to opus.css.  <sidebar-nav>.  I'm also a little unclear what sidebar-area is doing.
-The PS on the table is always 'fat', vs the PS on the gallery which 'grows' when you hover.  
-It looks like you did not move the PS on the data table into a diff element, as I can see the top is defined to a specified number of pix.  That may not work well for all browsers or sizes of media, so I'm hoping for a better resolution.
-Minor spelling typo on line 417 in opus.css
-I am unsure why you set the height of the PS for the dataTable to 11px ...?  Doesn't seem consistent with the other usages of PS
-Why does the tableScrollbar have min length but the others do not? I was confused by that.
-in browse.js, line 840, 841, curious why you need find().find() ... that seems more complicated than necessary, is there perhaps a simpler way to get the element you need?  Feel free to ask for ideas.
-in detail.js, lines 50 -I suspect you would have neater code if you refactored using $.each.  I am also a fan of moving away from naming the data retrieved from $.getJSON 'json' - perhaps 'categories' or even just 'data' ... NOTE: I realize you didn't write this code and are just updating it, so this is just a generic observation.
-It looks like you are creating the PS over and over again when you get the collection - I think you only need to make it once in a lifetime, can you please check on that? <line 72/80>
-in opus.js, line 400 - you may want to further identify those classes using #sidebar just for clarity, if nothing else..  I think .reset is pretty generic of a class, so just a thought.
-The confirm text should relate to the action, ie., if they click on 'reset', I think it should probably say 'reset' vs. restart.  I was thinking there would also be a confirm modal to empty the cart but i don't see one.
-do you really have to hide/show the ps__rail ?  I was thinking that library was smart enough to do it all by itself, it seems to do it in other places <or maybe i'm just confused.

juzen2003:
# Update:
* Add re-usable confirmModal.html
  * Add a multilines_template_tags.py file to allow multiline template tags (refer to http://zachsnow.com/#!/blog/2016/multiline-template-tags-django/)
    * this is to have better coding style when using confirmModal
* Make reset search and reset everything as buttons at the bottom-left corner
  * there will be a gray line between search sidebar and search widget when y-scrollbar is not needed in search sidebar

# Modification based on Debby's feedback:
* .sidebar-area: this is to set the fixed width so that search sidebar will not get wider when user click "Ring Geometry Constraints". I saw boostrap has classes for percentage of width relative to element's parent, like w-25, but I didn't find one for fixed size like width: 400px. That's why I just put that in opus.css.
* In browse.js, line 840 & 841: thanks for catching this, it's updated with only 1 .find to access the element
* In detail.js, line 50: update to use $.each and rename json to categories for better readability
* PS in collections.js: we need to create ps for the containers in collections page over and over because new html (from collections.html) is rendered every time when there is a change in collection page. If we don't create ps for the containers from newly rendered html, there will be no ps.
* hide/show ps__rail: theoretically .update should do all the stuff to hide/show ps. But when resizing happened, there are some cases:
  1. if scrollbar is not needed, the whole ps__rail & ps__thumb will not disappear until the first scroll event happened
  2. if scrollbar is not needed, ps__thumb will disappear but ps_rail will still be there
  * Manually hide/show ps will remove these cases
* .dataTable ps related feedback:
  * Ps can't be attached to non-typical container (like tbody or table, please refer to https://github.com/utatti/perfect-scrollbar/issues/256) If we want to attached ps to tbody, we will have to set display to block and this will break the structure of the table. In this case, we will have to re-calculate the width of td & th. We have to get the max width value for each column and apply for both th and td so that the table will look nice and aligned. This is not effective because we needs to do this when new contents are loaded. I'm still trying other methods on this post: https://stackoverflow.com/questions/17067294/html-table-with-100-width-with-vertical-scroll-inside-tbody
  * Eventually I'm planning to create a new branch and try the div table structure.

rfrenchseti:
I think it looks weird for the grey vertical separator to appear and disappear with the scrollbar. For example, shorten the window so that there is just no scrollbar, but only by little. Now expand one of the categories on the left. The scrollbar will appear and the vertical line will disappear. But because the PS has no "track" to run in, there is now no vertical separation near the bottom of the window at all. I think the separator should stay there all the time.

juzen2003:
# Update:
* Search page:
  * Reset buttons' behavior:
    * Reset Search:
      * Modal will open when one of the followings happened:
        1. Any new widget is open
        2. Any selections or values are inputed in widgets
      * It will reset all search params and widgets to default
    * Reset Search and Metadata:
      * Modal will open when one of the followings happened:
        1. Any new widget is open
        2. Any selections or values are inputed in widgets
        3. Any of metadata is selected
      * It will reset all search params, widgets, and metadata to default
  * Update color of reset buttons to bootstrap class btn-secondary
  * Vertical divider will exist all the time
    * Use bootstrap class to add color and border
  * Add the horizontal shadow divider between reset buttons and search sidebar
* Collection page:
  * Remove extra bottom space of download options sidebar (left pane) when scrolling happened
* Y-scrollbar disabled/overlapped issue in table view
  * This was caused by .hide() in updateBrowseNav doesn't fully hide gallery view ps rail after:
    1. InfiniteScroll load event happened
    2. Click tab to switch over to table view
  * Instead of using .hide()/.show(), use the method of class toggling to hide/show ps

juzen2003:
# Update:
* Search page:
  * Reset buttons' enabled/disabled behavior:
    * Reset Search:
      * Button will be enabled when one of the followings happened:
        1. Any new widget is open
        2. Any selections or values are inputed in widgets
    * Reset Search and Metadata:
      * Button will be enabled when one of the followings happened:
        1. Any new widget is open
        2. Any selections or values are inputed in widgets
        3. Any of metadata is selected
    * Remove cursor when buttons are grayed out
* base.html:
  * Rename reset buttons classes by adding prefix op and using "-" instead of underscore
* opus.js:
  * Remove unwanted variables

PR #600, opened 2019-02-28
New ui dave

juzen2003:
# Update:
1. vertical line goes all the way down to the footer
2. modify shadow divider so that it doesn’t touch/connect to vertical line
3. remove extra spaces at the bottom of widgets area when scrolling


juzen2003:
# Update:
* Widgets sortable:
    * if scrolling is needed, widgets column will scroll when a card is dragged and moved to the bottom of the area
    * Widgets or any widgets' options will not be dragged into other cards
* Make scrollbars in search sidebar and widget column have the same top position
* Make shadow divider thinner

PR #618, opened 2019-03-26
New ui dave

juzen2003:
# Update
1) Search: Add widgets until one is just off the bottom of the browser. The scrollbar doesn't show up until you add another widget.
    * Now scrollbar will show up in this case
5)  Search: Make the browser just tall enough to hold all the categories on the left without a scrollbar. Select a search result that creates a new category. The scrollbar does not appear.
    * Now scrollbar will show up when we select any input from widget area and a new category is open.
* Update codes with suggestions from the discussion with Debby this afternoon

### DIFF
--- a/opus/application/static_media/css/opus.css
+++ b/opus/application/static_media/css/opus.css
@@ -615,6 +615,13 @@ a:-webkit-any-link, a::any-link {
     cursor: default;
 }
 
+/* this is used to style grey out disabled buttons */ 
+.op-button-disabled {
+    pointer-events: none;
+    cursor: default;
+    opacity: 0.1;
+}
+
 #cart .cart-gallery-side {
     /*margin-top:.25em; */
     background-color: #303030; /* #383838; */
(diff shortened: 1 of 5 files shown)


dstopp (changes requested):
couple of things:
-in general, I think we should probably keep the length of variable names down to less than 40 or 50 chars; when they get too long they are unwieldy and make things worse.
-I was thinking you should be able to actually do away w/that global all together <displayPrevButtonInFirstMetadataModalOfThePage>; see notes in slack re: pseudo code suggestion.

dstopp (changes requested):
browse.js, line 1262 - I think it would be better to qualify the selector w/a non-bootstrap class.  My suggestion would be to use $("#galleryViewContents .prev")  .. etc.  That makes it less bootstrap-dependent re: logic, etc.
Also - can you set the show/hide of the hand such that it 'takes up space' instead of centering the single hand when the modal is at the end/beginning of the collection of observations?  It would be better if the hands did not move, but just appear and disappear as need be <or you can gray them out, that may be better yet, set it opaque like .1 or something>

PR #619, opened 2019-03-29
New ui dave s031 c028

juzen2003:
# Update
* S031:
    * Spinner next to "General Constraints" is added, it will show up when get menu is called and disappear when menu is updated.
    * It's between "General Constraints" and the down arrow (like current opus)
* C028:
    * Spinner next to "Total Size":
        * It will show up when:
            * any input in product area is selected or unselected, and disappear after the return of status.json (new total size value will be updated)
            * any item is added or removed in cart's gallery view, and disappear after the return of new total size value. (actions of fadeIn/Out of spinner is added in editCart in this case)
    * Spinner for cart badge:
        * It will show up when:
            * any item is added/removed in browse/cart gallery view, and disappear after the return of cart count
            * we refresh the page or click "OPUS", and disappear after the return of status.json in initCart

### DIFF
--- a/opus/application/apps/cart/templates/cart/cart.html
+++ b/opus/application/apps/cart/templates/cart/cart.html
@@ -8,8 +8,8 @@
             </div>
             <h1>Download Options</h1>
             <!-- <p>Total Files: <span id="total_files">{{ download_count }}</span></p> -->
-            <!-- <p>Total Size (before zip): <span id="total_download_size">{{ total_download_size_pretty }}</span></p> -->
-            <p class="op-total-size">Total Size (before zip): <span class="spinner">&nbsp;</span><span id="total_download_size">{{ total_download_size_pretty }}</span></p>
+            <!-- <p>Total Size (before zip): <span id="op-total-download-size">{{ total_download_size_pretty }}</span></p> -->
+            <p class="op-total-size">Total Size (before zip): <span class="spinner">&nbsp;</span><span id="op-total-download-size">{{ total_download_size_pretty }}</span></p>
 
             <div id="download_options">
                 {% for product_cat, product_list in product_cat_list %}
--- a/opus/application/apps/cart/templates/cart/cart.html
+++ b/opus/application/apps/cart/templates/cart/cart.html
@@ -1,7 +1,7 @@
 <div id="download-options-container" class="sidebar_wrapper pr-0">
     <div id="cart-sidebar">
         <div id="cart_summary" class="well well-lg">
-            <div id="download_links">
+            <div id="op-download-links">
                 <h1>Zipped File Links</h1>
                 <ul class="zippedFiles"></ul>
                 <span class="spinner">&nbsp;</span>
@@ -9,7 +9,7 @@
             <h1>Download Options</h1>
             <!-- <p>Total Files: <span id="total_files">{{ download_count }}</span></p> -->
             <!-- <p>Total Size (before zip): <span id="total_download_size">{{ total_download_size_pretty }}</span></p> -->
-            <p class="total-size">Total Size (before zip): <span class="spinner">&nbsp;</span><span id="total_download_size">{{ total_download_size_pretty }}</span></p>
+            <p class="op-total-size">Total Size (before zip): <span class="spinner">&nbsp;</span><span id="total_download_size">{{ total_download_size_pretty }}</span></p>
 
             <div id="download_options">
                 {% for product_cat, product_list in product_cat_list %}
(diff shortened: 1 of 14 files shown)


dstopp (changes requested):
-please remember to name opus-specific css things starting w/op- for consistency.
I'm a little perplexed by your class names; you have total-size, which appears to be a spinner; you also have menu_spinner which is also a spinner.  I suggest that if these are both classes for spinner id, there may be a way to make the class names more consistent and obvious, ie, size-spinner, menu-spinner, etc.
-browse.js - it is unclear to me why you need reachLastData; i don't see it being used in the code.  I was also sort of hoping we would be able to avoid adding in another global variable.  Also, i don't believe that just checking data.count === 0 will be sufficient to determine that you are at the end of the collection of data.

rfrenchseti (changes requested):
- It seems like "menu_spinner" was already present before these changes. Now would be a fine time to rename it to "op-menu-spinner" everywhere.
- Better yet, it seems like you're using CSS to do ".total-size .spinner" so why not just do ".menu-text .spinner" and get rid of "menu_spinner" entirely?
- I agree that the class should be named "op-total-size".
- I don't think the comment "Properly display menu spinner" before the ".menu_spinner" CSS entry is very useful to someone who doesn't know why this entry is there.
- I agree about reachLastData not being used.
- Please remember spaces after if/while/etc.
- I think checking that data.count === 0 tells you that you have tried to load a "page" that doesn't have anything in it. It doesn't tell you that the current page is actually the last page - you have to go one past the last page and do an extraneous API call. This doesn't seem good to me. We should already know that we're at the end of the available results and not make an API call we KNOW will return zero data.

juzen2003:
# Update 
Thanks for the feedback.
* All spinner classes are moved to the same area in css file, and they are renamed to start with "op".
* Comments describing each spinner are added.
* One space is added after if in browse.js, will clean up rest of js later 
* reachLastData is not used and removed.
* data.count === 0 checking is to remove the spinner when we reach to the end of results (because we don't need to load "next" page anymore). Instead of checking data.count === 0, checking if current page === opus.pages (total number of pages for the search results) makes more sense. This piece of code is related to B030 and is still in progress. The code was there before I branched out for S031 and C028, sorry about the confusion. I'll work on B030 tomorrow.

rfrenchseti:
Regarding B030, we’re planning to get rid of page so you might want to wait until Monday’s teleconf before working on it. But I would think the only real issue here is that right arrow and clicking on the finger are doing different things for some reason. I would expect the logic to be identical. But I’m confused by your comment in general...B030 is just talking about moving to the next observation valid observation. That’s a different issue from showing the spinner when trying to navigate past the last observation.

dstopp (changes requested):
might as well change #cart_count to op-cart-count, since it seems to only be used in a few places and you are working on that code area.  I guess same thing goes for total_download_size - op-total-download-size.
in menu.js, line 71; why do you need to set the css vs. fadeIn("fast")?  Same comment for cart.js, line 374 <and wherever else that might appear>.  I would prefer to not see css added this way if there is an alternate way to show the spinner <or element>.

juzen2003:
# Update
* rename #cart_count to #op-cart-count
* rename #result_count to #op-result-count
* rename #total_download_size to #op-total-download-size
* line 71 in menu.js and line 374 in cart.js, remove .css, use the method of adding/removing class to make display: inline-block for spinner
* clean up code styles in opus.css and js files
* regarding to B030, will wait until Monday

rfrenchseti (changes requested):
- When the spinner appears next to "General Constraints", it shows up between the text and the expand/collapse arrow. This means that the arrow moves over and then back again when the spinner appears and disappears. I think this is awkward-looking. The spinner should be to the right of the arrow. Ideally I think it would look nicest right-justified so it's sort of out of the way, but do what looks best.

- The behavior on the search page is now the same as the current version, but there's always been something that bothered me about the current version and now seems like a good time to fix it. The current process is:
  - Change the search
  - Put a results spinner and get the new results
  - Once the results count has been returned, query for a new menu and hint numbers and put spinners in those places
  - Once each of those has been returned, remove those spinners

  The end result is that when you change the search, you see a single spinner for result count, even though the menu and the hint numbers are wrong and will be reloaded soon. The result count can take a long time to return, so it gives the impression those numbers are correct.

  I think the right thing to do is put spinners on the menu and all the widgets with hints the instant the new result count is queried, rather than waiting for the individual API calls to be triggered. That way when you change the search, all the spinners will appear immediately, and then go out one by one as the API calls return.

- When you click on the "Cart" tab, it is displayed as black text on a blue background. However, when you (de)select one of the product types, the word "Cart" is changed to white on a blue background.

PR #622, opened 2019-03-30
Bug 427 bootstrap upgrade

dstopp:
-slider changes
-bug fixes for help panel, comment out help for the time being; make dropdown go left so it doesn't fall off page
-remove centering of observations
-remove page number from nav bar
-bug fixes for add/remove range
-modify transitions for the search menu
-fixed bug that added multiple opusId to metadata select
-fix for c044 <was c038>
-remove unused globals; moved globals to appropriate namespace
-add menu to dataTable column
note: merged in latest new_ui.


rfrenchseti (changes requested):
**ui/templates/base.html**
- We are not putting the total result count next to the slider, right? Then you can remove the commented-out total obs count field at line 111.

**opus.css**
- Please put a space after the ":"s for op-observation-slider and op-observation-text and ui-slider-handle
- It seems like there is excess CSS for HVR ripple in and out. We are only using one of these, right? The others should be removed.

**browse.js**
- Line 19 initializes galleryBoundingRext with two LISTS: [0] and [0]. However, every other use in browse.js assumes they are numbers, not lists. I think the initialization should be {'x': 0, 'y': 0}. (Also coding standard is space after ":" I think)
- Line 78 Coding standard is space after "," in argument list. But isn't this a problem anyway? I assume the scroll position should be the same in gallery and table when you switch back and forth.
- Line 360 Add space after ":"
- Line 514 We seem to be inconsistent about whether we prefer == or === for comparisons
- Line 521 should do something if the if is false
- Line 600 there's an extra space after "="
- Line 667 excess spaces around function argument
- Line 925 should have spaces around "="
- Line 1998 I find the unnecessary () confusing
- Line 1205 I think undefined needs to use "===" for comparison
- Line 1406/7 should these use "let"?
- Line 1413 I find it confusing to use "++" prefix instead of postfix here (xCount++)
- Line 1420 space after ":"

**hash.js**
- Line 190-191 I think this is obsolete

**menu.js**
- Line 74 Extra space after "("
- Line 87 Why is this ' instead of "? Same for line 94

**opus.js**
- Line 98 Use === for undefined

rfrenchseti:
Some problems I found with the slider. If you agree with these, please add issues:
- When you release the mouse from the slider, there is a pause before the spinner shows up saying it's loading data. It appears to do an API call first, then start the spinner. Most visible with API delays are turned on.
- Slider can go all the way to the max obs number. We probably only want it to go such that the browser is full vertically with the final obs on the bottom line and the slider obs number is in the top left.
- Start fresh. Select Jupiter. Go to gallery. Drag slider all the way to right (see above). Slider will pop back to 1861. The screen with have one "page" of data starting at 1861 but not fill the screen and you can't continue to scroll down.
- When you grab the slider to move it, it pops some amount to the right first.
- Sometimes when you move the slider, the resulting top row of the gallery cuts a thumbnail in half vertically.
- In table view, the slider moves a fairly random amount. Sometimes 100, sometimes 300...

rfrenchseti (changes requested):
- In table view, when you move the slider, the table jumps down and then back up as it's refreshed. Same thing happens with the scrollbar. This does not happen on the main new_ui branch so something broke here.

rfrenchseti (changes requested):
Not yet done:

**ui/templates/base.html**
- We are not putting the total result count next to the slider, right? Then you can remove the commented-out total obs count field at line 111.

**browse.js**
- Line 667 (now 668) excess spaces around function argument

Also the problem with the table jumping.

juzen2003:
# Comments:
**base.html**
* line 109-110: please convert tabs to spaces

**opus.css**
* line 1183-1220: in hvr-ripple-in area, please put 1 tab = 4 spaces to match the code style

**browse.js**
* line 516: please add a space after "if"

# Questions:
**browse.js**
* line 519: this is just my question, we only use .css when it's like this case that css value dynamically changes with a variable in js. If css value is not related to any variable in js (like line 743), we will prefer not to use .css and use an alternative way that separates css and js code. Is my understanding correct?
* line 1010: there is an extra space after thumbnail-container class, maybe we can remove that space.

**General code style question**
* Just to confirm, I remember that we decide to use "" (instead of '') in js and we will do that when adding new codes, but should we change '' to "" in existing code? Or just change when we see it? I saw Debby already changed lots of '' to "", and there are still many '' in the existing codes. 

# Notes:
* I think using === will be more clear and it will avoid some confusion caused by type coercion in js, like 0 == "" (true) or "77" == 77 (true).

rfrenchseti:
Yes that's exactly that kind of type coercion I was talking about. JavaScript does more hidden type conversion than any other language I've seen, and it can be the source of many bugs. I think using === most of the time is a good idea, makes the code clearer, and avoids hidden problems.

dstopp:
> * line 519: this is just my question, we only use .css when it's like this case that css value dynamically changes with a variable in js. If css value is not related to any variable in js (like line 743), we will prefer not to use .css and use an alternative way that separates css and js code. Is my understanding correct?

Since this is a dynamically modified css, this is appropriate - for static css, we are working to avoid directly setting via css but rather use classes.  There are likely other examples that we can clean up later.

> **General code style question**
> 
> * Just to confirm, I remember that we decide to use "" (instead of '') in js and we will do that when adding new codes, but should we change '' to "" in existing code? Or just change when we see it? I saw Debby already changed lots of '' to "", and there are still many '' in the existing codes.

I change them when i see them or if i'm in that area of code.

rfrenchseti:
There's a problem with your pull request. It appears that your opus.css undoes a lot of the changes that Dave recently did. For example, he had changed `.op-menu-text[space].spinner` to `.op-menu-text.spinner` and your pull undoes this change. He had also done some other cosmetic cleanup that you undo. Could you please recheck your version to make a clean merge?

PR #623, opened 2019-03-31
new ui dave S031 C028 

juzen2003:
# Re-open pull request #619 
# Update:
* More code styles clean-up in opus.css (these are missed from last Friday's update)

### DIFF
--- a/opus/application/apps/ui/templates/base.html
+++ b/opus/application/apps/ui/templates/base.html
@@ -82,7 +82,7 @@
                     <div class="col">
                         <div id="widget-container" class = "widget_column">
                             <!-- note to self:    this is where the widgets will be replace by cards; should no longer by ul/li -->
-                            <ul id="search_widgets" class = ""></ul>
+                            <ul id="op-search-widgets" class = ""></ul>
                         </div>
                     </div>
                 </div>
--- a/opus/application/apps/ui/templates/base.html
+++ b/opus/application/apps/ui/templates/base.html
@@ -82,7 +82,7 @@
                     <div class="col">
                         <div id="widget-container" class = "widget_column">
                             <!-- note to self:    this is where the widgets will be replace by cards; should no longer by ul/li -->
-                            <ul id="op-search-widgets" class = ""></ul>
+                            <ul id="op-search-widgets"></ul>
                         </div>
                     </div>
                 </div>
--- a/opus/application/apps/ui/templates/menu.html
+++ b/opus/application/apps/ui/templates/menu.html
@@ -5,15 +5,15 @@
 
                 {#  Each div has a top-level catgory  #}
           <a href="#submenu-{{ div.table_name }}" data-toggle="collapse" aria-expanded="false" class="dropdown-toggle" data-cat="{{ div.table_name }}">
-              <span class="op-menu-text title">{{ div.label }}
-                  {% if div.table_name == 'obs_general' %}
-                      <span class="spinner">&nbsp;</span>
-                  {% endif %}
+              <span class="title">{{ div.label }}
               </span>
           </a>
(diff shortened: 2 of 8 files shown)


juzen2003:
For further implementation/modification reference, move over Rob's last comments from #619 :
* When the spinner appears next to "General Constraints", it shows up between the text and the expand/collapse arrow. This means that the arrow moves over and then back again when the spinner appears and disappears. I think this is awkward-looking. The spinner should be to the right of the arrow. Ideally I think it would look nicest right-justified so it's sort of out of the way, but do what looks best.

* The behavior on the search page is now the same as the current version, but there's always been something that bothered me about the current version and now seems like a good time to fix it. The current process is:

  * Change the search
  * Put a results spinner and get the new results
  * Once the results count has been returned, query for a new menu and hint numbers and put spinners in those places
  * Once each of those has been returned, remove those spinners

The end result is that when you change the search, you see a single spinner for result count, even though the menu and the hint numbers are wrong and will be reloaded soon. The result count can take a long time to return, so it gives the impression those numbers are correct.

I think the right thing to do is put spinners on the menu and all the widgets with hints the instant the new result count is queried, rather than waiting for the individual API calls to be triggered. That way when you change the search, all the spinners will appear immediately, and then go out one by one as the API calls return.

* When you click on the "Cart" tab, it is displayed as black text on a blue background. However, when you (de)select one of the product types, the word "Cart" is changed to white on a blue background.

juzen2003:
# Update
* rename #search_widgets to #op-search-widgets
* move .op-menu-text spinner to the right of down arrow next to "General Constraints"
* display .op-menu-text and #op-search-widgets spinners once #op-result-count spinner starts spinning
    * remove .op-menu-text spinner when getMenu finishes loading menu.html
    * remove #op-search-widgets spinner when getRangeEndpoints or getValidMults ajax return is done
* update coreui/css/style.css to have .navbar-dark .navbar-nav .nav-link.active:hover,
.navbar-dark .navbar-nav .nav-link.active:focus with color: #FFF (this is to remove our navbar text color conflict in line 4357 and 11239)

rfrenchseti:
What is the point of changing “.op-menu-text .spinner” to “.op-menu-text.spinner”? I don’t know jQuery well but I thought the former was the normal was to select a child.

dstopp (changes requested):
base.html line 85 - remove empty class
menu.html line 11 - i think you can condense to a single if block
style.css - you are changing a template file, not an opus file, so if we upgrade to a diff version we will lose your changes.  Suggestion that you make those changes in opus.css w/a note, if need be.  we should discuss.

juzen2003:
# Update:
* base.html: remove empty class in line 85
* menu.html: condense to one if block
* move .navbar-dark .navbar-nav .nav-link.active:hover, .navbar-dark .navbar-nav .nav-link.active:focus with color: #FFF to opus.css 

# Note:
* jquery selector “.op-menu-text.spinner” is to select the element with both "op-menu-text" and "spinner" classes. To have spinner display at the right of down arrow, we have to move the spinner element out of "a" tag. And "op-menu-text" class is moved from the original "span" tag inside "a" tag to the "span" tag with "spinner" class. That's why jquery selector is changed to “.op-menu-text.spinner”.

rfrenchseti:
So if .spinner is always on the same element as .op-menu-text is there a reason to include both?

juzen2003:
.spinner is a shared class for multiple spinners (they are using the same background image), and the purpose of adding another class to the same element (like this case) or adding a class to parent container (like .op-total-size .spinner) is to customize the position of each spinner with margin, line height, etc to make it look better.

rfrenchseti:
Oh never mind. I see what you're doing. Thanks.

PR #626, opened 2019-04-02
Fix broken qtypes. Set proper qtype when coming in from permalink. Fi…

rfrenchseti:
It turns out that qtypes were completely broken. Changing a qtype in the dropdown box didn't actually change the hash. This was the result of an internal Django change I made awhile back and I didn't notice how it affected the UI.

This set of changes both fixes the fundamental working of qtypes, and handles qtypes coming in from permalinks. Previously these incoming qtypes were just ignored and the default was used.

Note there is no clean way to select a default item for a dropdown in Django 2.x, so this is a horrible hack. In fact ALL of the code in api_get_widget() and search/forms.py is horrible and needs to be rewritten from scratch. That's a project for another day.

### DIFF
--- a/opus/application/apps/ui/views.py
+++ b/opus/application/apps/ui/views.py
@@ -140,13 +140,20 @@ def api_get_widget(request, **kwargs):
             selections = {}
             extras = {}
 
+    param_qualified_name_no_num = strip_numeric_suffix(param_qualified_name)
+
+    initial_qtype = None
+    if 'qtypes' in extras:
+        qtypes = extras['qtypes']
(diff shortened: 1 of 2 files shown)


dstopp (commented):
search.js, line 251 -  I believe we are making a move to using the inclusive style technique w/ `` and ${}, ie., $(`#widget__${slug_no_num} select`)
Also note that js internal vars are typically camelCase, ie, slugNoNum <or preferably slugNo or slugNum>

rfrenchseti:
I didn’t write that code, but I can upgrade it.

PR #627, opened 2019-04-03
Modify base.html to remove spinner in hamburger menu

juzen2003:
# Update:
* Remove spinners in hamburger menu


rfrenchseti:
Not sure why the spinners were there in the first place, but I’m wondering why they were triggered now and had been invisible before. Was this related to the change to start the search spinners immediately after a search change? If so does that mean that code isn’t specific enough about which spinners to activate? I don’t want future spinners to pop up when we don’t expect them.

juzen2003:
Those spinners exist previously in html, and it's hidden by this css style (which was cleaned up last Friday):
`#op-obs-menu .spinner {
    display:none;
    margin:0px 10px;
    width:30px;
}`
I forgot to remove them from the html structure but only cleaned up css codes. They have nothing to do with the spinners triggered at the beginning of a search. For reference, these are the spinners triggered when a search is started:
* result count spinner 
* spinners next to the title of widgets 
* general constraints spinner

PR #630, opened 2019-04-03
Code cleanup from JSHint

rfrenchseti:
- Missing semicolons
- Remove unused variables
- Fix up undefined variables and convert var to let as needed
- Fix bugs found
- Cleanup formatting


dstopp (changes requested):
-where appropriate as we are changing the code, consider replacing "==" with "==="
-detail.js, line 44.  String should be OPUS_ID, not opusId

dstopp (approved):

PR #631, opened 2019-04-05
New ui dave a029

juzen2003:
# This is a dummy pull request to track the update on MutationObserver
# Update:
* A029:
    * search page:
        * Add mutationObserver to detect changes in search sidebar and update ps
            * Test cases:
                1. Change browser size in other tabs and switch to search, ps will get updated
                2. If left pane is just large enough to hold the contents without scrollbar, by clicking any dropdown arrow to expand a new menu, ps will show up (get updated). Undo previous action, ps will disappear.
                3. If left pane is just large enough to hold the contents without scrollbar, by clicking any input in widgets that adds a new item in the left pane, ps will show up (get updated). Undo previous action, ps will disappear.
                4. collapse/expand widgets, ps will get updated.
        * Add mutationObserver to detect changes in search widgets and update ps
            * Test cases:
                1. Add widgets until the last one is just off the bottom of the browser, ps will show up
                2. Change browser size in other tabs and switch to search, ps will get updated  
                3. Add/Remove/collapse/expand widgets, ps will get updated.

# Note:
* For now only two ps in search page are moved to MutationObserver.
* Currently update calls for those two ps are all at line 453-535 in opus.js. Will have to create a function for all codes related to MutationObserver eventually  
* All adjustSearchSideBarHeight, adjustSearchWidgetHeight, adjustSearchHeight scattered in other files are commented out.
* Reference: 
  * https://www.javascripture.com/MutationObserver
  * https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/MutationObserver
  * Online code editor for experiment:
    * http://cssdeck.com/labs/badnvznx


juzen2003:
# Update test cases:
* Search sidebar ps:
  * Test cases:
                1. Change browser size in other tabs and switch to search, ps will get updated
                2. If left pane is just large enough to hold the contents without scrollbar, by clicking any dropdown arrow to expand a new menu, ps will show up (get updated). Undo previous action, ps will disappear.
                3. If left pane is just large enough to hold the contents without scrollbar, by clicking any input in widgets that adds a new item in the left pane, ps will show up (get updated). Undo previous action, ps will disappear.
                4. collapse/expand submenu, ps will get updated.
* Search widget ps:
  * Test cases:
                1. Add widgets until the last one is just off the bottom of the browser, ps will show up
                2. Change browser size in other tabs and switch to search, ps will get updated  
                3. Add/Remove/collapse/expand widgets, ps will get updated.
                4. Open any mult_group dropdown menu inside widgets (like: intended target name), ps will get updated

juzen2003:
# Update:
* A029:
    * search page:
        * Add mutationObserver to detect changes in search sidebar and update ps (o_search.searchScrollbar)
            * Test cases:
                - Change browser size in other tabs and switch to search, ps will get updated.
                - If left pane is just large enough to hold the contents without scrollbar, by clicking any dropdown arrow to expand a new menu, ps will show up (get updated). Undo previous action, ps will disappear.
                - If left pane is just large enough to hold the contents without scrollbar, by clicking any input in widgets that adds a new item in the left pane, ps will show up (get updated). Undo previous action, ps will disappear.
                - collapse/expand submenu, ps will get updated.
        * Add mutationObserver to detect changes in search widgets and update ps (o_search.widgetScrollbar)
            * Test cases:
                - Add widgets until the last one is just off the bottom of the browser, ps will show up.
                - Change browser size in other tabs and switch to search, ps will get updated.  
                - Add/Remove/collapse/expand widgets, ps will get updated.
                - Open any mult_group dropdown menu inside widgets (like: intended target name), ps will get updated.
    * cart page:
        * Add mutationObserver to detect changes in cart page and update ps (o_cart.downloadOptionsScrollbar and o_cart.cartGalleryScrollbar)
            * Test cases:
                - Add data/url archive links, ps on left pane will get updated.
                - Change browser size in other tabs and switch to cart, ps will get updated.
                *  Note: ps in gallery view is having incorrect infiniteScroll so didn't fully test it. But ps in download options has been verified with above 2 test cases.
    * help panel:
        * Add mutationObserver to detect changes in help menu and update ps (opus.helpScrollbar)
            * Test cases:
                - continuously open and close each item in help panel, make sure ps get updated correctly.
    * browse page:
        * Add mutationObserver to detect changes in browse dialog and update ps (o_browse.modalScrollbar)
            * Test cases:
                - y-scrollbar will get updated when modal window is open with many columns (metadata) for the first time.
        * Add mutationObserver to detect changes in add metadata modal and update ps (o_browse.allMetadataScrollbar and o_browse.selectedMetadataScrollbar)
            * Test cases:
                - o_browse.selectedMetadataScrollbar (ps in right pane)
                    - Add/remove metadata, ps in the right pane will get updated.
                    - If right pane has many metadata items, ps will get updated when select metadata modal is open.
                    - Click reset to default, ps will get updated.
                    - Click to select metadata in the left pane, ps in the right pane will get updated.
                - o_browse.allMetadataScrollbar (ps in left pane)
                    - Expand/collapse menu list, ps will get updated.
        * Add mutationObserver to detect changes in table view and update ps (o_browse.tableScrollbar)
            * Test cases:
                - Add many columns, ps will get updated.
                - After adding many columns, reset to default columns, ps will get updated.
                - prefetch more data in gallery view and switch back to table view, ps will get updated.
        * Add mutationObserver to detect changes in gallery view and update ps (o_browse.galleryScrollbar)
            * Test cases:
                - prefetch more data in table view and switch back to gallery view, ps will get updated.
    * detail page:
        * We don't need to attached mutationObserver to detail page because the ps is initialized (and update) when all info are returned from multiple ajax calls, and every time we switch to detail page, getDetail will get called and render the html again (re-init ps as well).

juzen2003:
**NOTE**
S001 shows up after the commit (6cbc40ae22ce64068279c6a7efd6b2a9276c9ce1) of merging with upstream/new_ui. If we checkout to the commit (64ad9de7e4d05ba431bf8a40bff03ba307919938) right before "Merge with upstream/new_ui", S001 will be gone.

rfrenchseti:
Interesting about S001. I couldn't get it to reproduce before. Can you give an exact set of actions that make it happen?

juzen2003:
**Here is my test step:**
- Checkout to the latest commit with issue: 28a6eb79c0f85e153c9dfad68975fe146edec15e from this pull request or the latest commit from pds-opus/new_ui
- Clear browse cache, open up browser and run python manage.py runserver
- In search planet widget:
  - select Neptune --> url gets updated --> search is triggered --> result count is blinking and changes to 792 
  - unselect Neptune --> url gets updated --> search is NOT triggered --> result count is not blinking and stays at 792
  - select Pluto --> url gets updated --> search is triggered --> result count is blinking and changes to  6773
  - unselect Pluto --> url gets updated --> search is NOT triggered --> result count is not blinking and stays 6773

juzen2003:
For S001, it’s caused by the return data type of selections from getSelectionsFromHash function in hash.js
Previously we have:
` if (!$.isEmptyObject(selections)) {
            return selections;
}`
It will return undefined when there is no selection, but now it's been modified to return an empty object {} when there is no selection.

So for load function in opus.js, if we have selections = {}, this piece of code will not trigger search for the last unselect:
```js
        if ($.isEmptyObject(selections)) {
            // there are no selections found in the url hash
            if (!$.isEmptyObject(opus.last_selections)) {
                // last selections is also empty
                opus.last_selections = {};
                o_browse.resetQuery();
            }
        }
        
        if (o_utils.areObjectsEqual(selections, opus.last_selections))  {
            // selections have not changed from opus.last_selections
            if (!opus.force_load) { // so we do only non-reloading pref changes
                return;
            }
            opus.force_load = false;

        }
```
Both selections and opus.last_selections are {}, and opus.force_load is false, so for the last unselect action, it will also return, instead of triggering search.

Previously when selections is undefined and opus.last_selections is {}, o_utils.areObjectsEqual(selections, opus.last_selections) will be false, and trigger search for the last unselect action. 

So we can just put opus.force_load = true; here, and issue will be solved:
```js
        if ($.isEmptyObject(selections)) {
            // there are no selections found in the url hash
            if (!$.isEmptyObject(opus.last_selections)) {
                // last selections is also empty
                opus.last_selections = {};
                o_browse.resetQuery();
                opus.force_load = true;
            }
        }
```

rfrenchseti:
I find this code very confusing. Why is there a comment that says "last selections is also empty" right after an if that checked that it *wasn't* empty? Also, this sets force_load to true every time this function is called with selections being empty. Is it ever possible to call it multiple times in a row while selections continues to be empty? It seems like in that case it would be forcing reload after reload.

dstopp (changes requested):
- i think we can remove the commented out code
- try using $( "#foo" )[ 0 ]; to get DOM element per conversation to avoid adding all the id's
- move observer code to its own namespace
- opus.js - I prefer switch to if then else then if because it feels easier to read.

rfrenchseti:
There appear to be two spaces before all the "new MutationObserver" calls.

juzen2003:
Regarding to Rob's question:
That's the comment from old code and it's confusing. What it actually means is if our current selections is empty, update the last selections to empty object as well. (normally we update last selections later at line 142). And opus.force_load will be set to true in line 118 and update to false in line 127. And then it will proceed to reload for that time with result count updated. But in the next loop of load function call (if the selections is still empty), opus.force_load will stays false because last selection is still empty. And load function will return in line 125. So it will not be forcing reload after reload.

rfrenchseti:
OK. That whole logic seems like it's done backwards. Shouldn't we be testing the equivalence of selections and last_selections BEFORE overwriting last_selections? Perhaps you could go ahead and clean up the flow and the comments while you're in this piece of code.

rfrenchseti:
Is there a way to make a minimum height on the scrollbar? If you have a Detail page with a lot of data on it, and make the browser pretty short, the scrollbar goes down to about 1 pixel tall.

juzen2003:
Yes, we can set it when initializing ps. Like what we did for table scrollbar. (minScrollbarLength: 30)

rfrenchseti:
OK, you probably want to do that on all of them. I can get the same effect on the Cart page with the product types.

rfrenchseti:
I did a first pass through all the scrollbars, changing the data and the browser size, and they seem to work as I would expect.

juzen2003:
Got it, I will set all ps thumbnail minimum length to 30

juzen2003:
# Update:
* A031:
    * put back following 2 spinners (".op-menu-text.spinner" and "#op-search-widgets .spinner") right after result count spinner is triggered:
    ```js
    // start the result count spinner and do the yellow flash
    $("#op-result-count").html(opus.spinner).parent().effect("highlight", {}, 500);
    $("#op-observation-number").html(opus.spinner).effect("highlight", {}, 500);

    // start op-menu-text and op-search-widgets spinner
    // this is to trigger these two spinners right away when result count spinner is running
    $(".op-menu-text.spinner").addClass("op-show-spinner");
    $("#op-search-widgets .spinner").fadeIn();
    ```
    * For S031, I think the current new_ui doesn’t have “.op-menu-text.spinner” and “#op-search-widgets .spinner” triggered right after result count spinner is triggered. That piece of code is missing in opus.js load function (it should be there from previous merged pull-request). I put them back in new_ui_dave_A029.
* PerfectScrollbar:
    * Set minScrollbarLength to 30 to all ps
* A029:
    * Remove all newly added id, replace getElementById(selector) with $(selecto)[0]
    * Move codes related to mutationObserver to mutationObserver.js
* S001:
    * load function in opus.js:
        * removed unwanted logic block. Compare selections and last selections first, and update the last selections later on.

opus/application/static_media/js/opus.js
```diff
@@ -390,7 +386,7 @@ var opus = {
             $("#op-help-panel .card-body").scrollTop(0);
             opus.helpScrollbar.update();
         }
-    }
+    },
```
rfrenchseti: Does this comma serve a purpose?

opus/application/static_media/js/opus.js
```diff
@@ -313,11 +312,8 @@ var opus = {
             deferredArr.push($.Deferred());
             o_widgets.getWidget(slug,"#search_widgets",deferredArr[index]);
         });
-        // wait until all widgets are get
-        $.when.apply(null, deferredArr).then(function() {
-            let adjustSearchWidgetHeight = _.debounce(o_search.adjustSearchHeight, 800);
-            adjustSearchWidgetHeight();
-        });
+        // wait until all widgets are get, do somthing here:
```
rfrenchseti: Can we just get rid of these lines?

opus/application/static_media/js/browse.js
```diff
@@ -1440,6 +1435,54 @@ var o_browse = {
         o_browse.tableScrollbar.update();
     },
 
+    adjustmetadataModalMenu: function() {
```
rfrenchseti: What is a metadataModalMenu? It's not obvious to me what menu we're talking about.

rfrenchseti (changes requested):

rfrenchseti:
Everything seems to work as expected in my testing.

opus/application/static_media/js/opus.js
juzen2003: It's removed now. It was there because there was another function down there in the same namespace.

opus/application/static_media/js/opus.js
juzen2003: Commented out codes are removed.

opus/application/static_media/js/browse.js
juzen2003: Rename it to adjustMetadataSelectorMenu

dstopp (changes requested):
- i think we can remove the commented out code
- try using $( "#foo" )[ 0 ]; to get DOM element per conversation to avoid adding all the id's
- move observer code to its own namespace
- opus.js - I prefer switch to if then else then if because it feels easier to read.

-in the instantiation of the scrollbars, are tableScrollbar & galleryScrollbar created to be generic for both browse and cart?

-do we want to set the minScrollbarLength of 30 as a global const and use it throughout the file instead of '30'? <probably should live in opus.js since it's used all over>

-how does the type of functionality of 'adjustMetadataSelectorMenu' diff from 'adjustSelectedMetadataPS' ?  if they are both adjusting the scrollbar, might want to name similar for consistency.

-detail.js - why is the the scrollbar initialized everytime?  can't we init it like we do in browse and cart just once?

-don't we need to specify the namespace <browse vs cart> for the scrollbar?  

-in opus.js, why did you remove the if ($.isEmptyObject..) at line 112?
-in opus.js, if you don't change the lines at 122-126, we will avoid a conflict when we merge <changes are not necessary>

rfrenchseti:
He changed opus.js because I asked him to. See the discussion we had earlier in this pull request. The logic that was there didn’t make much sense and was the cause of the search bug S001. I think the new logic is much cleaner and straightforward and less prone to future bugs. I’m sure we can deal with the merge conflict.

rfrenchseti:
I was having trouble with adjustMetadataSelectorMenu too because it sounds like it’s adjusting a menu not a scroll bar. I assume it would make sense to name it adjustMetadataSelectorMenuPS?

juzen2003:
# Update based on comments:
-in the instantiation of the scrollbars, are tableScrollbar & galleryScrollbar created to be generic for both browse and cart?
    * They are not generic for both browse and cart. For cart, ps is initialized when view.html is done loading.

-do we want to set the minScrollbarLength of 30 as a global const and use it throughout the file instead of '30'? <probably should live in opus.js since it's used all over>
    * Now opus.minimumPSLength is created (set to 30) in opus.js.

-how does the type of functionality of 'adjustMetadataSelectorMenu' diff from 'adjustSelectedMetadataPS' ? if they are both adjusting the scrollbar, might want to name similar for consistency.
    * Rename "adjustMetadataSelectorMenu" to "adjustMetadataSelectorMenuPS".

-detail.js - why is the the scrollbar initialized everytime? can't we init it like we do in browse and cart just once?
    * Because we re-render html when switching to detail page. We have to initialize ps every time the html is rendered.

-don't we need to specify the namespace for the scrollbar?
    * Do you mean rename "opus_mutationObserver" to something like "opus_mutationObserverForPS"?

dstopp:
"-in the instantiation of the scrollbars, are tableScrollbar & galleryScrollbar created to be generic for both browse and cart?
* They are not generic for both browse and cart. For cart, ps is initialized when view.html is done loading."
I believe there needs to be namepsace specification to differentiate between browse and cart - I guess that's specific to what i was referring to.


"-detail.js - why is the the scrollbar initialized everytime? can't we init it like we do in browse and cart just once?
* Because we re-render html when switching to detail page. We have to initialize ps every time the html is rendered."
I believe it can be done exactly the same way as browse and search page;  You will need to update the scrollbar, but i don't believe it needs to be instantiated over and over. 


"-don't we need to specify the namespace for the scrollbar?
* Do you mean rename "opus_mutationObserver" to something like "opus_mutationObserverForPS"?"

No, i am referring to #browse and #cart.

rfrenchseti:
I don’t understand why you think you can use the same PS for the Detail page each time. For the metadata selector modal, which also gets rendered fresh on occasion, you create two new scrollbars each time rather than reusing them. Can you be more specific about what you want Dave to do? I don’t want him spending time hunting for something that doesn’t exist.

rfrenchseti:
I think you two are talking about different things regarding the gallery scrollbars. I think only the product info scrollbar is created by view.html, right? Not the right side gallery. So do the browse gallery and cart gallery use the same scrollbar and it’s just reset depending on which tab is active? Or does each gallery have its own PS? Right now it looks like there is a single gallery PS. Is it ok being generic or do you need two PS with a #cart or #browse identifier?

rfrenchseti (approved):

rfrenchseti:
Debby and I worked through the confusion about the browse/cart PS. Here are the issues (some of which existed before the mutationObserver changes, but should still be fixed).

browse.js line 8:
```
    galleryScrollbar: new PerfectScrollbar(".gallery-contents", {
        suppressScrollX: true,
        minScrollbarLength: opus.minimumPSLength
    }),
```

The ".gallery-contents" actually ends up selecting two DOM elements - one for the cart and one for the browse. It's just dumb luck that the PS ends up getting created for the browse version and the cart gets ignored, so this code actually works. But it would be much better to put a "#browse" in there.

The same is true of:
```
    tableScrollbar: new PerfectScrollbar(".dataTable", {
        minScrollbarLength: opus.minimumPSLength
    }),
```
even though there isn't a table in the cart yet, there will be in the future.

Once those are fixed, we have a separate problem with the cart.

cart.js line 222:
```
                    if (!o_cart.cartGalleryScrollbar) {
                        o_cart.cartGalleryScrollbar = new PerfectScrollbar("#cart .gallery-contents", {
                            suppressScrollX: true,
                            minScrollbarLength: opus.minimumPSLength
                        });
                    }
```
This creates a new PS for the cart (with proper selector). However, as far as we can tell this PS is not put under the control of a mutationObserver. I assume it should be.

Finally, to be consistent with other modules, it would be better to call the whole thing "o_mutationObserver" instead of "opus_mutationObserver".

juzen2003:
#Update:
 - rename opus_mutationObserver to o_mutationObserver
 - Add #browse to be more specific when initializing tableScrollbar and galleryScrollbar in browse page
 - Add comments to initAndUpdatePerfectScrollbar in detail.js
 - Update comments in mutationObserver for cart tab

dstopp (approved):

PR #633, opened 2019-04-09
First pass at new tooltips

rfrenchseti:
- Param info now has extra optional fields for dict_name_results and dict_context_results if we want a separate tooltip for the detail page.
- There are definitions and definitions_results in the table_schema JSON files as necessary.
- To avoid namespace conflicts there is a new OPUS_GENERAL_SEARCH context that behaves identically to OPUS_GENERAL. It is only used when necessary.
- The Help/Datasets page has been renamed to Help/Data Volumes and the API has been renamed accordingly.
- The Volumes page now displays the full name of the instrument (Cassini ISS) instead of just the abbreviation (COISS) above the list of volumes.
- To test param_info and the dictionary must be re-imported:
`python main_opus_import.py --create-param-info --import-dictionary`
- Currently there are new definitions for General, PDS, Image, and Surface Geometry Target.


dstopp (approved):

rfrenchseti:
So when I moved opus_id to PDS Constraints and made it searchable, I broke all sorts of things. These are now fixed. There were also problems with handling of ringobsid in normalizeurl, and these are fixed and new tests are added. All tests pass and code coverage of normalizeurl is once again at 100%.

PR #635, opened 2019-04-11
added a lock for the click to wait for the fadein/fadeout to complete.

dstopp:
I added a lock that clears when the fadein/fadeout is complete.

### DIFF
--- a/opus/application/static_media/css/opus.css
+++ b/opus/application/static_media/css/opus.css
@@ -437,7 +437,7 @@ th.sticky-header {
 }
 
 #metadataSelector i.fa-check {
-    display: none;
+    display: inline-block;
 }
 
 #metadataSelector ul li {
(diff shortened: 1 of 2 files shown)


rfrenchseti:
I don't really like this fix. It's nice to be able to just run down the list of metadata and click them in rapid succession to add them. But now you have to wait for the animation to finish in each case so the whole thing slows to a crawl.

rfrenchseti (changes requested):
Please see if you can find some way to do it that doesn't involve blocking user clicks on multiple entries.

dstopp:
I modified the code so that it keeps track of the one you clicked on so it will only prevent you from clicking on same one again, not a diff one.

rfrenchseti (changes requested):
I’m confused about this whole process. When you click on a metadata item for the first time, it isn’t :visible and it instantly gets inserted into cols while fading in. I assume the fade in takes time and :visible isn’t set until it’s done. So why not just check directly if the slug is in cols instead of checking :visible? Then there can’t be a race condition. In other words right now you’re letting the :visible state be the primary info about if a slug is selected. It seems a lot cleaned to let cols be the primary info. 

I think your new fix still has the same basic problem. Now that you store a single slug, I could conceivably click fast alternating between two items and still get a bad double click on one of them while the animation is going.

PR #637, opened 2019-04-15
Add JSHINT directives. Minor code cleanup.

rfrenchseti:
JSHINT is now clean for all code (except a few places where it complains about things that are actually correct). There are no more "var" statements except for top-level namespaces. Directives have been added to the top of each file to make JSHINT behave in the future. Simply copy-paste the code into jshint.com and look for warnings or errors.

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -31,7 +31,6 @@ var o_browse = {
 
     galleryScrollTo: 0,
     metadataSelectorDrawn: false,
-    currentSelectedMetadata: undefined,
 
     lastLoadDataRequestNo: 0,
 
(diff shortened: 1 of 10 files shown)


opus/application/static_media/js/browse.js
```diff
@@ -20,6 +31,7 @@ var o_browse = {
 
     galleryScrollTo: 0,
     metadataSelectorDrawn: false,
+    currentSelectedMetadata: undefined,
```
dstopp: this var is only global w/in the confines of the function; it can stay where it was, maybe a diff comment would be more appropriate.

dstopp (approved):

opus/application/static_media/js/opus.js
```diff
@@ -406,20 +420,20 @@ $(document).ready(function() {
...
+    let adjustDetailHeightDB = _.debounce(o_detail.adjustDetailHeight, 200);
+    let adjustHelpPanelHeightDB = _.debounce(opus.adjustHelpPanelHeight, 200);
+
+    $(window).on("resize", function() {
+        adjustSearchHeightDB();
+        adjustBrowseHeightDB();
+        adjustTableSizeDB();
+        adjustProductInfoHeightDB();
+        adjustDetailHeightDB();
+        adjustHelpPanelHeightDB();
```
juzen2003, line 436: Nothing wrong, just curious what is DB in these variables?

juzen2003 (approved):

PR #638, opened 2019-04-16
New ui dave update url

juzen2003:
# Update:
* Normalized URL logic:
    - In document.ready, normalized url api will be called and update the url (update window.location.hash)
        - if there is a message returned from api, create a "Update URL" link in top navbar for user to click.
            - The returned message will be displayed in the modal when user clicks "Update URL".
            - If user clicks "OK", "Update URL" link will disappear.
            - If user clicks "X" on the top right corner, "Update URL will still exist".
            (Please let me know if we want to change the behavior)
    - Once url is updated, we start opus timer to check load function every second.
        - Note: after above steps, all actions done in opus will always have new slugs updated in url

# Test cases:
* Test 1:
    - Open a new tab and paste in: http://localhost:8000/#/missionid=Cassini&view=search&browse=gallery&colls_browse=gallery&page=1&gallery_data_viewer=true&limit=100&order=time1&cols=opusid,instrument,planet,target,time1,observationduration&widgets=missionid,planet,target&widgets2=&detail=
    - URL is updated to: http://localhost:8000/#/mission=Cassini&cols=opusid,instrument,planet,target,time1,observationduration&widgets=mission,planet,target&order=time1,opusid&view=search&browse=gallery&startobs=1&cart_browse=gallery&detail=
    - "Update URL" shows up on navbar, and a modal with messages will show up when we click it.
        - click "OK", "Update URL" will be removed.
        - click "X", "Update URL" will stay.
* Test 2:
    - Open a new tab and paste in: http://localhost:8000/opus/#/cols=opusid,instrument,planet,target,time1,observationduration&widgets=planet,target&order=time1,opusid&view=search&browse=gallery&startobs=1&cart_browse=gallery&detail=
    - URL will not change, and no "Update URL".
* Test 3:
    - Open a new tab and paste in: http://localhost:8000/#/plnet=neptune,mars&view=search&browse=gallery&cart_browse=gallery&page=1&limit=100&order=time1,opusid&cols=opusid,instrument,planet,target,time1,observationduration&widgets=planet,target&detail=
    - URL is updated to: http://localhost:8000/#/cols=opusid,instrument,planet,target,time1,observationduration&widgets=planet,target&order=time1,opusid&view=search&browse=gallery&startobs=1&cart_browse=gallery&detail=
    - "Update URL" shows up on navbar, and a modal with messages will show up when we click it.
        - click "OK", "Update URL" will be removed.
        - click "X", "Update URL" will stay.
* Test 4:
    - In search page, paste in: http://localhost:8000/#/view=detail&browse=gallery&cart_browse=gallery&page=1&limit=100&order=time1,opusid&cols=opusid,instrument,planet,target,time1,observationduration&widgets=planet,target&detail=vg-iss-2-s-c4360004&startobs=1
    - URL will not change, it will only re-organize the order of slugs to the same as normalized url.

* Test 5:
    - In search page, paste in: http://localhost:8000/#/view=detail&browse=galleryx&cart_browse=gallery&page=1&limit=100&order=time1,opusid&cols=opusid,instrument,planet,target,time1,observationduration&widgets=planet,target&detail=vg-iss-2-s-c4360004&startobs=1
    - URL is updated to: http://localhost:8000/#/cols=opusid,instrument,planet,target,time1,observationduration&widgets=planet,target&order=time1,opusid&view=detail&browse=gallery&startobs=1&cart_browse=gallery&detail=vg-iss-2-s-c4360004
    - "Update URL" shows up on navbar, and a modal with messages will show up when we click it.
        - click "OK", "Update URL" will be removed.
        - click "X", "Update URL" will stay.


opus/application/apps/ui/templates/base.html
```diff
@@ -44,6 +44,9 @@
                             </div>
                         </a>
                     </li>
+                    <li class="nav-item">
+                        <a class="btn op-update-url-msg nav-link" data-toggle="modal" data-target="#op-update-url" title="Update to normalized URL">Update URL</a>
```
rfrenchseti: "Update URL" isn't very meaningful to the user. I think this should be a generic capability that allows messages to be sent to the user; then we can use it for other things in the future. It should also be eye-catching. I suggest putting some kind of icon (like the fancy "New" icon we put next to the "Recent Announcements", but different) in yellow, and the nav should read something like "Warning". The tooltip should be "Click to read an important message". We can discuss exact wording and icon with Debby.

opus/application/apps/ui/templates/confirm_modal.html
```diff
@@ -12,8 +12,12 @@
         {{confirm_msg}}
       </div>
       <div class="modal-footer">
-        <button type="submit" class="btn btn-primary" data-dismiss="modal" data-target={{target}}>Yes</button>
-        <button type="cancel" class="btn btn-danger">No</button>
+        {% if confirm_title == "Update URL" %}
```
rfrenchseti: Putting the specific dialog box name here removes some of the ability to make this a generic feature. Instead of that, have parameters that read something like "yes_text" and "no_text". If either is empty, then don't display that button. If there is text there, then use that as the button contents.

opus/application/apps/ui/views.py
```diff
@@ -1079,9 +1079,12 @@ def _escape_or_label_results(old_slug, pi):
     if final_msg == '':
         final_msg = None
 
+    reqno = get_reqno(request)
+
```
rfrenchseti: Generally we've been either requiring reqno to be present, or requiring it to be absent, so we don't have cases where we accidentally leave it off and then it causes a bug. The general code looks like this:
```
    if reqno is None:
        log.error('api_cart_status: Missing or badly formatted reqno')
        ret = Http404(settings.HTTP404_MISSING_REQNO)
        exit_api_call(api_code, ret)
        raise ret
```
But if we put that code here, we also need to update all the normalizeurl tests in test_ui_api.py to include a reqno. So we should really decide now, if we can, if the reqno is needed.

opus/application/static_media/js/hash.js
```diff
@@ -47,14 +47,19 @@ var o_hash = {
                         // opus.prefs.view = search, browse, cart, or detail
                         // opus.prefs.browse =  'gallery' or 'dataTable',
                         let page = o_browse.getCurrentPage();
-                        hash.push("page=" + page);
+                        // we will no longer have page in url, so no need to push page
```
rfrenchseti: As discussed, we need to leave page in there for now. Once Debby integrates this onto her startobs branch, this case should be removed entirely because page is no longer a valid slug.

opus/application/static_media/js/opus.js
```diff
@@ -37,22 +37,24 @@ var opus = {
     lastRequestNo: 0,          // holds request numbers for main result count loop,
     lastAllNormalizeRequestNo: 0,
     lastResultCountRequestNo: 0,
+    lastNormalizeurlRequestNo: 0,
```
rfrenchseti: Let's capitalize URL to be consistent.

opus/application/static_media/js/opus.js
```diff
@@ -37,22 +37,24 @@ var opus = {
...
-
-
+    // prefs key:value pair order has been re-organized to match up with normalized url
+    prefs: {
+        "cols": default_columns.split(","),  // default result table columns by slug
+        "widgets": [], // search tab widget columns
+        "order": default_sort_order.split(","),  // result table ordering
+        "view": "search", // search, browse, cart, detail
+        "browse": "gallery", //either 'gallery' or 'data'
+        "startobs": 1,
```
rfrenchseti: Debby: Is there now going to be a cart_startobs, so that the main gallery and the cart page can have different scroll positions saved? If so we need to add it to a bunch of places in the Django.

opus/application/static_media/js/hash.js
```diff
@@ -47,14 +47,19 @@ var o_hash = {
...
+                        // hash.push("page=" + page);
+                        break;
+                    case "startobs":
+                        hash.push("startobs=" + opus.currentObs);
```
rfrenchseti: Likewise we should comment out this code (and the definition of currentObs) until merged with the startobs branch.

opus/application/static_media/js/opus.js
```diff
@@ -135,11 +139,37 @@ var opus = {
...
+            // console.log(selections);
+            // console.log("=== modifiedSelections ===")
+            // console.log(modifiedSelections);
+            // console.log("=== opus.selections ===");
+            // console.log(opus.selections);
+            // console.log("Are modifiedSelections and opus.selections the same ?");
+            // console.log(opus.checkIfSelectionsObjectsAreTheSame(modifiedSelections, opus.selections));
+
+            // if data in selections !== data in opus.selections, it means selections are modified manually in url, reload the page (modified url in url bar and hit enter)
+            if (!opus.checkIfSelectionsObjectsAreTheSame(modifiedSelections, opus.selections)) {
```
rfrenchseti: Generally when a function returns a True/False value, you don't want to call it something that starts with a verb. "Check If" means perform some kind of action, but doesn't tell you that it's going to return a True/False value based on the result. Instead, use a wording like "(!opus.selectionsObjectsAreTheSame...)". Then it makes sense when you read it as "if not selections objects are the same".

opus/application/static_media/js/opus.js
```diff
@@ -135,11 +139,37 @@ var opus = {
...
-            // reset the pages: {"gallery":1, "data":1, "cart_gallery":1, "cart_data":1 };
+            // reset the pages:
             opus.prefs.page = default_pages;
 
-            // and reset the query:
-            o_browse.resetQuery();
+            // create an object from selections to compare with opus.selections
+            let modifiedSelections = {};
+            $.each(Object.keys(selections), function(idx, slug) {
+                // qtype is not updated in opus.selections
```
rfrenchseti: I don't understand this code or this comment. What does it mean for qtype to not "be updated"?

opus/application/static_media/js/opus.js
```diff
@@ -159,6 +189,89 @@ var opus = {
         o_search.allNormalizedApiCall().then(opus.getResultCount).then(opus.updatePageAfterResultCountAPI);
     },
 
+    // If the pasted url has selections, init opus.selections & opus.last_selections to the current hash's selections.
+    // This is like we init both opus.selections & opus.last_selections to {} when starting opus fresh.
+    InitSelections: function() {
+        let selections = o_hash.getSelectionsFromHash();
+        let modifiedSelections = {};
+        $.each(Object.keys(selections), function(idx, slug) {
+            // qtype is not updated in opus.selections
```
rfrenchseti: Same question about this code as above.

opus/application/static_media/js/opus.js
```diff
@@ -159,6 +189,89 @@ var opus = {
...
+    // These two cases: obj1 & obj2 are the same:
+    // obj1 (convert from selections) = {planet: [mars, netpune], mission: [hubble]}
+    // obj2 (opus.selections) = {planet: [neptune, mars], mission: [hubble]}
+    // OR
+    // obj1 (convert from selections) = {planet: [mars, netpune]}
+    // obj2 (opus.selections) = {planet: [neptune, mars], mission: []}
+    // This is used to check if selections are changed manually by user or changed by actions done in opus
+    // Selections and opus.selections are the same when user makes selections in opus (url gets updated as well).
+    // If selections and opus.selections are not the same, it means the url is changed manually by user
+    checkIfSelectionsObjectsAreTheSame: function(obj1, obj2) {
```
rfrenchseti: This entire function can't work at all.
- Two of the three return statements don't return a value, so the function will return "undefined" instead.
- The variable isTheSame isn't actually necessary, because you can always just "return True" or "return False".
- The loop doesn't catch the case of there being slugs in obj2 that aren't in obj1.
- I suggest you start with a simple slug list comparison...get the sorted list of keys from obj1 and obj2 and compare the two lists for equality. If they're not the same, then one of the two lists has a slug not in the other and you can return False.
- Otherwise, the slug list is guaranteed to be the same, so just iterate through and compared their values.

opus/application/static_media/js/opus.js
```diff
@@ -159,6 +189,89 @@ var opus = {
...
+        // Because in our implementation, this api is call during document ready (or when reload), and every time this event is triggered, it means everything is reloaded so reqno will always be starting from 0 and passed in as 1 in API call.
+        opus.lastNormalizeurlRequestNo++;
+        let url = "/opus/__normalizeurl.json?" + hash + "&reqno=" + opus.lastNormalizeurlRequestNo;
+        $.getJSON(url, function(normalizeurlData) {
+
+            if (normalizeurlData["reqno"] < opus.lastNormalizeurlRequestNo) {
+                return;
+            }
+
+            $.each(normalizeurlData.new_slugs, function(idx, slug) {
```
rfrenchseti: This should be commented out until on the startobs branch.

opus/application/static_media/js/opus.js
```diff
@@ -545,6 +658,10 @@ $(document).ready(function() {
                     case "op-empty-cart":
                         o_cart.emptyCart();
                         break;
+                    case "op-update-url":
+                        // if user clicks "ok", we hide the link to url message on navabr
```
rfrenchseti: navbar is spelled wrong.

rfrenchseti (changes requested):

opus/application/apps/ui/templates/confirm_modal.html
rfrenchseti: Also, let's call the dialog something other than "Update URL". That's a verb phrase, and it sounds like it's asking the user to do something. How about "Message About Your URL".

opus/application/apps/ui/templates/base.html
dstopp: I suggest making this like we see banking notifications - and more generic; perhaps one day we might have more user messages to display.  Something like "User Notification Available".  Icon suggesions: https://fontawesome.com/icons/envelope-open-text?style=solid, https://fontawesome.com/icons/envelope-open?style=regular, https://fontawesome.com/icons/paper-plane?style=regular

opus/application/static_media/css/opus.css
```diff
@@ -38,6 +38,15 @@ body {
     }
 }
 
+.op-update-url-msg {
```
dstopp: I think we want to make this entire message thing a little more generic - perhaps op-user-message or something along that line, so we can reuse if we choose to

opus/application/static_media/js/opus.js
dstopp: there is a cart_startobs and startobs in opus.prefs.  It is already implemented in the other branch.

opus/application/static_media/js/opus.js
```diff
@@ -100,6 +102,8 @@ var opus = {
     helpPanelOpen: false,
 
     minimumPSLength: 30,
+
+    currentObs: 1,
```
dstopp: this is browse-specific and doesn't belong in opus.js.  I am working to get away from keeping these globals; please see the startobs branch.

opus/application/static_media/js/opus.js
dstopp: I am unclear; are you testing to see if two objects are identical?  we already have a function for that.  I agree that the name could use some help; suggestion might be "selectionObjectsEqual()  ... assuming there is more going on in there that just checking for object === not sufficient.

rfrenchseti (changes requested):
I'm seeing a fundamental problem with your normalizeurl changes. When the user pastes a URL, it is first parsed by hash, then unparsed by hash back into a URL that is sent to normalizeurl. This process modifies the URL in all sorts of ways. For example, currently "page" is not implemented in hash.js since we don't want it to be there in the future. But that means the URL that is sent to normalizeurl doesn't have a page= slug even if it was present in the original URL. If a field like cols= is missing, OPUS is adding it before normalizeurl can see the original so the user doesn't get a notification about it. normalizeurl is designed to be sent the raw URL as originally typed/bookmarked by the user without OPUS making any modifications to it. I think normalizeurl should be called as the VERY first thing in document.ready, and should use the getHash() function to get the raw string.

opus/application/static_media/js/search.js
```diff
@@ -134,9 +134,17 @@ var o_search = {
             let slug = $(this).attr("name");
             let currentValue = $(this).val().trim();
             let values = [];
-            values.push(currentValue);
-            opus.selections[slug] = values;
+            // values.push(currentValue);
+            // opus.selections[slug] = values;
 
+            if (currentValue) {
```
dstopp, line 137: I'm not sure I understand why you need to create an array w/a single var to set opus.selections[slug] - wouldn't it be simpler to do something like:
if (currentValue) {  // currentValue.length > 0 ..?
   opus.selections[slug] = [currentValue];
} else { ...

dstopp (changes requested):

juzen2003:
# Update:
* Modify to have "User Notification Available" with an envelope icon shows up in top navbar when there is a returned message.
    * Update the tooltip to "Click to read an important message".
    * Update the modal title to "Message About Your URL".
    * Modify confirm_modal.html to make it more generic.
* Put back "page" and comment out "startobs" based on feedback.
* Rename "op-update-url-msg" to "op-user-msg".
* Put normalized url api call as the 1st action to run when document ready is triggered, rest of steps will be run after it.
* Modify the update action of opus.selections in search.js.
    * If there is no selections for a slug, the slug will be removed from opus.selections. (previously it will still be in the opus.selections and store as either slug:[] for choice input or slug:[""] for string/range input)
    * Removed selectionsObjectsAreTheSame function.
        * we can now use o_utils.areObjectsEqual because we are comparing if two objects are exactly the same.
* Remove the reqno for normalized url api call, because it's been call during the document ready callback function, and during this time all variables are reloaded to default, so reqno number will always be 1 here when it's passed to url for api call.

opus/application/static_media/js/search.js
juzen2003: Thanks for the prompt feedback.
- After modifying the update action of opus.selections in search.js, opus.selections and modifiedSelections will be exactly the same, so we can use o_utils.areObjectsEqual. checkIfSelectionsObjectsAreTheSame function is removed.
- Regarding to the logic of updating opus.selections, it's been updated based on feedback.

opus/application/static_media/js/opus.js
rfrenchseti: Ok I will add it to the back end after this pull is finished. I have variety of other changes to make to normalize url as well.

opus/application/static_media/js/opus.js
rfrenchseti: === tests for the exact same object, not the same contents. It’s obvious from the code that the objects will never be the same object. But they could have the same contents. As discussed in slack the problem is that opus allows both “planet=“ and (nothing) for an empty search string. This is supposed to handle those cases, which are unique to opus search selections.  EDIT: Except I guess this is no longer needed at all now that this dichotomy has been fixed.

rfrenchseti:
Can you make the "envelope" symbol bright yellow, like we do with the "New" badge next to "Recent Announcements"?

rfrenchseti:
Also can we make the escape key get rid of the confirm modal? It should be equivalent to clicking the "No" button.

dstopp (approved):

rfrenchseti (approved):

PR #639, opened 2019-04-16
B051

juzen2003:
# Update:
* Display a modal when cart status api call has error returned. Call loadData to render the page again.
* Rename tableSorting to reRenderData


opus/application/apps/ui/templates/base.html
```diff
@@ -289,18 +289,31 @@ <h3 class='op-header-text'></h3>
...
             id_name="op-empty-cart"
             confirm_title="Confirm Remove All Observations From Cart"
             confirm_msg="Are you sure you want to remove all observations from your cart?"
             target="op-empty-cart"
+            yes_text="Yes"
+            no_text="No"
+        %}
+        {% include "confirm_modal.html" with
+            id_name="op-cart-status-error-msg"
+            confirm_title="Message About Cart Status"
```
rfrenchseti: It's only possible to have a failure return with a status code on add/addrange/addall, so let's make the title something more meaningful to the user: "Failed to Add Observations to Cart".

opus/application/apps/ui/templates/confirm_modal.html
```diff
@@ -12,8 +12,12 @@
         {{confirm_msg}}
       </div>
       <div class="modal-footer">
-        <button type="submit" class="btn btn-primary" data-dismiss="modal" data-target={{target}}>Yes</button>
-        <button type="cancel" class="btn btn-danger">No</button>
+        {% if yes_text != None %}
```
rfrenchseti: You can just say `{% if yes_text %}` and `{% if no_text %}`

opus/application/static_media/js/cart.js
```diff
@@ -388,6 +388,19 @@ var o_cart = {
 
         o_cart.lastCartRequestNo++;
         $.getJSON(url  + add_to_url + "&reqno=" + o_cart.lastCartRequestNo, function(statusData) {
+            if (statusData.reqno < o_cart.lastCartRequestNo) {
```
rfrenchseti: Ooooo...this is going to be a problem. While we only want to display the cart badge number for the final API call, any one of the intermediate calls might return an error that the user cares about. So you have to pop up the dialog for *all* returns, but only update the badge for the "last" one. Just move this if below the statusData if below and it should be fine.

rfrenchseti (changes requested):
Now that I see the return error messages, they are pretty ugly to the user. Once Dave finishes this change I'll update the error messages in the backend.

juzen2003:
# Update: 
* B051
  - Update modal title to "Failed to Add Observations to Cart".
  - Disable the ability to close modal when clicking outside of modal area
  - Update to have statusDataErrorCollector and display the whole series of error modals one after another. New modal will show up when previous one is close.

rfrenchseti (approved):

dstopp (approved):

opus/application/settings.py
```diff
@@ -352,7 +352,7 @@
 
 THUMBNAIL_NOT_FOUND = 'https://tools.pds-rings.seti.org/static_media/img/thumbnail_not_found.png'
 
-MAX_SELECTIONS_ALLOWED = 10000
+MAX_SELECTIONS_ALLOWED = 3
```
dstopp: oops probably you don't want to check this in :)

opus/application/static_media/css/opus.css
```diff
@@ -236,8 +236,9 @@ ul.searchMenu {
     opacity: 0.15;
 }
 
-#confirm-modal {
+#op-confirm-modal {
     border-radius: 6px;
+    border: solid 3px #D3D3D3;
```
dstopp: please use a bootstrap color and not a hard-coded color.

PR #640, opened 2019-04-18
New cart error messages

rfrenchseti:


opus/application/apps/cart/views.py
```diff
@@ -693,22 +693,23 @@ def _get_cart_count(session_id):
...
+                +'to cart')
 
     # Note that this doesn't handle the case where some or all of the opus_ids
     # are already in the cart. It's difficult to handle this well, would slow
     # down the API, and should never happen when using the UI anyway.
     num_selections = _get_cart_count(session_id)
-    if num_selections+len(opus_id_list) > settings.MAX_SELECTIONS_ALLOWED:
-        return 'Too many observations in cart'
+    if num_selections >= settings.MAX_SELECTIONS_ALLOWED:
+        return (f'Your request to add OPUS ID {opus_id} to the cart failed '
```
dstopp, line 710: is this going to post a message w/all the opus IDs that did not get added?  I"m not clear that OPUS_ID {opus_id} is needed/useful.

rfrenchseti: Add only allows one opus id so this message just displays one opus id. Since you can have multiple add requests outstanding due to network delay, I thought it would be useful to tell people which one failed. Otherwise you could get a series of modals that all say exactly the same thing, which would probably be annoying.

dstopp:
ok, looks good.

PR #642, opened 2019-04-18
Tooltip changes requested by Mitch, plus fix Select Metadata

rfrenchseti:


opus/import/table_schemas/obs_pds.json
```diff
@@ -92,7 +92,7 @@
         "pi_category_name": "obs_pds",
         "pi_dict_context": "OPUS_PDS",
         "pi_dict_name": "DATA_SET_ID",
-        "definition": "A string assigned by the PDS describing a collection of observations. A Data Set ID usually encodes, at a minimum, the mission, instrument, and version number of the data as provided by the instrument team.",
+        "definition": "A string assigned by the PDS describing a PDS3 dataset. Datasets are often aggregations of multiple volumes. A Data Set ID usually encodes, at a minimum, the mission, instrument, and version number of the data as provided by the instrument team.",
```
dstopp: we are referring to Dataset and Data Set - i think we should have a consistent way of referring to that entity.

dstopp (changes requested):

dstopp (approved):

opus/import/table_schemas/obs_pds.json
rfrenchseti: Replaced with "data set" everywhere except in slug names.

PR #643, opened 2019-04-18
Startobs

dstopp:
Initial stab at replacing page w/startobs.


rfrenchseti:
- When you drag the slider to a number and then release the mouse, the slider sometimes jumps a few "lines". This seems to happen when you slide somewhere, then use the scrollbar to move a fraction of a thumbnail, then slide again. It seems likely that the math used to go to a particular obs in the top left and the math used to figure out where the slider should be based on the top left don't quite agree. [FIXED]
- Similarly, if you do a fractional thumbnail scroll, and then switch to the table view, the slider might move one line. [FIXED]
- When you use the slider to go to the last observation line, that line is sometimes mostly off the top of the browser and you can only see the bottom tool row. [FIXED]
- When you use the slider to go to the last observation, then use the slider to go back earlier somewhere, you only get a partial load of the data, even though there is more available to fill out the browser. This only seems to happen when you go to the last obs first. [FIXED]
- As you scroll the table with the mouse wheel or up/down arrows, the slider doesn't move until a new page is prefetched. (It does move if you drag the scrollbar)
- If you scroll in the table view and then return to the gallery, the gallery hasn't scrolled to the same location so now the slider and the top left thumbnail differ hugely.
- Scrolling up in the gallery prefetches the previous page but has a huge jump to the beginning of that page.
- Using the right or left arrows in the browse dialog does very weird things when hitting the end of the prefetched data.
- Unrelated to these changes, but as you keep infinite scrolling, the browser memory footprint can get REALLY big. We need to delete obs that are far from where we're currently looking.
- If you display the browse dialog and then use the slider, the browse dialog gets very confused because the data is no longer there.
- When you click on the add/trashcan icon for a thumbnail in the cart, the screen can scroll slightly and place part of the bottom icons row off the bottom of the browser. You can't scroll down any further so it's now impossible to see the toolbar on those icons. (This particularly happens when you click on a thumbnail at the bottom of the screen)
- I think we need the slider on the cart page.
- Clicking a sort pill on the cart page doesn't do anything at all.

rfrenchseti:
In general the problem with the slider jumping seems to be related to whether the main gallery scrollbar is at the top of its travel or somewhere further down. As long as it's at the top, everything is fine. The further it moves down, the bigger the problem with the slider. I assume this means that the position of the scrollbar is being ignored in one of the computation directions to figure out the top left thumbnail.

rfrenchseti:
Problem:

I don't understand this line:
```
        if (lastObs === undefined || firstObs === undefined || $(elem).length === 0 ||
            (startObs > lastObs + 1) || (startObs < firstObs - 1)) {
```
Why are you comparing to lastObs+1 and firstObs-1 using > and <? As written, if startObs == lastObs+1 (which I assume is past the end of the current data), then it won't trigger the reload. However, you luck out in this case because $(elem).length === 0 since the obs isn't in the DOM. After fixing the comparison, given that you're checking whether or not the element is in the DOM, what's the point of checking firstObs and lastObs anyway? That's just redundant. I'm not sure which one is more efficient, but it seems like you should check either $(elem).length or firstObs and lastObs, but not both. firstObs and lastObs are used later in this function, so you have to compute them anyway, so elem seems like the thing to get rid of to avoid an additional DOM lookup.

rfrenchseti:
Problem:

There is one big bug here, though. We pass the reqno as:
        `let reqno = (opus.prefs.view === "browse" ? ++o_browse.lastLoadDataRequestNo : ++o_cart.lastLoadDataRequestNo);`
but then we only compare:
            `if (data.reqno < o_browse.lastLoadDataRequestNo) {`
so whenever you're looking at the cart, the reqno is going to be wrong, and depending on how much time you spent on the browse page vs. the cart page it will either work or not.

rfrenchseti:
So here's the first problem with the jumping slider. When you slide to an obs that isn't cached, it clears the cache and starts completely over. But it leaves the vertical scrollbar where it was before. So first it loads the new data. Then it asks what is the obs at the top left corner? But this isn't the beginning of the data that was just loaded, because it is scrolled off the top of the screen. So it jumps the slider to whatever that obs happens to be, which is the obs we want plus the amount of vertical scroll. I think the easiest solution is to reset the vertical scrollbar to the top when you start over with a cleared cache. There's no reason to leave the scrollbar partially down the screen anyway. It's just confusing, and there actually isn't any data before what's displayed anyway. So here's the fix, in browse.js at line 1137:
```
         if (!o_browse.galleryBegun) {
                o_browse.initTable(data.columns);
                $(`${selector}`).scrollTop(0); // THIS
                $(`${selector} .dataTable`).scrollTop(0); // AND THIS

                if (!$(selector).data("infiniteScroll")) {
```
With this change the slider no longer jumps for me.

rfrenchseti:
Aren't there two .gallery-contents DOM elements? This code (in setScrollbarOnSlide) doesn't distinguish them:
```
let galleryContainerTopPosition = $(".gallery-contents").offset().top;
let galleryScrollbarPosition = $(".gallery-contents").scrollTop();
```

rfrenchseti:
OK there's still a second problem with the jumping slider. This one appears to be related to `setScrollbarOnSlide`. If you only move the slider a small amount at a time (like one or two rows), it works great. If you move the slider a lot, such that it runs out of cached data, you hit the code above, which now works great. But something happens in between - move the slider so it's near the end of the cached data (like if you have 72 obs prefetched in a 12x6 grid, slide to 72+1-12=61). Now it jumps again. 
I feel like this second slider jump problem is related to infinite scroll. As you move the slider by more and more rows from firstObs, the vertical scrollbar moves further down the screen. But at some point, I think infinite scroll says "that's enough" and changes the size or position or something of the scrollbar so it's back higher on the screen. In other words infinite scroll doesn't let the scrollbar get too close to the bottom. I think what's probably happening is that `setScrollbarOnSlide` is moving the scrollbar far enough down that infinite scroll is deciding it doesn't like the new position and moves it again, changing the actual scroll position on the screen.
I'm really not sure how to fix this one.

rfrenchseti:
When there are no items in the cart, the cart page displays two "Your Cart is Empty" messages.

rfrenchseti:
Fixed: Checkbox in cart not working when hamburger add/remove is used.
Fixed: Shift+click on table row.
Fixed: Metadata Selector not saving changes.
Fixed: Metadata Selector causing empty gallery message to go away.

Question: Why do we want to set the startobs to 1 when the user changes the columns? Can't we keep showing them the part of the gallery/table they were already looking at?

Bug: Right now it appears that when you go "up" it doesn't move smoothly, but instead moves you to the top of the new page, so there's a huge page-sized jump in what data you're looking at. The same thing happens even if you move the binoculars.

opus/application/static_media/js/browse.js
```diff
@@ -51,18 +51,18 @@ var o_browse = {
...
-                    }
-                } else {
-                    if ($(`${namespace} .gallery-contents`).scrollTop() === 0) {
-                        $(`${namespace} .gallery-contents`).infiniteScroll("loadNextPage");
-                    }
+            let view = o_browse.getViewInfo();
+            let tab = view.namespace;
+            if (opus.prefs[`${view.prefix}startobs`] > 0) {
+                let prev = $(`${view.namespace} [data-obs]`).first().data("obs") - o_browse.getLimit();
+                if ($(`${tab} .gallery-contents`).scrollTop() || $(`${tab} .dataTable`).scrollTop() === 0) {
```
rfrenchseti: The use of || here is confusing me. As written, this says "If gallery-contents scrollTop is NOT zero, OR dataTable scrollTop IS zero". Don't you want to say if either is zero?

rfrenchseti (changes requested):

opus/application/static_media/js/browse.js
```diff
@@ -957,19 +958,18 @@ var o_browse = {
...
-            if (namespace == "#browse") {   // not yet supported for cart
-                $(".dataTable tbody").prepend(tableHtml);
-            }
-        } else {
-            $(".gallery", namespace).append(galleryHtml);
-            if (namespace == "#browse") {   // not yet supported for cart
-                $(".dataTable tbody").append(tableHtml);
+            // wondering if there should be more logic here to determine if the new block of observations
+            // is contiguous w/the existing block of observations, not just before/after...
+            if ($(`${namespace} .thumbnail-container`).first().data("obs") > data.start_obs) {
```
rfrenchseti, line 1069: If the gallery is empty, isn't there no first() element? In which case I assume first().data("obs") returns "undefined". That makes the if confusing. Perhaps you should check if there's a first() element at all, and only then do the comparison?

rfrenchseti (changes requested):

PR #645, opened 2019-04-19
New ui dave b005

juzen2003:
# Update:
* Sorting pill:
    * Sorting pill arrows will toggle right away when it's clicked.
    * Sorting pill will be removed right away when "x" is clicked.
* Table header:
    * Arrows will toggle right away when it's clicked.

### DIFF
--- a/opus/application/apps/results/views.py
+++ b/opus/application/apps/results/views.py
@@ -92,17 +92,18 @@ def api_get_data_and_images(request):
             },
             ...
          ],
-         'page_no':         page_no,   OR   'start_obs': start_obs,
-         'limit':           limit,
-         'order':           comma-separate list of slugs,
-         'order_list':      [entry, entry...]
-                            entry is {'slug': slug_name,
-                                      'label': pretty_name,
-                                      'descending': True/False,
-                                      'removeable': True/False},
-         'count':           len(page),
-         'columns':         columns (corresponds to <col1> etc. in 'metadata'),
-         'reqno':           reqno
+         'page_no':             page_no,   OR   'start_obs': start_obs,
+         'limit':               limit,
+         'order':               comma-separate list of slugs,
+         'order_list':          [entry, entry...]
+                                entry is {'slug': slug_name,
+                                        'label': pretty_name,
+                                        'descending': True/False,
+                                        'removeable': True/False},
+         'count':                       len(page),
+         'columns':             columns with units(corresponds to <col1> etc. in 'metadata'),
+         'columns_no_units':    columns without units,
+         'reqno':               reqno
         }
     """
     api_code = enter_api_call('api_get_data_and_images', request)
@@ -169,7 +170,8 @@ def api_get_data_and_images(request):
     cols = request.GET.get('cols', settings.DEFAULT_COLUMNS)
 
     labels = labels_for_slugs(cols_to_slug_list(cols))
-    if labels is None:
+    labels_no_units = labels_for_slugs(cols_to_slug_list(cols), units=False)
+    if labels is None or labels_no_units is None:
         ret = Http404(settings.HTTP404_UNKNOWN_SLUG)
         exit_api_call(api_code, ret)
         raise ret
@@ -202,13 +204,14 @@ def api_get_data_and_images(request):
         exit_api_call(api_code, ret)
         raise ret
 
-    data = {'page':         new_page,
-            'limit':        limit,
-            'count':        len(image_list),
-            'order':        order,
-            'order_list':   order_list,
-            'columns':      labels,
-            'reqno':        reqno
+    data = {'page':             new_page,
+            'limit':            limit,
+            'count':            len(image_list),
+            'order':            order,
+            'order_list':       order_list,
+            'columns':          labels,
+            'columns_no_units': labels_no_units,
+            'reqno':            reqno
            }
 
     if page_no is not None:
--- a/opus/application/apps/results/views.py
+++ b/opus/application/apps/results/views.py
@@ -187,7 +187,7 @@ def api_get_data_and_images(request):
 
     order_list = []
     for idx, (slug, label) in enumerate(zip(order_slugs, order_labels)):
-        removeable = idx != len(order_slugs)-1;
+        removeable = not slug.endswith('opusid')
         desc = False
         if slug[0] == '-':
             slug = slug[1:]
--- a/opus/application/apps/results/views.py
+++ b/opus/application/apps/results/views.py
@@ -101,7 +101,8 @@ def api_get_data_and_images(request):
                                         'descending': True/False,
                                         'removeable': True/False},
          'count':                       len(page),
-         'columns':             columns with units(corresponds to <col1> etc. in 'metadata'),
+         'columns':             columns with units
+                                (corresponds to <col1> etc. in 'metadata'),
          'columns_no_units':    columns without units,
          'reqno':               reqno
         }
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -148,7 +148,6 @@ var o_browse = {
             $(".op-page-loading-status > .loader").show();
             let slug = $(this).parent().attr("data-slug");
             let descending = $(this).parent().attr("data-descending");
-            // Sync up arrows in table header
             let headerOrderIndicator = $(`.dataTable th a[data-slug="${slug}"]`).find("span:last");
             let pillOrderIndicator = $(this);
             o_browse.reRenderData = true;
@@ -446,6 +445,8 @@ var o_browse = {
         });
     }, // end browse behaviors
 
+    // update order arrows right away when user clicks on sorting arrows in pill or table header
+    // sync up arrows in both sorting pill and table header
     updateOrderIndicator: function(headerOrderIndicator, pillOrderIndicator, isDescending) {
         let headerOrder = isDescending ? "sort-desc" : "sort-asc";
         let headerOrderArrow = isDescending ? o_browse.tableSortUpArrow : o_browse.tableSortDownArrow;
@@ -1136,6 +1137,7 @@ var o_browse = {
             let icon = ($.inArray(slug, order) >= 0 ? "-down" : ($.inArray("-"+slug, order) >= 0 ? "-up" : ""));
             let columnSorting = icon === "-down" ? "sort-asc" : icon === "-up" ? "sort-desc" : "none";
             let columnOrdering = `<a href='' data-slug='${slug}'><span>${header}</span><span data-sort='${columnSorting}' class='column_ordering fas fa-sort${icon}'></span></a>`;
+
             $(".dataTable thead tr").append(`<th id='${slug} 'scope='col' class='sticky-header'><div>${columnOrdering}</div></th>`);
         });
 
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -476,6 +476,31 @@ var o_browse = {
         });
     }, // end browse behaviors
 
+    updateOrderIndicator: function(headerOrderIndicator, pillOrderIndicator, descending) {
+        let headerOrder = descending ? "sort-desc" : "sort-asc";
+        let headerOrderArrow = descending ? o_browse.tableSortUpArrow : o_browse.tableSortDownArrow;
+        let isPillOrderDesc = descending ? "true" : "false";
+        let pillOrderTooltip = descending ? "Change to ascending sort" : "Change to descending sort";
+        let pillOrderArrow = descending ? o_browse.pillSortUpArrow : o_browse.pillSortDownArrow;
+
+        headerOrderIndicator.data("sort", `${headerOrder}`);
+        headerOrderIndicator.attr("class", `column_ordering ${headerOrderArrow}`);
+        pillOrderIndicator.parent().attr("data-descending", `${isPillOrderDesc}`);
+        pillOrderIndicator.attr("title", `${pillOrderTooltip}`);
+        pillOrderIndicator.find("i").attr("class", `${pillOrderArrow}`);
+        // headerOrderIndicator.data("sort", "sort-asc");
+        // headerOrderIndicator.attr("class", `column_ordering ${o_browse.tableSortDownArrow}`);
+        // pillOrderIndicator.parent().attr("data-descending", "false");
+        // pillOrderIndicator.attr("title", "Change to descending sort");
+        // pillOrderIndicator.find("i").attr("class", `${o_browse.pillSortDownArrow}`);
+        //
+        // headerOrderIndicator.data("sort", "sort-desc");
+        // headerOrderIndicator.attr("class", `column_ordering ${o_browse.tableSortUpArrow}`);
+        // pillOrderIndicator.parent().attr("data-descending", "true");
+        // pillOrderIndicator.attr("title", "Change to ascending sort");
+        // pillOrderIndicator.find("i").attr("class", `${o_browse.pillSortUpArrow}`);
+    },
+
     renderSortedDataFromBeginning: function() {
         opus.lastPageDrawn.browse = 0;
         opus.prefs.page = default_pages; // reset pages to 1 when col ordering changes
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -320,7 +320,8 @@ var o_browse = {
                 slug = "-"+slug;
             }
             let slugIndex = $.inArray(slug, opus.prefs.order);
-            // The clicked-on slug should always be in the order list; this is just a safety precaution
+            // The clicked-on slug should always be in the order list;
+            // The "if" is a safety precaution and the condition should always be true
             if (slugIndex >= 0) {
                 opus.prefs.order.splice(slugIndex, 1);
             }
@@ -353,7 +354,8 @@ var o_browse = {
                 isDescending = true;
             }
             let slugIndex = $.inArray(slug, opus.prefs.order);
-            // The clicked-on slug should always be in the order list; this is just a safety precaution
+            // The clicked-on slug should always be in the order list;
+            // The "if" is a safety precaution and the condition should always be true
             if (slugIndex >= 0) {
                 opus.prefs.order[slugIndex] = new_slug;
             }
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -485,35 +485,27 @@ var o_browse = {
             headers.attr("class", defaultTableSortArrow);
         }
 
-        // If pill already exists, we update the pill, else we re-render the pill
-        // Note: we have to re-render the pill if user clicks on opus id table header, because in this case, we only want one pill (opus id) displayed.
-        if (pillOrderIndicator.length !== 0 && slug !== "opusid") {
-            pillOrderIndicator.parent().attr("data-descending", `${isPillOrderDesc}`);
-            pillOrderIndicator.attr("title", `${pillOrderTooltip}`);
-            pillOrderIndicator.find("i").attr("class", `${pillOrderArrow}`);
-        } else {
-            // re-render each pill
-            let listHtml = "";
-            $.each(opus.prefs.order, function(index, orderEntry) {
-                isPillOrderDesc = orderEntry[0] === "-" ? "true" : "false";
-                pillOrderArrow = orderEntry[0] === "-" ? pillSortUpArrow : pillSortDownArrow;
-                orderEntry = orderEntry[0] === "-" ? orderEntry.slice(1) : orderEntry;
+        // Re-render each pill
+        let listHtml = "";
+        $.each(opus.prefs.order, function(index, orderEntry) {
+            isPillOrderDesc = orderEntry[0] === "-" ? "true" : "false";
+            pillOrderArrow = orderEntry[0] === "-" ? pillSortUpArrow : pillSortDownArrow;
+            orderEntry = orderEntry[0] === "-" ? orderEntry.slice(1) : orderEntry;
 
-                // retrieve label from either displayed header's data-label attribute or displayed pill's text
-                let label = $(`#browse .dataTable th a[data-slug="${orderEntry}"]`).data("label") || $(`#browse .sort-contents span[data-slug="${orderEntry}"] .flip-sort`).text();
+            // retrieve label from either displayed header's data-label attribute or displayed pill's text
+            let label = $(`#browse .dataTable th a[data-slug="${orderEntry}"]`).data("label") || $(`#browse .sort-contents span[data-slug="${orderEntry}"] .flip-sort`).text();
 
-                listHtml += "<li class='list-inline-item'>";
-                listHtml += `<span class='badge badge-pill badge-light' data-slug="${orderEntry}" data-descending="${isPillOrderDesc}">`;
-                if (orderEntry !== "opusid") {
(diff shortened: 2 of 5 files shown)


opus/application/static_media/js/browse.js
```diff
@@ -45,6 +45,12 @@ var o_browse = {
     tempHash: "",
     dataNotAvailable: false,
     onRenderData: false,
+
+    // font awesome icon class
+    pillSortUpArrow: "fas fa-arrow-circle-up",
```
rfrenchseti: Since these will never change would it be better to somehow define them as "const"?

opus/application/static_media/js/browse.js
```diff
@@ -424,6 +445,22 @@ var o_browse = {
         });
     }, // end browse behaviors
 
+    // update order arrows right away when user clicks on sorting arrows in pill or table header
+    // sync up arrows in both sorting pill and table header
+    updateOrderIndicator: function(headerOrderIndicator, pillOrderIndicator, isDescending) {
```
rfrenchseti: I don't really understand how this works currently. It assumes that there is always the same column in the sort order pills and the table header, so all the input parameters are valid and you can swap both. But it's entirely possible to be sorting on some column that is not currently in the table (e.g. sort on a column, then select metadata and delete that column). In this case I assume some of these variables are invalid so I don't know what it does then.

rfrenchseti (changes requested):
1) If you watch the API calls, the first time you click on a table header you get something like "order=target" for the first call to dataimages.json. But the second call (prefetching the second page) says "order=target,opusid". This is because after the first one returns, it told us what sort order it was actually using, and it always makes sure that opusid is in the sort order somewhere so that the order is guaranteed to be unique. We should simulate what the backend is going to do by making sure that opusid is always in the sort order. Thus when you click on a table header, and order=target, look through the string and see if "opusid" occurs anywhere (it could be opusid or -opusid, so just look for opusid). If it does, you're good (right now this would only happen if the user actually clicked on the opusid column, but in the future we'll allow sorting on multiple columns so it could be anywhere). If it doesn't, then append ",opusid" to the sort order. That way it will agree with what the backend returns to us and things will be cleaner in the future.

2) If you click on a table header, the arrow switches instantly, but the pills don't change until dataimages.json returns. We should have both places react instantly. Luckily you have the "full pretty name" of the column because it's in the table header, so you know what to put in the pill. As above, it should always have opusid somewhere, so if you click on target, there should be two pills: target and opusid. If you do #1 first, then the sort order string will be accurate for the creation of the pills. Obviously make sure the arrows are going the right way in the new pills.

3) Click on a table header, like target. The pills say target(down) opusid(down) and the table header says target(down). Now click on the table header for target. The pills will say target(up) opusid(down) as expected but the table header will say target(up) AND opusid(up). We only have the ability to sort by a single column using the table view at this point, so it really only makes sense to have the first sort order slug set the table header. We will need to figure out how to do this better in the future.

rfrenchseti:
Here are some good test cases to try:

1) Using the table, sort on a column other than opusid not currently being sorted on (see table header and pills updated immediately - opusid added to end)
2) Using the table, sort on opusid (see the table header and pill updated immediately - only one pill)
3) Using the table, sort on a column, then select metadata and remove that column, then toggle the pill

Watch the API calls to make sure the order slug is the same for the original fetch and the prefetch.

Slow down the API and do lots of clicks on the same column and on multiple columns and switching between table header and pills before API returns.

rfrenchseti (changes requested):
Another issue:
- Slow down the API return. Click on a table header. Click on another table header. Click on a third table header. All three headers will have up or down arrows, even though only one of them is actually being used for sorting. Once the API returns, they all return to up+down except for the one being sorted on.

juzen2003:
# Update Summary:
* Order slug values in url will get updated right away when user click table header/sorting pill.
    * The one in prefetch will be the same as the one in final fetch.
* Sorting pill arrow/label will match table header arrow/title immediately.
* Click on sorting pill:
    * If slug only exists in sorting pill but not in table header, arrows in table header will all be up + down.
        * Sorting pill arrow and order slug in url will get updated
    * If there is another sorting pill other than opus id, clicking on opus id pill will not change the arrow in table header (will update the order slug in url)
* Click on table header:
    * Click on a table header that is currently not in the order slug, both arrows and pill label switch instantly.
    * Only one arrow will get updated to either up or down at a time, rest of arrows will be up + down.
        * There will be no more cases like 3 headers with down arrows while waiting for API return
    * Click on opus id, only opus id pill will be displayed.

opus/application/static_media/js/browse.js
```diff
@@ -8,6 +8,13 @@
 /* globals default_pages, default_columns */
 
 /* jshint varstmt: false */
+// font awesome icon class
```
rfrenchseti, line 10: The /* jshint */ line needs to go after your const lines and right before the "var".

opus/application/static_media/js/browse.js
```diff
@@ -312,29 +277,102 @@ var o_browse = {
             // we will hide page status loader from infiniteScroll if op-page-loading-status loader is spinning
             $(".op-page-loading-status > .loader").show();
             let orderBy =  $(this).data("slug");
+            let targetSlug = orderBy;
 
+            // get order of opusid when table header is clicked
+            let hash = o_hash.getHashArray();
+            let opusidOrder = (hash.order || hash.order.match(/(-?opusid)/)) ? hash.order.match(/(-?opusid)/)[0] : "opusid";
```
rfrenchseti: I don't think you want to use || here. What this is currently saying is "If hash.order is not empty...retrieve hash.order.match...[0]". I think what you're trying to say is "If hash.order is not empty AND hash.order.match... is true THEN return hash.order.match...[0]". && would work better.

opus/application/static_media/js/browse.js
```diff
@@ -424,6 +462,59 @@ var o_browse = {
...
+
+            // Reset arrows on rest of table headers
+            // let headers = $(`.dataTable th a:not([data-slug="opusid"], [data-slug=${slug}])`).find("span:last");
+            let headers = $(`.dataTable th a:not([data-slug=${slug}])`).find("span:last");
+            headers.data("sort", "none");
+            headers.attr("class", defaultTableSortArrow);
+        }
+
+        // If pill already exists, we update the pill, else we re-render the pill
+        if (pillOrderIndicator.length !== 0 && slug !== "opusid") {
```
rfrenchseti: Why are you treating opusid specially here? Why don't we just flip the arrow on the opusid pill like we do any other?

rfrenchseti: This causes a really weird bug as written. Add Right Ascension (Min) and Right Ascension (Max) as columns. Go to the table and sort on Right Ascension (Min). Then go to select metadata and remove Right Ascension (Min). Now click on the opus id pill to swap the sort order. Due to this line, it will need to rerender all the pills. But below it gets the name of the pill from the table header - but the column isn't in the table anymore! It will thus do something weird until the data returns, at which point we once again redraw all the pills due to updateSortOrder() and that will get the names right.

rfrenchseti (changes requested):

opus/application/static_media/js/browse.js
```diff
@@ -312,29 +277,102 @@ var o_browse = {
...
+            let slug = $(this).parent().attr("data-slug");
+            let descending = $(this).parent().attr("data-descending");
+            o_browse.reRenderData = true;
+
+            if (descending == "true") {
+                slug = "-"+slug;
+            }
+            let slugIndex = $.inArray(slug, opus.prefs.order);
+            if (slugIndex >= 0) {
+                // The clicked-on slug should always be in the order list; this is just a safety precaution
```
dstopp: please move the comment above the if, since that's what it's referring to, and clean up the text so it's clear that the if should always be true

opus/application/static_media/js/browse.js
```diff
@@ -312,29 +277,102 @@ var o_browse = {
...
+            if (descending == "true") {
+                slug = "-"+slug; // Old descending, new ascending
+                isDescending = false;
+            } else {
+                new_slug = "-"+slug; // Old ascending, new descending
+                isDescending = true;
+            }
+            let slugIndex = $.inArray(slug, opus.prefs.order);
+            if (slugIndex >= 0) {
+                // The clicked-on slug should always be in the order list; this is just a safety precaution
```
dstopp: same comment for this line; please move the comment above the if, since that's what it's referring to, and clean up the text so it's clear that the if should always be true

opus/application/static_media/js/browse.js
```diff
@@ -424,6 +462,59 @@ var o_browse = {
...
+            pillOrderIndicator.find("i").attr("class", `${pillOrderArrow}`);
+        } else {
+            // re-render each pill
+            let listHtml = "";
+            $.each(opus.prefs.order, function(index, orderEntry) {
+                isPillOrderDesc = orderEntry[0] === "-" ? "true" : "false";
+                pillOrderArrow = orderEntry[0] === "-" ? pillSortUpArrow : pillSortDownArrow;
+                orderEntry = orderEntry[0] === "-" ? orderEntry.slice(1) : orderEntry;
+
+                let label = $(`#browse .dataTable th a[data-slug="${orderEntry}"]`).find("span:first").text() || $(`#browse .sort-contents span[data-slug="${orderEntry}"] .flip-sort`).text();
```
dstopp: I find this access method easier to read:
$(`#browse .dataTable th a[data-slug="${orderEntry}"]`).first("span").text().

opus/application/static_media/js/browse.js
```diff
@@ -424,6 +462,59 @@ var o_browse = {
...
+                let indexOfUnit = label.indexOf("(")
```
dstopp: It seems like you are going to a big effort to get a piece of information that you could just put in the data- bit of the element in the way you want to see it... example:
$(`#browse .dataTable th a[data-slug="${orderEntry}"]`).data("label").  Just a thought, because then you won't have to mess around removing units and all that biz here, it's already there and on the element easy to access.

dstopp (changes requested):

opus/application/static_media/js/browse.js
juzen2003: The reason of treating opusid specially is for the case when we click "OPUS ID" table header and there are two sorting pills displayed. At this time, we only want to display "OPUS ID" pill, so we have to re-render the pill again (to remove the extra one pill). 

For the bug mentioned above, it was caused by method I used to get the label. It was creating the label by chopping all the string after "(" to remove the unit, but "(Max)" and "(Min)" also got chopped out. 

I follow Debby's feedback to store the proper label in data-label for each table header, and modify the logic of getting labels. It will now only remove "(" with units and keep "(Max)" and "(Min)" in label. (line 1194-1198 in browse.js)

opus/application/static_media/js/browse.js
```diff
@@ -320,8 +320,8 @@ var o_browse = {
                 slug = "-"+slug;
             }
             let slugIndex = $.inArray(slug, opus.prefs.order);
+            // The clicked-on slug should always be in the order list; this is just a safety precaution
```
rfrenchseti: Sorry, perhaps we weren't entirely clear. The position of the comment needs to move like you did, but the word "this" is confusing. (Yes I know I wrote the comment originally) Let's say "the if is a safety precaution and the condition should always be true".

opus/application/static_media/js/browse.js
```diff
@@ -353,8 +353,8 @@ var o_browse = {
                 isDescending = true;
             }
             let slugIndex = $.inArray(slug, opus.prefs.order);
+            // The clicked-on slug should always be in the order list; this is just a safety precaution
```
rfrenchseti: Same for this comment.

opus/application/static_media/js/browse.js
rfrenchseti: As we discussed in person, the easier way to handle this is just to re-render all the pills each time. That will make the code a lot less complex and also get rid of the dependence on opusid as a special case. Let's go ahead and do that now so the code is cleaner.

opus/application/static_media/js/browse.js
```diff
@@ -1189,10 +1188,17 @@ var o_browse = {
         $(".dataTable thead tr").append("<th scope='col' class='sticky-header'></th>");
         $.each(columns, function(index, header) {
             let slug = slugs[index];
+
```
rfrenchseti, line 1185: I understand why you're doing this, but this really can't be the right way to handle this. It's too inflexible for future cases that might have things other than (Min) or (Max) that can be confused with units. Here's the right thing to do, even though it's a bit more complicated:

- Update results/views.py api_get_data_and_images().
  - In addition to the existing "columns" return field, there should be one called "columns_no_units".
  - Update the comment in the function header to indicate that columns returns units, and the new one doesn't.
  - After the line that says `labels = labels_for_slugs(cols_to_slug_list(cols))` add a new line that defines labels_no_units and passes the argument units=False into the labels_for_slugs function. See example code nearby.
  - Check both of them for None and throw an error (see next line after above).
  - In the definition of "data" below add the "columns_no_units" field in the obvious way.
- When you call initTable in loadData, pass in both columns and columns_no_units
- Change initTable to take two arguments.
- Store the second one in opus.col_labels_no_units
- Use the no-units version to initialize the data label
- Kill off all this stuff about (Min) and (Max)
- You'll have to change the metadata selector behavior to pass in the second argument as well

rfrenchseti (changes requested):

opus/application/static_media/js/browse.js
juzen2003: Comments are updated in the latest commit.

opus/application/static_media/js/browse.js
juzen2003: - The logic of getting labels (without units) is updated in the latest commit.
- The logic of updateOrderIndicator is updated, now we will re-render the pill every time.

opus/application/apps/results/views.py
```diff
@@ -92,17 +92,18 @@ def api_get_data_and_images(request):
...
+         'page_no':             page_no,   OR   'start_obs': start_obs,
+         'limit':               limit,
+         'order':               comma-separate list of slugs,
+         'order_list':          [entry, entry...]
+                                entry is {'slug': slug_name,
+                                        'label': pretty_name,
+                                        'descending': True/False,
+                                        'removeable': True/False},
+         'count':                       len(page),
+         'columns':             columns with units(corresponds to <col1> etc. in 'metadata'),
```
rfrenchseti: Our Python coding standard is that no line should be longer than 80 characters (to conform with the PEP8 standard). Please put the (corresponds...) stuff on the next line (properly aligned with the "columns with units" above).

opus/application/static_media/js/browse.js
```diff
@@ -312,29 +277,104 @@ var o_browse = {
...
             $(".op-page-loading-status > .loader").show();
             let orderBy =  $(this).data("slug");
+            let targetSlug = orderBy;
 
+            // get order of opusid when table header is clicked
+            let hash = o_hash.getHashArray();
+            let opusidOrder = (hash.order && hash.order.match(/(-?opusid)/)) ? hash.order.match(/(-?opusid)/)[0] : "opusid";
+            let isDescending = true;
             let orderIndicator = $(this).find("span:last");
+            let pillOrderIndicator = $(`.sort-contents span[data-slug="${orderBy}"] .flip-sort`)
```
rfrenchseti: Please always run your new JS code through jshint.com. It caught a missing semicolon at the end of this line.

opus/application/static_media/js/browse.js
```diff
@@ -424,6 +464,50 @@ var o_browse = {
...
+        let listHtml = "";
+        $.each(opus.prefs.order, function(index, orderEntry) {
+            isPillOrderDesc = orderEntry[0] === "-" ? "true" : "false";
+            pillOrderArrow = orderEntry[0] === "-" ? pillSortUpArrow : pillSortDownArrow;
+            orderEntry = orderEntry[0] === "-" ? orderEntry.slice(1) : orderEntry;
+
+            // retrieve label from either displayed header's data-label attribute or displayed pill's text
+            let label = $(`#browse .dataTable th a[data-slug="${orderEntry}"]`).data("label") || $(`#browse .sort-contents span[data-slug="${orderEntry}"] .flip-sort`).text();
+
+            listHtml += "<li class='list-inline-item'>";
```
rfrenchseti, line 497: It bugs me that we now have two places in the code that create the pill list in almost identical ways. However, it's true that different information is available in each place. I think we have to leave this as-is right now and rewrite all of this once we start allowing the user to select multiple sort fields, because then we have all sorts of new problems.

opus/application/static_media/js/cart.js
```diff
@@ -190,8 +190,8 @@ var o_cart = {
             if (data.reqno < o_cart.lastRequestNo) {
                 return;
             }
-            if (opus.col_labels.length === 0) {
-                opus.col_labels = data.columns;
+            if (opus.colLabels.length === 0) {
```
rfrenchseti, line 193: I'm confused by this. Why are we updating colLabels here without updating colLabelsNoUnits?

opus/application/static_media/js/browse.js
```diff
@@ -424,6 +464,50 @@ var o_browse = {
...
+            // let headers = $(`.dataTable th a:not([data-slug="opusid"], [data-slug=${slug}])`).find("span:last");
+            let headers = $(`.dataTable th a:not([data-slug=${slug}])`).find("span:last");
+            headers.data("sort", "none");
+            headers.attr("class", defaultTableSortArrow);
+        }
+
+        // Re-render each pill
+        let listHtml = "";
+        $.each(opus.prefs.order, function(index, orderEntry) {
+            isPillOrderDesc = orderEntry[0] === "-" ? "true" : "false";
```
rfrenchseti: Although it works in this case, I think it's generally a bad idea to overwrite function parameters, because then if someone tries to use them later in the function they will be surprised they aren't the values what were passed in. It would be better to use new variables with local "let" scope inside the $.each.

opus/application/static_media/js/browse.js
```diff
@@ -424,6 +464,50 @@ var o_browse = {
...
+            pillOrderArrow = orderEntry[0] === "-" ? pillSortUpArrow : pillSortDownArrow;
+            orderEntry = orderEntry[0] === "-" ? orderEntry.slice(1) : orderEntry;
+
+            // retrieve label from either displayed header's data-label attribute or displayed pill's text
+            let label = $(`#browse .dataTable th a[data-slug="${orderEntry}"]`).data("label") || $(`#browse .sort-contents span[data-slug="${orderEntry}"] .flip-sort`).text();
+
+            listHtml += "<li class='list-inline-item'>";
+            listHtml += `<span class='badge badge-pill badge-light' data-slug="${orderEntry}" data-descending="${isPillOrderDesc}">`;
+            if (orderEntry !== "opusid") {
```
rfrenchseti: So this is really interesting. We are using different logic here and in updateSortOrder() to decide which pills should have the (x). In this case, you're saying any "opusid" needs to stay. In updateSortOrder(), it is using the "removeable" field from the returned data from dataimages.json. The logic for that field is in results/views.py line 189, where it says that the *last* field shouldn't be removeable, regardless of what it is. If the opusid isn't the last field, these two methods will cause different behaviors, resulting in the weird case of changing the sort order rendering the pill with opusid without an (x), and then when dataimages.json returns re-rendering them with the last field without an (x).

Note: This is a classic example of why it's bad to have mostly duplicated code that does things in different ways!

In this particular case, I think you have exposed a bug in the API. The way we're doing things right now, we must always have opusid *somewhere* in the order list, and thus we should make that one not removeable. Please change results/views.py line 189 from:
`removeable = idx != len(order_slugs)-1;` (I have no idea why this line ends with a ; - that's not a Python thing)
to:
`removeable = not slug.endswith('opusid')`
Don't forget to re-run the tests since you've touched the Django.

rfrenchseti (changes requested):

opus/application/static_media/js/browse.js
rfrenchseti: I think you missed a couple of cases: You reuse isPillOrderDesc and pillOrderArrow, overwriting the parameters that were passed in. I think these need to be new variables too. Also I'm not sure why you renamed this variable to "order". I would think of "order" as being the entire order slug, not just this one entry. Perhaps call it "orderEntrySlug", since that's what it really is?

rfrenchseti (changes requested):

juzen2003:
For updateOrderIndicator function, isPillOrderDesc and pillOrderArrow are defined at line 473 and 474 previously, and are used to update or re-render the pill originally. Since now we will re-render the pill every time, we can just declare those two variables inside the loop of iterating through order slugs.

rfrenchseti (approved):

opus/application/static_media/js/browse.js
```diff
@@ -312,29 +277,104 @@ var o_browse = {
...
+            }
+            let slugIndex = $.inArray(slug, opus.prefs.order);
+            // The clicked-on slug should always be in the order list;
+            // The "if" is a safety precaution and the condition should always be true
+            if (slugIndex >= 0) {
+                opus.prefs.order.splice(slugIndex, 1);
+            }
+
+            // remove the sort pill right away
+            $(this).closest("li").remove();
```
dstopp: this is a little bit fuzzy, not terrible but maybe better if the li you need to remove has a class you can better find it on...?

opus/application/static_media/js/browse.js
```diff
@@ -728,7 +810,7 @@ var o_browse = {
             // update the data table w/the new columns
             if (!o_utils.areObjectsEqual(opus.prefs.cols, currentSelectedMetadata)) {
                 o_browse.resetData();
-                o_browse.initTable(opus.col_labels);
+                o_browse.initTable(opus.colLabels, opus.colLabelsNoUnits);
```
dstopp, line 814: it bothers me that we now have two global vars that are almost identical except a bit of text removed - i would rather see the colLabelsNoUnits become a function; if need be for timely sake you can return/set that global til we find a better way but I see these getting out of sync pretty easily.

dstopp (changes requested):

opus/application/static_media/js/browse.js
rfrenchseti: You can't make it a function. That's the whole point. See the discussion Dave and I had above. Only the Django knows what the two text strings should be. He tried to make the JS compute one from the other, and you just can't do it in a portable way that will stand the test of time.

rfrenchseti: I do agree this whole thing is kind of ugly. This is some of the code I was mentioning earlier today that is going to have to be totally rewritten once we allow multiple search terms, so I just want to wait until then to clean it up because I don't know how we will eventually handle it.

opus/application/static_media/js/browse.js
juzen2003: Update "li" to the class ".list-inline-item".

opus/application/static_media/js/browse.js
```diff
@@ -327,7 +327,7 @@ var o_browse = {
             }
 
             // remove the sort pill right away
-            $(this).closest("li").remove();
+            $(this).closest(".list-inline-item").remove();
```
dstopp, line 331: so... what I am trying desperately to avoid is using "closest", as that could change w/future things.  Hence my suggestion to use something a little more opus-definitive that will get you exactly what you want w/out having to use "closest".  Sorry, i should have been clearer in my original comment.

dstopp (approved):

PR #652, opened 2019-04-23
Fix selections and extras and add default qtype

juzen2003:
# Update:
- Add default qtype in url when widgets with qtype is open.
- Modify getSelectionsFromHash to getSelectionsExtrasFromHash.
  - Return format of selections will match opus.selections.
  - Return format of extras will match opus.extras.
- Fix double result count loading issue.
- Clean up indents of getWidget function in widgets.js.

### DIFF
--- a/opus/application/static_media/css/opus.css
+++ b/opus/application/static_media/css/opus.css
@@ -252,9 +252,10 @@ ul.searchMenu {
     border-width: 3px;
 }
 
-.modal {
+/* We have to comment this out to prevent user from clicking navbar link when modal is open */
+/* .modal {
     top: 4%;
-}
+} */
 
 .modal-content {
     border-radius: 1%;
--- a/opus/application/static_media/css/opus.css
+++ b/opus/application/static_media/css/opus.css
@@ -252,11 +252,6 @@ ul.searchMenu {
     border-width: 3px;
 }
 
-/* We have to comment this out to prevent user from clicking navbar link when modal is open */
-/* .modal {
-    top: 4%;
-} */
-
 .modal-content {
     border-radius: 1%;
 }
--- a/opus/application/static_media/js/hash.js
+++ b/opus/application/static_media/js/hash.js
@@ -59,8 +59,7 @@ var o_hash = {
                 }
             }
         );
-        // console.log("hash from hash.js");
-        // console.log(hash);
+
         if (updateURL) {
             window.location.hash = '/' + hash.join('&');
         }
@@ -83,9 +82,7 @@ var o_hash = {
 
     getHashArray: function() {
         let hashArray = [];
-        // this is a workaround for firefox
-        // when user hit enter in input#page, the url hash will be "", we will remove input#page eventually
-        let hashInfo = this.getHash() ? this.getHash() : o_browse.tempHash;
+        let hashInfo = o_hash.getHash();
         $.each(hashInfo.split('&'), function(index, valuePair) {
             let paramArray = valuePair.split("=");
             hashArray[paramArray[0]] = paramArray[1];
--- a/opus/application/static_media/js/hash.js
+++ b/opus/application/static_media/js/hash.js
@@ -107,6 +107,7 @@ var o_hash = {
 
         hash = (hash.search('&') > -1 ? hash.split('&') : [hash]);
         let selections = {};  // the new set of pairs that will not include the result_table specific session vars
+
         $.each(hash, function(index, pair) {
             let slug = pair.split('=')[0];
             let value = pair.split('=')[1];
--- a/opus/application/static_media/js/hash.js
+++ b/opus/application/static_media/js/hash.js
@@ -99,7 +99,7 @@ var o_hash = {
     },
 
     // part is part of the hash, selections or prefs
-    getSelectionsFromHash: function() {
+    getSelectionsExtrasFromHash: function() {
         let hash = o_hash.getHash();
         if (!hash) {
             return;
@@ -107,21 +107,28 @@ var o_hash = {
 
         hash = (hash.search('&') > -1 ? hash.split('&') : [hash]);
         let selections = {};  // the new set of pairs that will not include the result_table specific session vars
+        let extras = {}; // store qtype from url
 
         $.each(hash, function(index, pair) {
             let slug = pair.split('=')[0];
             let value = pair.split('=')[1];
 
             if (!(slug in opus.prefs) && value) {
-                if (slug in selections) {
-                    selections[slug].push(value);
+
+                if (slug.match(/qtype/)) {
+                    extras[slug] = [value];
                 } else {
-                    selections[slug] = [value];
+                    if (slug in selections) {
+                        selections[slug].push(value);
+                    } else {
+                        selections[slug] = [value];
+                    }
                 }
+
             }
         });
 
-        return selections;
+        return [selections, extras];
     },
 
 
--- a/opus/application/static_media/js/hash.js
+++ b/opus/application/static_media/js/hash.js
@@ -116,12 +116,14 @@ var o_hash = {
             if (!(slug in opus.prefs) && value) {
 
                 if (slug.match(/qtype/)) {
+                    // each qtype will only have one value at a time
                     extras[slug] = [value];
                 } else {
                     if (slug in selections) {
                         selections[slug].push(value);
                     } else {
-                        selections[slug] = [value];
+                        // selections[slug] = [value];
+                        selections[slug] = value.replace("+", " ").split(",");
                     }
                 }
 
--- a/opus/application/static_media/js/hash.js
+++ b/opus/application/static_media/js/hash.js
@@ -115,7 +115,7 @@ var o_hash = {
 
             if (!(slug in opus.prefs) && value) {
 
-                if (slug.match(/qtype/)) {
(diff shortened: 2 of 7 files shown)


opus/application/static_media/js/hash.js
```diff
@@ -101,29 +98,39 @@ var o_hash = {
...
             if (!(slug in opus.prefs) && value) {
-                if (slug in selections) {
-                    selections[slug].push(value);
+
+                if (slug.match(/qtype/)) {
+                    // each qtype will only have one value at a time
+                    extras[slug] = [value];
                 } else {
-                    selections[slug] = [value];
+                    if (slug in selections) {
```
rfrenchseti: I don't think this is necessary. You can't have a slug in the hash more than once. This concept is left over from a feature we no longer support. In fact the fact that these are lists at all instead of just values is left over. Some day we should clean them up. But for now you can treat selections slugs just like qtype slugs and not deal with push().

opus/application/static_media/js/opus.js
```diff
@@ -111,9 +112,9 @@ var opus = {
         whether to fire an ajax call.
         */
 
-        let selections = o_hash.getSelectionsFromHash();
+        let [selections, extras] = o_hash.getSelectionsExtrasFromHash();
 
-        if (selections === undefined) {
+        if (selections === undefined && extras === undefined) {
```
rfrenchseti: Maybe this should be ||? If one of them is undefined, how the object comparison handle it? Not that this should ever actually happen.

opus/application/static_media/js/opus.js
```diff
@@ -127,30 +128,21 @@ var opus = {
             $(".op-reset-button button").prop("disabled", true);
         }
 
-        // compare selections and last selections
-        if (o_utils.areObjectsEqual(selections, opus.last_selections)) {
-            // selections have not changed from opus.last_selections
+        // compare selections and last selections, extras and last extras
+        if (o_utils.areObjectsEqual(selections, opus.lastSelections) && o_utils.areObjectsEqual(extras, opus.lastExtras)) {
+            // selections have not changed from opus.lastSelections
```
rfrenchseti: This comment should be updated to include extras

opus/application/static_media/js/opus.js
```diff
@@ -127,30 +128,21 @@ var opus = {
...
             if (!opus.force_load) { // so we do only non-reloading pref changes
                 return;
             }
         } else {
-            // selections in the url hash is different from opus.last_selections
+            // selections in the url hash is different from opus.lastSelections
```
rfrenchseti: Same for this comment

rfrenchseti (changes requested):

opus/application/static_media/js/opus.js
juzen2003: From getSelectionsExtrasFromHash, both selections and extras will be undefined if getHash return an empty string. There won't be a case that only of them is undefined. If one of them doesn't exist in url, it will be an empty object, like [{"planet": ["mars", "neptune"]}, {}] or [{}, {"qtype-time": ["any"]}]

opus/application/static_media/js/hash.js
juzen2003: It's removed in the new commit.

opus/application/static_media/js/opus.js
juzen2003: Comments are updated to include extras.

opus/application/static_media/js/hash.js
```diff
@@ -101,29 +98,39 @@ var o_hash = {
...
+
         $.each(hash, function(index, pair) {
             let slug = pair.split('=')[0];
             let value = pair.split('=')[1];
 
             if (!(slug in opus.prefs) && value) {
-                if (slug in selections) {
-                    selections[slug].push(value);
+
+                if (slug.match(/qtype/)) {
```
rfrenchseti: We really don't need to use "match" here. match matches anywhere in the string. We want to know if the string starts with "qtype-". Thus I think we should be using `slug.startsWith("qtype-")`.

opus/application/static_media/js/opus.js
```diff
@@ -68,34 +69,34 @@ var opus = {
...
 
     // searching - making queries
-    selections:{},        // the user's search
-    extras:{},            // extras to the query, carries units, string_selects, qtypes, size, refreshes result count!!
-    last_selections:{},   // last_ are used to moniter changes
-    last_hash:'',
-    result_count:0,
+    selections: {},        // the user's search
+    extras: {},            // extras to the query, carries units, string_selects, qtypes, size, refreshes result count!!
+    lastSelections: {},   // last_ are used to moniter changes
```
rfrenchseti: "moniter" should be "monitor"

opus/application/static_media/js/opus.js
rfrenchseti: Right...so it's confusing that you are checking for *both* to be true, since it's impossible for one to be true without the other. With any other language, what I'd want to do instead is:
```
if (selections === undefined) {
    assert(extras === undefined);
    ....
}
```
to catch latent or new bugs, but JavaScript doesn't have anything like an assert function. Perhaps we could still do something like this, but instead use `console.error()` to log an error message saying that there was a problem with extras.

opus/application/static_media/js/opus.js
rfrenchseti: This comment (meaning the one after the if) is still wrong. Although frankly I'm not sure there's any reason to have a comment here at all. The code is self-explanatory, and we should never have comments that just restate what the code says, because then they get out of date! Comments should only be for higher-level ideas that are not obvious to the reader.

opus/application/static_media/js/opus.js
rfrenchseti: Same comment here about this comment being wrong.

opus/application/static_media/js/opus.js
```diff
@@ -127,30 +128,21 @@ var opus = {
...
-                // Therefore, we will not put qtype in modifiedSelections to compare with opus.selection
-                if (!slug.match(/qtype/)) {
-                    modifiedSelections[slug] = selections[slug][0].replace("+", " ").split(",");
-                }
-            });
-
-            // if data in selections !== data in opus.selections, it means selections are modified manually in url, reload the page (modified url in url bar and hit enter)
-            if (!o_utils.areObjectsEqual(modifiedSelections, opus.selections)) {
-                opus.selections = modifiedSelections;
+            // if data in selections !== data in opus.selections or extras !== data in opus.extras, it means selections/qtype are modified manually in url, reload the page (modified url in url bar and hit enter)
```
rfrenchseti, line 146: This is a good comment, because it tells the read *why* this event might happen.

opus/application/static_media/js/opus.js
```diff
@@ -214,7 +208,7 @@ var opus = {
         o_search.validateRangeInput(normalizedData, true);
 
         // query string has changed
-        // opus.last_selections = selections;
+        // opus.lastSelections = selections;
```
rfrenchseti: Do we need this line anymore? It can't work even if uncommented.

rfrenchseti (changes requested):
Didn't hit request changes earlier by mistake

rfrenchseti (approved):

opus/application/static_media/css/opus.css
```diff
@@ -252,9 +252,10 @@ ul.searchMenu {
     border-width: 3px;
 }
 
-.modal {
+/* We have to comment this out to prevent user from clicking navbar link when modal is open */
```
dstopp: if we are commenting it out and no longer using it, why do we leave it here...?

opus/application/static_media/js/widgets.js
```diff
@@ -394,198 +394,178 @@ var o_widgets = {
...
-                    $('.' + widget + ' .widget-main>ul>ul').append('<li class = "range-qtype-helper">' + help_icon + '</li>');
+            }
+        }).done(function() {
+            // If there is no specified qtype in the url, we want default qtype to be in the url
+            // This will also put qtype in the url when a widget with qtype is open.
+            // Need to wait until api return to determine if the widget has qtype selections
+            let hash = o_hash.getHashArray();
+            let qtype = "qtype-" + slug;
+            if ($(`#widget__${slug} select[name="${qtype}"]`).length !== 0 && !hash[qtype]) {
+                let defaultOption = $(`#widget__${slug} select[name="${qtype}"]`).first("option").val();
```
dstopp, line 410: it's not clear to me why you are getting .first("option").

dstopp (changes requested):

opus/application/static_media/css/opus.css
rfrenchseti: BTW, Dave and I were talking about this the other day. I think you added this in so the modal dialog would be slightly further down on the screen. But if it's there, then you can still click in the 4% region at the top, allowing you to do things like change tabs. If we want the modal to be further down, we need a different way to do it.

opus/application/static_media/js/widgets.js
juzen2003: It's because when we open a widget with qtype, the value of the first option tag (under select tag) is the default displayed value for qtype.

dstopp (approved):

PR #653, opened 2019-04-26
New ui startobs browse ps

juzen2003:
# This is a dummy pull-request to track the changes of PS in browse view
# Update:
- rename browse_view to op-browse-view
- Move PS for browse gallery view to .op-gallery-view (container)
    - Under .gallery-contents:
        - PS for gallery view is attached to .op-gallery-view
        - PS for table view is attached to .dataTable (same as original implementation)
    - Update mutationObserver for switching between gallery and table view
# TODO:
- Fix the scrollbar size in gallery view
- Instantiate infiniteScroll on both gallery view and table view
- Display spinner when infiniteScroll is loading in gallery view

### DIFF
--- a/opus/application/apps/ui/templates/base.html
+++ b/opus/application/apps/ui/templates/base.html
@@ -98,7 +98,7 @@
                         <div class="browse-nav-container top_navbar collapse navbar-collapse">
                             <ul class="navbar-nav mr-auto">
                                 <li class="nav-item">
-                                    <a href="#" class="browse_view nav-link"><i class="far fa-list-alt"></i></a>
+                                    <a href="#" class="op-browse-view nav-link"><i class="far fa-list-alt"></i></a>
                                 </li>
                                 <li class="nav-item">
                                     <a href="#" class="metadataModal nav-link" data-toggle="modal" data-target="#metadataSelector" title="Select metadata to display or include in CSV files for each observation"><i class="fas fa-columns"></i>&nbsp;Select Metadata</a>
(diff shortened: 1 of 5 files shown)


opus/application/apps/ui/templates/base.html
```diff
@@ -122,7 +122,7 @@
                         </div>
                     </nav>
                     <div class="panel-body gallery-contents container-fluid browse-infiniteScroll-element">
-                        <div class="row justify-content-start op-height-0 op-gallery-view">
+                        <div class="row justify-content-start op-gallery-view mr-0">
```
dstopp, line 125: go ahead and remove op-height-0 from opus.css, as i'm pretty sure this was the only place it was used

dstopp (changes requested):

juzen2003:
# Update:
- Modify infiniteScroll selector from "$(`${tab} .gallery-contents`).infiniteScroll(..."  to "$(`${tab} ${contentsView}`).infiniteScroll(..." where contentsView is either ".op-gallery-view" or ".dataTable".
- Will work on instantiation of infiniteScroll in another branch

opus/application/static_media/js/browse.js
```diff
@@ -62,16 +62,17 @@ var o_browse = {
         $(".gallery-contents, .dataTable").on('wheel ps-scroll-up', function(event) {
             let startobs = (opus.prefs.view === "cart" ? "cart_startobs" : "startobs");
             let tab = `#${opus.prefs.view}`;
+            let contentsView = opus.prefs.browse === "gallery" ? ".op-gallery-view" : ".dataTable";
```
dstopp: Thinking on this - seems like you should just create a single unified class, like "op-scroll-container" and then you don't need to do this if so many places.  My first thought was to create a function that returns the class since you are using it multiple places, but i think a unified class is a better way to go.

dstopp (changes requested):

PR #658, opened 2019-04-30
Add back __api/data.csv, which is used internally. Add HTML download …

rfrenchseti:
- Add back __api/data.csv, which is used internally.
- Add HTML download tag for downloading CSV/archive files and remove target="_blank" in those cases.
- Initially hide internal URLs from user.
- Fix bug in qtypes introduced by last pull.
- Make Ctrl+click and Shift+click on detail (i) properly fill in URL. (B045)
- Other random cleanup and bug fixes.

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -5,7 +5,7 @@
 /* jshint varstmt: true */
 /* globals $, _, PerfectScrollbar */
 /* globals o_cart, o_hash, o_menu, o_utils, opus */
-/* globals default_pages, default_columns */
+/* globals default_columns */
 
 // font awesome icon class
 const pillSortUpArrow = "fas fa-arrow-circle-up";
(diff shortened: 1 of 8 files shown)


opus/application/static_media/js/browse.js
```diff
@@ -102,17 +102,22 @@ var o_browse = {
...
-        $("#browse").on("click", ".download_csv", function() {
-            let col_str = opus.prefs.cols.join(',');
-            let hash = [];
+        $("#browse").on("click", ".op-download-csv", function() {
+            let colStr = opus.prefs.cols.join(',');
+            let selectionsHash = [];
             for (let param in opus.selections) {
                 if (opus.selections[param].length) {
-                    hash[hash.length] = param + '=' + opus.selections[param].join(',').replace(/ /g,'+');
+                    selectionsHash.push(param + '=' + opus.selections[param].join(',').replace(/ /g,'+'));
```
dstopp: might as well change this one to `${}` format as well

dstopp (changes requested):

opus/application/static_media/js/browse.js
rfrenchseti: I didn't want to put that whole complicated replace thing inside a {}. I can make it another variable.

dstopp (approved):

PR #659, opened 2019-04-30
New ui startobs infinite scroll down direction

juzen2003:
# Update
- This is branched out from new_ui_startobs_browse_ps
- InfiniteScroll scroll down is implemented.
- Because we instantiate two infiniteScroll instances, in checkScroll, we only call updateSliderHandle. We don't need the extra logic to trigger loadNextPage for table, loadNextPage will be triggered in scrollThreshold event handler (this is setup when instantiating infiniteScroll instance on .dataTable).
 
# Todo:
- Display spinner in while infiniteScroll load event is still happening

### DIFF
--- a/opus/application/apps/ui/templates/base.html
+++ b/opus/application/apps/ui/templates/base.html
@@ -98,7 +98,7 @@
                         <div class="browse-nav-container top_navbar collapse navbar-collapse">
                             <ul class="navbar-nav mr-auto">
                                 <li class="nav-item">
-                                    <a href="#" class="browse_view nav-link"><i class="far fa-list-alt"></i></a>
+                                    <a href="#" class="op-browse-view nav-link"><i class="far fa-list-alt"></i></a>
                                 </li>
                                 <li class="nav-item">
                                     <a href="#" class="metadataModal nav-link" data-toggle="modal" data-target="#metadataSelector" title="Select metadata to display or include in CSV files for each observation"><i class="fas fa-columns"></i>&nbsp;Select Metadata</a>
--- a/opus/application/apps/ui/templates/base.html
+++ b/opus/application/apps/ui/templates/base.html
@@ -122,7 +122,7 @@
                         </div>
                     </nav>
                     <div class="panel-body gallery-contents container-fluid browse-infiniteScroll-element">
-                        <div class="row justify-content-start op-gallery-view mr-0">
+                        <div class="row justify-content-start op-gallery-view mr-0 op-scroll-container">
                             <div class="mr-0 thumb gallery gallery-scroll"></div>
                             <div class="page-load-status">
                                 <div class="infinite-scroll-request loader-ellips">
@@ -132,7 +132,7 @@
                                 <p class="infinite-scroll-error">No more pages to load</p>
                             </div>
                         </div>
-                        <div class="row dataTable browse-infiniteScroll-element mx-0">
+                        <div class="row dataTable browse-infiniteScroll-element mx-0 op-scroll-container">
                             <div class="col-lg p-0">
                             <!-- <div class="col-lg pr-0"> -->
                                 <!-- <div class="op-page-loading-status">
--- a/opus/application/apps/ui/templates/base.html
+++ b/opus/application/apps/ui/templates/base.html
@@ -124,20 +124,9 @@
                     <div class="panel-body gallery-contents container-fluid browse-infiniteScroll-element">
                         <div class="row justify-content-start op-gallery-view mr-0">
                             <div class="mr-0 thumb gallery gallery-scroll"></div>
-                            <!-- <div class="page-load-status">
-                                <div class="infinite-scroll-request loader-ellips">
-                                    <div class='loader'></div>
-                                </div>
-                                <p class="infinite-scroll-last">End of content</p>
-                                <p class="infinite-scroll-error">No more pages to load</p>
-                            </div> -->
                         </div>
                         <div class="row op-dataTable-view browse-infiniteScroll-element mx-0">
                             <div class="col-lg p-0">
-                            <!-- <div class="col-lg pr-0"> -->
-                                <!-- <div class="op-page-loading-status">
-                                    <div class="loader"></div>
-                                </div> -->
                                 <table id="dataTable" class="table table-hover table-bordered table-striped table-dark table-sm op-noSelect">
                                     <thead>
                                         <tr></tr>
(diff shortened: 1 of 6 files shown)


opus/application/static_media/js/browse.js
```diff
@@ -1209,21 +1185,88 @@ var o_browse = {
...
+            });
+            $(selector).on("load.infiniteScroll", o_browse.infiniteScrollLoadEventListener);
+        }
+    },
+
     loadData: function(startObs) {
         let view = o_browse.getViewInfo();
-        let selector = `${view.namespace} .gallery-contents`;
+        let contentsView = o_browse.getScrollContainerClass();
+        let selector = `${view.namespace} ${contentsView}`;
```
dstopp: same comment here re: getViewInfo.  At some point, we will be able to get rid of that function if we keep converting it.

opus/application/static_media/js/browse.js
```diff
@@ -1209,21 +1185,88 @@ var o_browse = {
...
+    // return the infiniteScroll container class for either gallery or table view
+    // NOTE: the reason we don't want to use a single unified class is because we have two infiniteScroll instances.
+    // If they share a same class say "op-scroll-container", then later on $(selector).infiniteScroll("loadNextPage") will call load event handler twice (selector selects both container) and render the data twice (duplicated data).
+    getScrollContainerClass: function() {
+        return (opus.prefs.browse === "gallery" ? ".op-gallery-view" : ".dataTable");
+    },
+
+    // Instantiate infiniteScroll
+    initInfiniteScroll: function(selector) {
+        let view = o_browse.getViewInfo();
```
dstopp: suggestion that we stop using getViewInfo and use opus.prefs directly; example:
            let startobsLabel = (opus.prefs.view === "cart" ? "cart_startobs" : "startobs");
            if (opus.prefs[startobsLabel] > 0) {

opus/application/static_media/js/hash.js
```diff
@@ -155,6 +155,10 @@ var o_hash = {
                         case "order":
                             opus.prefs[slug] = value.split(',');
                             break;
+                        case "startobs" || "cart_startobs":
```
dstopp: better case syntax, vs. using an OR:
case "startobs":
case "cart_startobs":
     doStuffHere;
     break;

opus/application/static_media/js/mutationObserver.js
```diff
@@ -190,8 +201,9 @@ var o_mutationObserver = {
         let metadataSelector = $("#metadataSelector")[0];
         let metadataSelectorContents = $("#metadataSelectorContents")[0];
         let browseDialogModal = $("#galleryView.modal")[0];
-        let galleryView = $(".op-gallery-view")[0];
+        let galleryView = $(".gallery")[0];
```
dstopp, line 204: consider using $(".gallery").first(), etc

dstopp (changes requested):

juzen2003:
# Update:
- Stop using getViewInfo in loadData and initInfiniteScroll
- Rename .dataTable to .op-dataTable-view
  - Now we can use let selector = `${tab} .op-${opus.prefs.browse}-view`;
  - Remove getScollContainer function

opus/application/static_media/js/mutationObserver.js
juzen2003: I think we need to keep using $(".gallery")[0] because we have to pass in a Node to mutationObserver. If we use $(".gallery").first(), it will complain about the passed in $(".gallery").first() is not of type Node.

opus/application/static_media/js/hash.js
```diff
@@ -155,10 +155,14 @@ var o_hash = {
                         case "order":
                             opus.prefs[slug] = value.split(',');
                             break;
-                        case "startobs" || "cart_startobs":
+                        case "startobs":
```
dstopp, line 158: you can make this a fall-through case so you don't have to duplicate code, that's why I suggested the code:
case "startobs":
case "cart_startobs":
    do stuff();
    break;

Doing it that way makes it clear to all that read that those two cases are handled the same.  So it's not that your code is wrong, but if they are always the same, it will be clearer in the long run to make it obvious in the code.

opus/application/static_media/js/mutationObserver.js
dstopp: i wondered that; thanks for explaination.

dstopp (changes requested):

opus/application/static_media/js/hash.js
juzen2003: New commit has been updated to have fall-through case as suggested, thanks for the feedback.

opus/application/static_media/js/hash.js
rfrenchseti: Just for the record, the original code won’t work at all. The || on two strings will yield true so it was just equivalent to “case true:” which is not at all what you wanted.

opus/application/apps/ui/templates/base.html
```diff
@@ -124,13 +124,13 @@
                     <div class="panel-body gallery-contents container-fluid browse-infiniteScroll-element">
                         <div class="row justify-content-start op-gallery-view mr-0">
                             <div class="mr-0 thumb gallery gallery-scroll"></div>
-                            <div class="page-load-status">
+                            <!-- <div class="page-load-status">
```
dstopp: just delete the code vs. comment out...

opus/application/static_media/js/browse.js
```diff
@@ -1235,7 +1241,8 @@ var o_browse = {
             $(selector).on("request.infiniteScroll", function(event, path) {
                 // hide default page status loader if op-page-loading-status loader is spinning
                 // && o_browse.tableSorting
-                $(".infinite-scroll-request").hide();
+                // Comment this out so that we will have spinner displayed
+                // $(".infinite-scroll-request").hide();
```
dstopp, line 1245: is this going away forever....?

opus/application/static_media/js/browse.js
```diff
@@ -1298,6 +1306,8 @@ var o_browse = {
 
                 $(`${tab} .op-gallery-view`).scrollTop(0);
                 $(`${tab} .op-dataTable-view`).scrollTop(0);
+                // $(`${selector}`).scrollTop(0);
```
dstopp: what are these guys here for?

dstopp (changes requested):

opus/application/apps/ui/templates/base.html
juzen2003: Commented out codes are removed.

opus/application/static_media/js/browse.js
juzen2003: Commented out codes are removed.

opus/application/static_media/js/browse.js
juzen2003: I keep those codes commented because those codes were there for B061. But in startobs branch, that line of code breaks the infiniteScroll spinner. For now, I comment it out, and will revisit B061 to confirm if we can remove those codes once infiniteScroll issue is fixed.

dstopp (approved):

PR #661, opened 2019-05-02
b078 - deallocate memory/element cache if it gets too big.  Current d…

dstopp:
b078 - deallocate memory/element cache if it gets too big. Current default is 1000 observations.

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -36,6 +36,7 @@ var o_browse = {
     galleryBegun: false, // have we started the gallery view
     galleryData: {},  // holds gallery column data
     imageSize: 100,     // default
+    maxCachedObservations: 1000,    // max number of obserations to store in cache; at some point, we can probably figure this out dynamically
 
     limit: 100,  // results per page
     lastLoadDataRequestNo: 0,
@@ -1006,7 +1007,6 @@ var o_browse = {
         } else {
             let startobsLabel = o_browse.getStartObsLabel();
             let append = (data.start_obs > $(`${tab} .thumbnail-container`).last().data("obs"));
-            console.log(`append: ${append}`);
 
             o_browse.manageObservationCache(data.count, append);
             $(`${tab} .navbar`).show();
@@ -1205,8 +1205,8 @@ var o_browse = {
     // Two functions; one to delete single elements, just a tool for the main one, manageObservationCache, to loop.
     manageObservationCache: function(count, append) {
         let tab = `#${opus.prefs.view}`;
-        var galleryObsElem =  $(`${tab} .gallery [data-obs]`);
-        var tableObsElem = $(`${tab} .op-dataTable-view [data-obs]`);
+        let galleryObsElem = $(`${tab} .gallery [data-obs]`);
+        let tableObsElem = $(`${tab} .op-dataTable-view [data-obs]`);
 
         function deleteCachedObservation(index) {
             // don't delete the metadata if the observation is in the cart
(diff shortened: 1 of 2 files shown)


opus/application/static_media/js/browse.js
```diff
@@ -999,18 +997,24 @@ var o_browse = {
                     galleryHtml += 'then click on the cart icon.  </p>';
                     galleryHtml += '</div>';
                 }
-                $(".gallery", namespace).html(galleryHtml);
+                $(".gallery", tab).html(galleryHtml);
```
rfrenchseti, line 1001: I would really like to see us break the direct connection between the URL and the DOM class. For example, the class should really be op-browse or op-cart or whatever. We should have a function that looks up the prefs.view and returns the appropriate class name.

opus/application/static_media/js/browse.js
```diff
@@ -999,18 +997,24 @@ var o_browse = {
...
                 // we've hit the end of the infinite scroll.
                 $(".op-page-loading-status > .loader").hide();
                 return;
             }
         } else {
-            $("#cart .navbar").show();
-            $("#cart .sort-order-container").show();
+            let startobsLabel = o_browse.getStartObsLabel();
+            let append = (data.start_obs > $(`${tab} .thumbnail-container`).last().data("obs"));
+            console.log(`append: ${append}`);
```
rfrenchseti: Remove debugging log.

opus/application/static_media/js/browse.js
```diff
@@ -1196,6 +1200,53 @@ var o_browse = {
...
 
+    // check the cache of both rendered elements and variable o_browse.galleryData; remove 'far away' observations
+    // from cache to avoid buildup of too much data in the browser which slows things down
+    // Two functions; one to delete single elements, just a tool for the main one, manageObservationCache, to loop.
+    manageObservationCache: function(count, append) {
+        let tab = `#${opus.prefs.view}`;
+        var galleryObsElem =  $(`${tab} .gallery [data-obs]`);
+        var tableObsElem = $(`${tab} .op-dataTable-view [data-obs]`);
+
+        function deleteCachedObservation(index) {
```
rfrenchseti, line 1211: So how does this work for obs that are now in the cart but in the future we remove them from the cart? Do they end up sitting around forever because there's no reference to them in the gallery and thus we will never try to remove them?

opus/application/static_media/js/browse.js
```diff
@@ -1196,6 +1200,53 @@ var o_browse = {
...
+
+        let beginningIndex = galleryObsElem.first().index();
+        let endingIndex = galleryObsElem.last().index();
+        if (beginningIndex < 0 || endingIndex < 0) {
+            // this only happens when there are no elements rendered, so why bother...
+            // probably don't need to check both, but why not...
+            return;
+        }
+
+        let totalObs = endingIndex - beginningIndex;
```
rfrenchseti: Actually total # of observations is ending - beginning + 1, right?

opus/application/static_media/js/browse.js
```diff
@@ -1196,6 +1200,53 @@ var o_browse = {
...
+        if (totalObs + count > opus.maxBufferSize) {
+            // if we are appending, remove from the top
+            console.log(`before: ${Object.keys(o_browse.galleryData).length}`);
```
rfrenchseti: Remove debugging log.
Might be easier to read by using Math.min instead of the ?: operator.

opus/application/static_media/js/browse.js
```diff
@@ -1196,6 +1200,53 @@ var o_browse = {
...
+            count = (count > endingIndex ? endingIndex : count);
+            if (append) {
+                // this is theoretically the faster way to delete lots of data, as jquery selector eval is slow
+                for (let index = 0; index < count ; index++) {
+                    deleteCachedObservation(index);
+                }
+            } else {
+                let deleteTo = endingIndex - count;
+                deleteTo = (deleteTo >= 0 ? deleteTo : 0);
```
rfrenchseti: I think this has an off-by-one error. It should be endingIndex-count+1. That way if count=1, then we go from endingIndex to endingIndex and remove one element.

Math.max(endingIndex-count+1, 0) might be easier to read than using ?:

opus/application/static_media/js/browse.js
```diff
@@ -1196,6 +1200,53 @@ var o_browse = {
...
+                for (let index = endingIndex; index >= deleteTo; index--) {
+                    deleteCachedObservation(index);
+                }
+            }
+            console.log(`after: ${Object.keys(o_browse.galleryData).length}`);
```
rfrenchseti: Remove debugging

opus/application/static_media/js/opus.js
```diff
@@ -55,11 +55,12 @@ var opus = {
...
     // searching - making queries
-    selections:{},        // the user's search
-    extras:{},            // extras to the query, carries units, string_selects, qtypes, size, refreshes result count!!
+    selections: {},        // the user's search
+    extras: {},            // extras to the query, carries units, string_selects, qtypes, size, refreshes result count!!
     lastSelections: {},   // last_ are used to monitor changes
     lastExtras: {},
-    resultCount:0,
+    resultCount: 0,
+    maxBufferSize: 1000,    // max number of obserations to store in cache; at some point, we can probably figure this out dynamically
```
rfrenchseti: I think "buffer" is too generic a name for this. The can be lots of buffers. How about maxCachedObservations or something like that?

rfrenchseti (changes requested):

opus/application/static_media/js/browse.js
dstopp: do you want that sort of major change to be part of this change?  it's a really wide-spread modification.

rfrenchseti: If this is a common pattern we can delay replacing it.

opus/application/static_media/js/browse.js
dstopp: good point.  I suppose we have several choices:
-leave them hanging around <seems not fabulous but truthfully may not matter in the bigger scheme of things>
-when we remove things from the cart, check to see if these are orphaned observations and remove if so <is it really worth the trouble....?  ..premature optimization comes to mind>
-create a separate global var for the cart, so the info is managed independently.  If we did this, then we might as well allow different sorting for browse vs. cart.

opus/application/static_media/js/browse.js
dstopp: nope; i changed this so it's 0-based when i changed to 'index' vs. 'obsNumber'.

opus/application/static_media/js/opus.js
dstopp: ah good suggestion; i didn't like the name but couldn't come up w/something better in the moment.

opus/application/static_media/js/browse.js
rfrenchseti: It doesn't matter if it's zero-based. End-start is still off by 1 for the total number of obs.
0 1 2 3
3-0=3 but there are 4 obs

rfrenchseti (approved):

PR #662, opened 2019-05-02
New ui startobs maintain previous scroll position

juzen2003:
# Update:
- obsNum in slider:
    - In gallery view, the item with obsNum in slider will be in one of the top row thumbnail containers.
    - In table view, the item with obsNum in slider will be the top item.
- When scrolling in gallery view, obsNum in slider will be the same as the top left item in gallery view.
    - click "View Table" to switch to table view, the top item in table will have the same obsNum.
- When scrolling in table view, obsNum in slider will be the same as the top item in table.
    - click "View Gallery" to switch to gallery view, the item with obsNum will be somewhere in one of the first row thumbnail containers.
# Test case:
- Case 1:
    - Fresh start OPUS, go to browse gallery view, and scroll down until obsNum is 29. The top left item in gallery has data-obs = 29.
    - Go to table view by clicking "View Table", obsNum is still 29 and the top item in table has data-obs = 29.
    - Go to detail page and switch back to browse table view, the top item in table is the same with data-obs = 29.
    - Go to gallery view by clicking "View Gallery", the top left item in gallery is the one with data-obs = 29.
    - Go to search page and switch back to browse gallery view, the top left item in gallery is the same with data-obs = 29.

- Case 2: 
    - Fresh start OPUS, go to browse gallery view, and go to obsNum 4649 using slider. The top left item in gallery has data-obs = 4649.
    - Go to table view, the top item in table has data-obs = 4649. Use scrollbar to scroll down to 4867, the top item in table has data-obs = 4867.
    - Go to gallery view, item with data-obs = 4867 is in the top row (but not at the top left corner). 
    - Do the similar test in case 1, scrollbar will be in the same position where item with data-obs = 4867 is at the top in table view and in the first row of gallery view.
    - In gallery view, use scrollbar to scroll down to 4915, the top left item has data-obs = 4915.
    - Go to table view, the top item has data-obs = 4915.
    - Do the similar test in case 1, scrollbar will be in the same position where item with data-obs = 4915 is at the top in table view and at top left in gallery view.


dstopp:
Notes: I propose that the gallery work as it always does. As you scroll the table the slider updates on gallery row boundaries only. The slider step is always the number of obs in a gallery row. And when you drag the slider no matter what you’re looking at it moves the table to that obs and puts that obs in the gallery top left.
This is definitely not what Dave described as the current behavior but hopefully it’s easy to implement. You just need to update the slider rounding down the obs number from the table. Whatever the JS equivalent is of
Floor(obsnum / rowsize) * rowsize (maybe adjusted for a 1 start instead of 0)

opus/application/static_media/js/browse.js
```diff
@@ -599,12 +599,14 @@ var o_browse = {
         // trigger an artificial DOMMouseScroll event with delta -650
         //$( ".gallery-contents" ).trigger( e );
```
dstopp, line 600: what are we going to do w/this bit of comment?

opus/application/static_media/js/browse.js
```diff
@@ -624,12 +626,26 @@ var o_browse = {
 
     // find the first displayed observation index & id in the upper left corner
     updateSliderHandle: function() {
+        let tab = `#${opus.prefs.view}`;
+        let contentsView = o_browse.getScrollContainerClass();
```
dstopp, line 630: curiously, is the contenstView selector for the PS not the same as the infinite scroll container?

opus/application/static_media/js/browse.js
```diff
@@ -624,12 +626,26 @@ var o_browse = {
...
         let selector = (opus.prefs.browse === "dataTable") ? `#${opus.prefs.view} #dataTable tbody tr` : `#${opus.prefs.view} .gallery .thumbnail-container`;
+
+        // Fot gallery view, the topBoxBoundary is the top of .gallery-contents
+        // For table view, we will set the topBoxBoundary to be the bottom of thead (account for height of thead)
+        let topBoxBoundary = (opus.prefs.browse === "dataTable") ? $(".gallery-contents").offset().top + $(`${tab} #dataTable thead th`).outerHeight() : $(".gallery-contents").offset().top;
```
dstopp, line 635: is the top of the box not the same as the PS container or the infinite scroll container?  also note that this code is not tab-specific.

opus/application/static_media/js/browse.js
```diff
@@ -624,12 +626,26 @@ var o_browse = {
...
+
+        let currentObsNum = $(`${tab} ${contentsView}`).data("infiniteScroll").options.obsNum;
```
dstopp: I assume the infinite scroll object has been instantiated by this time so you will not throw an exception...?

opus/application/static_media/js/browse.js
```diff
@@ -642,6 +658,13 @@ var o_browse = {
                     "step": o_browse.gallerySliderStep,
                     "max": opus.resultCount,
                 });
+
+                // update obsNum in both infiniteScroll instances
+                // store the most top left obsNum in gallery for gallery infiniteScroll instance
+                // store the most top obsNum in table for table infiniteScroll instance
+                $(`${tab} .op-gallery-view`).infiniteScroll({"obsNum": obsNum});
+                $(`${tab} .op-dataTable-view`).infiniteScroll({"obsNum": obsNum});
```
dstopp: based on notes in review, i suspect this will change.
You just need to update the slider rounding down the obs number from the table. Whatever the JS equivalent is of
Floor(obsnum / rowsize) * rowsize (maybe adjusted for a 1 start instead of 0)

dstopp (changes requested):

juzen2003:
# Update:
In gallery view: 
1. If we scroll the scrollbar, slider obsNum will be the same as top left item in gallery.
2. If we scroll the slider, slider obsNum will be the same as top left item in gallery.
3. When switch to table view, the top item will have the same obsNum.

In table view: 
1. If we scroll the scrollbar, slider obsNum will ONLY get updated on gallery row boundaries. (which means slider obsNum will get updated when current top item in table has the same obsNum as the top left item in gallery)
2. If we scroll the slider, the top item in table has the same obsNum as the slider. 
3. When switch to gallery view, we want the same obsNum to be at the top left item.

opus/application/static_media/js/browse.js
juzen2003: It's the same as the infinite scroll container. (".op-gallery-view" and ".op-dataTable-view")

opus/application/static_media/js/browse.js
juzen2003: This is the old comment, I'm not quite sure about what that TODO comment is. Maybe we can remove that? (Because now setScrollbarOnSlide is setting the scrollbar position for both gallery and table view)

opus/application/static_media/js/browse.js
juzen2003: The box (.gallery-contents) are the same for both gallery and table. But in table view, there is that one row of sticky table header which will always cover one item of the table as we scroll down, so we have to set the top boundary to be the bottom of the table header. That way we can always get the top item of the table seen by the user.

dstopp: i believe .gallery-contents is both in #browse and #cart, no?

dstopp (changes requested):
Note: I moved the slider to the middle and could not scroll up.  Does this review have 'up' working?

juzen2003:
> Note: I moved the slider to the middle and could not scroll up. Does this review have 'up' working?

Scroll up (B034) is not included in this review. I have scroll up in another local branch, I'll create a pull request for that branch tomorrow. Scroll up branch still has the following in progress:
- Fix the spinner behavior of infiniteScroll load at both top and bottom ends.
- When we scroll the slider, prefetch some data ahead of current obsNum so scrollbar will be at the middle.

PR #668, opened 2019-05-08
New ui startobs infiniteScroll up/down direction (B034), maintain scrollbar position (A048)

juzen2003:
**This pull request has the following items:**
- InfiniteScroll scroll up/down direction (B034)
- scrollbar position for gallery and table view (A048)
Note: it has all the changes in pull request #662


opus/application/static_media/js/browse.js
```diff
@@ -38,15 +37,15 @@ var o_browse = {
     imageSize: 100,     // default
     maxCachedObservations: 1000,    // max number of obserations to store in cache; at some point, we can probably figure this out dynamically
 
-    limit: 100,  // results per page
+    // set default to 200 so loadData will fetch enough number of data for the first time in large screen
```
rfrenchseti, line 40: We should get rid of this default limit entirely. There's no reason countGalleryImages() can't be called in the beginning to set the limit to the proper value.

opus/application/static_media/js/browse.js
```diff
@@ -1260,70 +1325,72 @@ var o_browse = {
...
 
-                    console.log(`${selector} in path - startObs: ${startObs}`);
                     let infiniteScrollData = $(selector).data("infiniteScroll");
                     if (infiniteScrollData !== undefined && infiniteScrollData.options.loadPrevPage === true) {
-                        // Direction: scroll up
-                        console.log(`loadPrevPage = true`);
-                        infiniteScrollData.options.loadPrevPage = false;
+                        // Direction: scroll up, we prefetch 1 * o_browse.getLimit() items
+                        if (startObs !== 1) {
+                            let originalStartObs = $(`${tab} .thumbnail-container`).first().data("obs");
```
rfrenchseti: We shouldn't call these "StartObs". StartObs is a specific thing - the slug that indicates the top left corner. Just call it "ObsNum".

opus/application/static_media/js/browse.js
```diff
@@ -1260,70 +1325,72 @@ var o_browse = {
...
+                            // prefetch o_browse.getLimit() items ahead of current obsNum
+                            let prevStartObs = originalStartObs - o_browse.getLimit();
+                            startObs = prevStartObs > 0 ? prevStartObs : 1;
```
rfrenchseti: This can be Math.max. I thought we changed this while I was in your cube?

opus/application/static_media/js/browse.js
```diff
@@ -57,25 +56,31 @@ var o_browse = {
...
+        // - Lower the scroll threshold point in infiniteScroll.
+        // - Adding wheel event: because wheel event can (ps-scroll-up can't) trigger another loadNextPage with correct url if we reach to the top end in that corner case.
+        $(".op-gallery-view, .op-dataTable-view").on("wheel ps-scroll-up", function(event) {
+            if (event.originalEvent.type === "ps-scroll-up" || (event.originalEvent.type === "wheel" && event.originalEvent.deltaY < 0)) {
+                let startObsLabel = o_browse.getStartObsLabel();
+                let tab = `#${opus.prefs.view}`;
+                let contentsView = o_browse.getScrollContainerClass();
+
+                if (opus.prefs[startObsLabel] > 0) {
+                    let fristObs = $(`${tab} [data-obs]`).first().data("obs");
```
rfrenchseti: Do you mean firstObs?

rfrenchseti (changes requested):

juzen2003:
# Note:
- Previous issue: When we keep scrolling down, at the point of cleaning up cached observations, the top 1 * getLimit() (or the same number of returned data appended to the bottom) observations will be removed. We didn't manually set the scrollbar position for scroll down, and let the infiniteScroll to take care of it, and this will move the scrollbar all the way to the very bottom. (theoretically, scrollbar should bounce back when 1 * getLimit() items are appended to the end, but because at that time we also have 1 * getLimit() items get cleaned up at the top, the scrollbar position will stick to the bottom and break the scroll down behavior)
- Now we manually set the scrollbar position in renderGalleryAndTable to deal with the following cases:
    - scroll up: when we scroll up and a new page is fetched, we want to keep scrollbar position at the current startObs, instead of at the first item in newly fetched page.
    - scroll slider: when we load 3 * getLimit items, we want to keep scrollbar in the middle page.
    - scroll down: theoretically, infiniteScroll will take care of scrollbar position, but we still have to manually set it for the case when cached data is removed so that scrollbar position is always correct (and never reaches to the end until it reaches to the end of the data)
- Another observation on scrollbar behavior, when rendered data reach to 1000 (current default maxCachedObservations is 1000), it will sometimes be laggy when scrolling, and changing maxCachedObservations to 500, it feels much more smoothly. This could be because in every scroll (debounce with 0.5s), we are looping through every item to found out the startObs in current page in updateSliderHandle. (When number of items rendered are doubled, the looping time is also doubled)

opus/application/static_media/js/browse.js
```diff
@@ -38,7 +38,7 @@ var o_browse = {
     maxCachedObservations: 1000,    // max number of obserations to store in cache; at some point, we can probably figure this out dynamically
 
     // set default to 200 so loadData will fetch enough number of data for the first time in large screen
-    limit: 200,  // results per page
+    // limit: 200,  // results per page
```
rfrenchseti, line 41: You can remove these commented-out limit lines. It appears to work fine.

opus/application/static_media/js/browse.js
```diff
@@ -1325,31 +1326,44 @@ var o_browse = {
         if (!$(selector).data("infiniteScroll")) {
             $(selector).infiniteScroll({
                 path: function() {
-                    let startObs = opus.prefs[startObsLabel];
+                    let obsNum = opus.prefs[startObsLabel];
```
rfrenchseti, line 1329: These variable names seem confusing to me. Isn't opus.prefs[startObsLabel] the real startobs as specified by the slug? The original variable name of startObs was correct. But then in a minute you have originalStartObs be the first() element in the thumbnail container. But isn't that really the first obs we have cached? It should be called something like firstCachedObs. Again the phrase "startobs" should be reserved for only the actual startobs number from the hash.

opus/application/static_media/js/browse.js
```diff
@@ -57,25 +56,32 @@ var o_browse = {
...
-                    // $(`${tab} ${contentsView}`).infiniteScroll({
-                    //     "loadPrevPage": true
-                    // });
-                    // $(`${tab} ${contentsView}`).infiniteScroll("loadNextPage");
+
+            if (event.originalEvent.type === "ps-scroll-up" || (event.originalEvent.type === "wheel" && event.originalEvent.deltaY < 0)) {
+
+                if (opus.prefs[startObsLabel] > 0) {
+                    let firstObs = $(`${tab} [data-obs]`).first().data("obs");
+                    if ($(`${tab} ${contentsView}`).scrollTop() < 100 && firstObs !== 1) {
```
rfrenchseti, line 78: Magic constants like "100" should be separated out. Can you make these a const declaration somewhere?

opus/application/static_media/js/browse.js
```diff
@@ -613,23 +620,61 @@ var o_browse = {
...
+            $(`${tab} .op-gallery-view`).infiniteScroll({"obsNum": value});
+            $(`${tab} .op-dataTable-view`).infiniteScroll({"obsNum": value});
         } else {
-            o_browse.loadData(value);
+            // when scrolling on slider and loadData is called, we will fetch 3 * getLimit items (one current page, one next page, and one previous page).
+            // firstObs will be the very first obsNum for data rendering this time
+            let firstObs = Math.max(value - o_browse.getLimit(), 1);
+
+            // if value + 1 * getLimit is larger than result counts, we will prefetch more data ahead of firstObs
+            if ((value + o_browse.getLimit() - 1) > opus.resultCount) {
```
rfrenchseti, line 639: I don't understand the purpose of this code. If we aren't going to have enough data to prefetch, what's the purpose of prefetching extra data above the requested obs? We could just prefetch less data and still have the (most of) current and entire previous page available. Also, as written, if there requested page runs even 1 obs over the result count limit, we prefetch an *entire* extra previous page, which seems wasteful.

opus/application/static_media/js/browse.js
```diff
@@ -613,23 +620,61 @@ var o_browse = {
...
+                firstObs = Math.max(firstObs - o_browse.getLimit(), 1);
+            }
+
+            let customizedLimitNum = firstObs === 1 ? value - 1 + 2 * o_browse.getLimit() : 3 * o_browse.getLimit();
```
rfrenchseti, line 643: This code is confusing me. I'm not sure what you're trying to accomplish. At the very least all this work with firstObs deserves a better comment. Better yet I think this could probably be replaced with some kind of min() or max() expression that would make it more clear what you're doing.

opus/application/static_media/js/browse.js
```diff
@@ -1175,18 +1240,18 @@ var o_browse = {
...
-        $(`${selector} .op-dataTable-view`).scrollTop(0);
+    setScrollbarPosition: function(obsNum) {
+        o_browse.setScrollbarOnSlide(obsNum);
     },
 
     // number of images that can be fit in current window size
     getLimit: function() {
-        o_browse.limit = (o_browse.galleryBoundingRect.x !== 0 ? (o_browse.galleryBoundingRect.x * o_browse.galleryBoundingRect.y) : o_browse.limit);
-        return o_browse.limit;
+        // o_browse.limit = (o_browse.galleryBoundingRect.x !== 0 ? (o_browse.galleryBoundingRect.x * o_browse.galleryBoundingRect.y) : o_browse.limit);
```
rfrenchseti, line 1249: These comments can go away

rfrenchseti (changes requested):
- I think the new scroll is working a *lot* better. There is still a problem with the "up" scroll though. I think it's the mirror of the down scroll problem you fixed. Hold down the "Page Down" key to continually scroll down. This works. Then use the "Page Up" key to continually scroll up. The scrollbar will get stuck at the top and not move any more. If you bring it down slightly and then back up, or if you use the scroll wheel, then things start working again.
- Also, the scrollbar is changing size based on how much data has been prefetched. This doesn't make any sense to the user. Can we keep it a constant size? (Same for table)
- Can we put less of a debounce (like none) on the update of the slider when scrolling? I would much rather see the slider update in real-time as the user scrolls up and down in either the gallery or the table. Since moving the slider doesn't actually do anything else, would this cause some kind of problem?
- Go to the table. Drag the slider all the way to the end. Note that the slider value actually stays set to the last available obs number. Now switch to the gallery. The slider will adjust to be the top-left obs. I think there are a couple of problems here. The first is that we require the gallery to have all rows filled. This shouldn't be necessary - it's possible to use the table to scroll down such that startobs is close to the last obs, then when you switch to the gallery that obs should be in the top left corner (approximately) and there should be blank rows at the bottom. The second is that we are allowing the slider to stay at the last obs even though that isn't actually a valid startobs value (it's not on a row boundary).

rfrenchseti:
The current version has been merged into new_ui but the pull request is left open because my comments have not been addressed and Debby still needs to review.

PR #671, opened 2019-05-10
New ui reset buttons s069

juzen2003:
# Update
- Fix reset buttons behavior
- Rename search_widgets to op-search-widgets

### DIFF
--- a/opus/application/static_media/js/opus.js
+++ b/opus/application/static_media/js/opus.js
@@ -103,7 +103,7 @@ var opus = {
             return;
         }
 
-        if (!$.isEmptyObject(selections) || !opus.isWidgetsDefault()) {
+        if (!$.isEmptyObject(selections) || !opus.isDrawnWidgetsListDefault()) {
             $(".op-reset-button button").prop("disabled", false);
         } else if (!opus.isMetadataDefault()) {
             $(".op-reset-button .op-reset-search-metadata").prop("disabled", false);
@@ -378,7 +378,7 @@ var opus = {
     },
 
     // check if current drawn widgets are default ones
-    isWidgetsDefault: function() {
+    isDrawnWidgetsListDefault: function() {
         return o_utils.areObjectsEqual(opus.prefs.widgets, opus.default_widgets);
     },
 
@@ -530,7 +530,7 @@ var opus = {
         $(".op-reset-button button").on("click", function() {
             let targetModal = $(this).data("target");
 
-            if (!$.isEmptyObject(opus.selections) || !opus.isWidgetsDefault()) {
+            if (!$.isEmptyObject(opus.selections) || !opus.isDrawnWidgetsListDefault()) {
                 $(targetModal).modal("show");
             } else if (targetModal === "#op-reset-search-metadata-modal" && !opus.isMetadataDefault()) {
                 $(targetModal).modal("show");
--- a/opus/application/static_media/js/opus.js
+++ b/opus/application/static_media/js/opus.js
@@ -102,18 +102,10 @@ var opus = {
             }
             return;
         }
-        // console.log("=== current selections load ===");
-        // console.log(selections);
-        // console.log("=== is selection empty? ===");
-        // console.log($.isEmptyObject(selections));
-        // console.log("=== is widget not default? ===");
-        // console.log(!opus.areDrawnWidgetsDefault());
-        // console.log("=== condition ===");
-        // console.log(!$.isEmptyObject(selections) || !opus.areDrawnWidgetsDefault());
-        // if (!$.isEmptyObject(opus.selections) || !opus.areDrawnWidgetsDefault() || !opus.checkIfMetadataAreDefault()) {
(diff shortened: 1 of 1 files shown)


opus/application/static_media/js/opus.js
```diff
@@ -336,11 +335,11 @@ var opus = {
         opus.widgets_drawn = [];
         opus.widget_elements_drawn = [];
 
-        if (resetMetadata && !opus.checkIfMetadataAreDefault()) {
+        if (resetMetadata && !o_utils.areObjectsEqual(opus.prefs.cols, default_columns.split(','))) {
```
rfrenchseti: If I'm reading this code right, it seems cleaner to take the areObjectsEqual out first:
```
if (!areObjectsEqual...) {
   if (resetMetadata) {
      do thing 1
   } else {
      do thing 2
   }
}

File: opus/application/static_media/js/opus.js [hunk 2]
```diff
@@ -350,9 +349,9 @@ var opus = {
         o_menu.markDefaultMenuItem();
 
         let deferredArr = [];
-        $.each(opus.default_widgets, function(index, slug) {
+        $.each(opus.default_widgets.slice().reverse(), function(index, slug) {
             deferredArr.push($.Deferred());
-            o_widgets.getWidget(slug,"#search_widgets",deferredArr[index]);
+            o_widgets.getWidget(slug,"#op-search-widgets",deferredArr[index]);
```
rfrenchseti: Add space after commas

opus/application/static_media/js/opus.js
```diff
@@ -336,11 +335,11 @@ var opus = {
         opus.widgets_drawn = [];
```
rfrenchseti, line 338: I'm confused by this line. Why are we setting widgets to [] instead of to the default_widgets?

rfrenchseti (changes requested):

opus/application/static_media/js/opus.js
juzen2003: Here is what we do in startOver:
1. We empty the widget area first (opus.js line 323). 
2. Later on when we are looping through opus.default_widgets and calling o_widgets.getWidget for each item in opus.default_widgets (opus.js line 354), we will update opus.widgets_drawn in o_widgets.getWidget (widgets.js line 559). 

So If we set opus.widgets_drawn to opus.default_widgets here, in o_widgets.getWidget, it will think those widgets are already drawn and return from the function (widgets.js line 374). That way those widgets won't be drawn, and widget area will still be empty.

opus/application/static_media/js/opus.js
```diff
@@ -321,7 +320,7 @@ var opus = {
         // the application default widgets
 
         clearInterval(opus.main_timer);  // stop polling for UI changes for a moment
-        $("#search_widgets").empty(); // remove all widgets on the screen
+        $("#op-search-widgets").empty(); // remove all widgets on the screen
```
rfrenchseti: Debby and I had a talk about empty() awhile back. My undertstanding is that "empty" detaches the children and their contents from the DOM, but still leaves the elements in the browser, essentially wasting memory. They never go away. Instead we should be using remove(), which both detaches and deletes the elements from memory. Debby do you agree?

rfrenchseti (changes requested):

opus/application/static_media/js/opus.js
dstopp: yes - see http://mundrisoft.com/tech-bytes/jquery-difference-between-remove-empty-and-detach-methods/

opus/application/static_media/js/opus.js
```diff
@@ -103,10 +103,9 @@ var opus = {
             return;
         }
 
-        // if (!$.isEmptyObject(opus.selections) || !opus.checkIfDrawnWidgetsAreDefault() || !opus.checkIfMetadataAreDefault()) {
-        if ($.isEmptyObject(selections) || !opus.checkIfDrawnWidgetsAreDefault()) {
+        if (!$.isEmptyObject(selections) || !o_utils.areObjectsEqual(opus.prefs.widgets, opus.default_widgets)) {
             $(".op-reset-button button").prop("disabled", false);
-        } else  if (!opus.checkIfMetadataAreDefault()) {
+        } else if (!o_utils.areObjectsEqual(opus.prefs.cols, default_columns.split(','))) {
```
dstopp: i think since you are checking for default metadata columns multiple times, keeping the function but updating may be better plan.  Same comment for drawnWidgets.

dstopp (changes requested):

opus/application/static_media/js/opus.js
juzen2003: Thanks for the suggestion, the code is updated to use .remove() instead of .empty().

opus/application/static_media/js/opus.js
juzen2003: Put back areDrawnWidgetsDefault and areSelectedMetadataDefault functions for readability.

dstopp (approved):

rfrenchseti (approved):

PR #672, opened 2019-05-10
allow the indicator to change pages as the user moves thru the metada…

dstopp:
b004 - follow the metadata modal in the gallery and table view, scrolling the page as needed.


opus/application/static_media/js/utils.js
```diff
@@ -32,3 +32,28 @@ var o_utils = {
...
 };
+
+
+
+/**
+ * returns true if an element is visible, with decent performance
+ * @param [scope] scope of the render-window instance; 
+ * @returns {boolean}
+ */
+$.fn.isOnScreen = function(scope){
```
rfrenchseti: Add space before {
Is there a reason to assign this to element when you could just use this throughout?

opus/application/static_media/js/utils.js
```diff
@@ -32,3 +32,28 @@ var o_utils = {
...
+    if (target.is(':visible') === false) {
+        return false;
+    }
+    scope = $(scope);
+    let top = scope.offset().top;
+    let bottom = top + scope.height();
+    let elementTop = target.offset().top;
+    let elementBottom = elementTop + target.outerHeight();
+
+    return ((elementBottom <= bottom) && (elementTop >= top));
```
rfrenchseti: Don't actually need any of these parens.

rfrenchseti (commented):
There are several ways in which this doesn't work:
- Put the binoculars in the bottom row. Resize the browser vertically so that it is aligned about 1/3 of the way down from the top of the thumbnail with the binoculars. Now hit left/right/left/right and watch the screen jump around.
- Scroll to any location. Put the binoculars somewhere. Use the left arrow to go to the top left obs. You can't go any further and the left finger icon is disabled.

rfrenchseti (changes requested):
x

opus/application/static_media/js/utils.js
dstopp: I used code from another source and thought it interesting that they did it that way - I hadn't gotten to 'personalizing' the code other than seeing how it worked for me.

opus/application/static_media/js/utils.js
dstopp: i left the parens in from the original code for clarity.  certainly they are not needed but i thought it was easier to read.

rfrenchseti:
- While the modal is up, it's possible to scroll the background using the scrollbar, wheel, or page up/down buttons. Either way, you can easily get to a place where the cached data for the modal has gone away. Since we already close the modal on slider drag, perhaps we should do the same on scroll events? That would save a lot of trouble.
- I don't see any improvement on browser resize. If I do the exact thing I reported earlier, I get the same odd behavior of the scroll jumping back and forth as I move left and right.

opus/application/apps/ui/templates/ui/base.html
```diff
@@ -92,7 +92,7 @@
                 </div>
             </div><!-- /search -->
 
-            <div id="browse" class="tab-pane fade" role="tabpanel" aria-labelledby="browse-tab">
+            <div id="browse" class="tab-pane fade op-no-select" role="tabpanel" aria-labelledby="browse-tab">
```
rfrenchseti, line 95: There is weird behavior with the op-no-select.
1) You can select the text in the metadata modal, but if you do a Ctrl+A to select it all, you also get the text in the top nav bar, plus a whole bunch of blank lines. You can also select the top nav bar text separately with mouse-drag.
2) You can drag-select the text in the left column of the Select Metadata dialog.
3) You can select stuff in the right column of the detail page
4) You can select the main footer, which also shows up with Ctrl+A

opus/application/static_media/js/browse.js
```diff
@@ -471,6 +471,7 @@ var o_browse = {
             },
             stop: function(event, ui) {
                 o_browse.onUpdateSlider(ui.value);
+                $("#galleryView").modal('hide');
```
rfrenchseti: Now that you're killing the modal when the cache empties, I'm not sure this is necessary. As long as the slider is just repositioning the scrollbar with the existing cached data, it's no different than just moving the main scrollbar. We only need to kill the modal when the cache for the viewed observation is deleted.

opus/application/static_media/js/browse.js
```diff
@@ -572,15 +573,25 @@ var o_browse = {
         let contentsView = o_browse.getScrollContainerClass();
         let maxObs = (opus.prefs.view === "cart" ? o_cart.cartCount : opus.resultCount);
 
-        let obsNum = $(`${tab} .thumbnail-container[data-id=${opusId}]`).data("obs") + 1;
+        let obsNum = $(`${tab} .thumbnail-container[data-id=${opusId}]`).data("obs");
         if (obsNum <= maxObs) {
-            let nextElem = $(`${tab} .thumbnail-container[data-obs=${obsNum}]`);
+            let nextElem = (contentsView === ".op-gallery-view" ? $(`${tab} .thumbnail-container[data-obs=${obsNum}]`) : $(`${tab} tr[data-obs=${obsNum}]`));
```
rfrenchseti: I thought you told Dave you didn't want to compare against the scroll class name? Shouldn't you use something like isGalleryView?

opus/application/static_media/js/browse.js
```diff
@@ -572,15 +573,25 @@ var o_browse = {
...
                 $("#galleryViewContents").addClass("op-disabled");
                 opus.prefs[startObsLabel] = obsNum;
                 $(`${tab} ${contentsView}`).infiniteScroll("loadNextPage");
+            } else {
+                // make sure the current element that the modal is displaying is viewable; if not, scroll til it is
+                let scope = $(`${tab} ${contentsView}`);
+                if (!nextElem.isOnScreen($(`${tab} .gallery-contents`))) {
+                    let bottom = scope.offset().top + scope.height();
+                    let elemHeight = nextElem.height();
+                    let offset = Math.max(nextElem.offset().top + elemHeight - bottom, elemHeight);
```
rfrenchseti: This math only works if the elem is off the bottom of the screen, not off the top.

opus/application/static_media/js/browse.js
```diff
@@ -595,9 +606,9 @@ var o_browse = {
         let contentsView = o_browse.getScrollContainerClass();
         o_browse.currentOpusId = opusId;
         // decrement obsNum to see if there is a previous one to retrieve
-        let obsNum = $(`${tab} .thumbnail-container[data-id=${opusId}]`).data("obs") - 1;
+        let obsNum = $(`${tab} .thumbnail-container[data-id=${opusId}]`).data("obs");
         if (obsNum > 0) {
-            let prevElem = $(`${tab} .thumbnail-container[data-obs=${obsNum}]`);
+            let prevElem = (contentsView === ".op-gallery-view" ? $(`${tab} .thumbnail-container[data-obs=${obsNum}]`) : $(`${tab} tr[data-obs=${obsNum}]`));
```
rfrenchseti: Same comment about isGalleryView.

opus/application/static_media/js/browse.js
```diff
@@ -606,6 +617,13 @@ var o_browse = {
                 let startObs = obsNum - o_browse.getLimit();
                 opus.prefs[startObsLabel] = (startObs > 0 ? startObs : 1);
                 $(`${tab} ${contentsView}`).infiniteScroll("loadNextPage");
+            } else {
+                // make sure the current element that the modal is displaying is viewable; if not, scroll til it is
+                let scope = $(`${tab} ${contentsView}`);
+                if (!prevElem.isOnScreen($(`${tab} .gallery-contents`))) {
+                    let newPosition = scope.scrollTop() - prevElem.height();
```
rfrenchseti: I think this math only works if the elem is off the top, not off the bottom.

opus/application/static_media/js/browse.js
```diff
@@ -1416,22 +1432,30 @@ var o_browse = {
         // don't delete the metadata if the observation is in the cart
         if (!galleryObsElem.eq(index).hasClass("in")) {
             let delOpusId = galleryObsElem.eq(index).data("id");
+            if ($("#galleryView").hasClass("show")) {
```
rfrenchseti, line 1408: Should rename these classes at some point to op-gallery-view and op-gallery-view-contents

opus/application/static_media/js/browse.js
```diff
@@ -1416,22 +1432,30 @@ var o_browse = {
...
+            }
             delete o_browse.galleryData[delOpusId];
         }
         galleryObsElem.eq(index).remove();
         tableObsElem.eq(index).remove();
     },
 
+    isGalleryView: function(view) {
+        let browse = o_browse.getBrowseView(view);
+        return opus.prefs[browse] === "gallery" ;
```
rfrenchseti: Extra space before ;

opus/application/static_media/js/browse.js
```diff
@@ -1801,6 +1831,9 @@ var o_browse = {
         // list columns + values
         let html = "<dl>";
         $.each(opus.colLabels, function(index, columnLabel) {
+            if (opusId === "" || o_browse.galleryData[opusId] === undefined || o_browse.galleryData[opusId][index] === undefined) {
+                console.log("uh oh");
```
rfrenchseti: Ummmmm....

Nothing wrong with doing a sanity change, but go ahead and use the new opus.logError function and make the description meaningful

rfrenchseti (changes requested):

opus/application/static_media/js/browse.js
dstopp: oops, lol.  that was just for me to debug... I meant to remove it.  Ah well...

opus/application/apps/ui/templates/ui/base.html
dstopp: Will address the remaining no-select issues in a separate pull-request.  These issues are unrelated to the original code set, more or less...

opus/application/static_media/js/browse.js
dstopp: yes - missed this, good catch

opus/application/static_media/js/browse.js
dstopp: we discussed and decided to leave the rename for a different pull request.

opus/application/apps/ui/templates/ui/base.html
```diff
@@ -253,14 +253,14 @@ <h3 class='op-header-text'></h3>
 
         <!-- modal dialog for the gallery views -->
         <div id="galleryView" class="modal fade" role="dialog" tabbindex="-1" data-backdrop="false" data-keyboard="false" aria-labelledby="galleryViewLabel" aria-hidden="true">
-            <div class="modal-dialog modal-lg" role="document">
+            <div class="modal-dialog modal-dialog-centered modal-lg" role="document">
```
rfrenchseti, line 256: I thought we wanted to take this out for now

opus/application/static_media/js/browse.js
```diff
@@ -174,7 +174,7 @@ var o_browse = {
                     o_cart.toggleInCart(fromOpusId, opusId);
                 }
             } else {
-                o_browse.showModal(opusId);
+                o_browse.showGalleryViewModal(opusId);
```
rfrenchseti: Why is this called GalleryViewModal? Doesn't it also work in table view? Perhaps call it "BrowseModal" (except it works in cart...) or "MetadataDetailModal"?

opus/application/static_media/js/browse.js
```diff
@@ -561,51 +557,45 @@ var o_browse = {
...
-    loadNextPageIfNeeded: function(opusId) {
+    loadPageIfNeeded: function(direction, opusId) {
         // opusId will be empty at the end of the observations, so just return out.
         if (opusId === "") {
             return;
         }
         let startObsLabel = o_browse.getStartObsLabel();
         let tab = opus.getViewTab();
         let contentsView = o_browse.getScrollContainerClass();
+        o_browse.currentOpusId = opusId;
```
rfrenchseti: Also wondering about this name. I assume this is where the binoculars are pointed? That's not obvious from the name. current opus id is pretty generic. How about something like "metadataDetailOpusId"?

opus/application/static_media/js/browse.js
```diff
@@ -561,51 +557,45 @@ var o_browse = {
...
                 // disable keydown on modal when it's loading
                 // this will make sure we have correct html elements displayed for prev observation
                 $("#galleryViewContents").addClass("op-disabled");
-                let startObs = obsNum - o_browse.getLimit();
-                opus.prefs[startObsLabel] = (startObs > 0 ? startObs : 1);
                 $(`${tab} ${contentsView}`).infiniteScroll("loadNextPage");
+
+                if (direction === "prev") {
+                    let startObs = obsNum - o_browse.getLimit();
+                    opus.prefs[startObsLabel] = (startObs > 0 ? startObs : 1);
```
rfrenchseti: or Math.max(startObs, 1)

opus/application/static_media/js/browse.js
```diff
@@ -561,51 +557,45 @@ var o_browse = {
...
-        let tab = opus.getViewTab();
-        let contentsView = o_browse.getScrollContainerClass();
-        o_browse.currentOpusId = opusId;
-        // decrement obsNum to see if there is a previous one to retrieve
-        let obsNum = $(`${tab} .thumbnail-container[data-id=${opusId}]`).data("obs") - 1;
-        if (obsNum > 0) {
-            let prevElem = $(`${tab} .thumbnail-container[data-obs=${obsNum}]`);
-            // if it's not there go retrieve it...
-            if (prevElem.length === 0) {
+        if (checkView) {
```
rfrenchseti, line 579: Maybe I'm just too tired, but I find this very confusing. We just got obsNum by looking it up in the thumbnail-container by opusid. So we know it's there, right? Now we're immediately looking it up by obsNum and checking to see whether it's there or not. But it has to be there...  If it's not there, we do this weird thing where we set the startobs slug back an entire browser page. Why would we ever do that? If I'm right about the first point, this code is never called anyway.

rfrenchseti (changes requested):

opus/application/static_media/js/browse.js
```diff
@@ -565,37 +565,23 @@ var o_browse = {
...
-                    let newStartObs = startObs + Math.floor((obsNum - startObs) / rowCount) * rowCount;
-                    let jsElement = $(`${tab} .thumbnail-container[data-id=${opusId}]`)[0];
-                    jsElement.scrollIntoView();
-                    o_browse.onUpdateSlider(newStartObs);
-                }
+            // make sure the current element that the modal is displaying is viewable
+            if (!element.isOnScreen($(`${tab} .gallery-contents`))) {
+                let startObs = $(`${tab} ${contentsView}`).data("infiniteScroll").options.obsNum;
+                let rowCount = opus.getViewNamespace().galleryBoundingRect.x;
+                let newStartObs = startObs + Math.floor((obsNum - startObs) / rowCount) * rowCount;
```
rfrenchseti: This still doesn't work properly for binocs off the bottom of the screen. It jumps the scroll an entire page so that the binocs end up in the first row. This happens even if you just hit right arrow. I'm assuming we just want the gallery to scroll by a single row in this case. I really think you need to treat the obsNum < startObs and obsNum > startObs cases differently. You can't just to one formula that covers both.

Also, thanks to the joys of floating point, you can't rely on floor(x/y) to do what you want. You need flloat(x/y + 0.0000001) or some other very small number.

rfrenchseti: Also there's something very weird with this in the table view. As you go backwards, you have to go one observation past the topmost one you can see before it triggers a scroll. Also, the amount of scroll when running off the top and when running off the bottom are totally different.

rfrenchseti (changes requested):
The metadata dialog is disappearing when the slider moves, even if the slider didn't cause a cache flush. I thought we had decided that it only needed to disappear on cache flush, not just when the slider was moved at all.

opus/application/static_media/js/browse.js
rfrenchseti: This is not resolved.

opus/application/apps/ui/templates/ui/base.html
dstopp: my changes include a fix for this issue.

rfrenchseti:
- The problems with the table view are not fixed. Using right/left arrows to move off the screen does weird things in both directions. Going up seems to scroll two lines at a time. Going down moves the table such that the highlighted obs isn't on the screen.
- (Unrelated to binocs, but still a problem) When you move the metadata modal near the bottom of the screen, a browser scrollbar shows up on the right. It's as if the dialog is taller than it appears. This was introduced by your change to allow the modal to be centered.

opus/application/static_media/js/browse.js
```diff
@@ -789,6 +769,11 @@ var o_browse = {
 
     },
 
+    hideGalleyViewModal: function() {
```
rfrenchseti: Spelled gallery wrong here and various other places in the code - do a search

rfrenchseti (approved):

PR #673, opened 2019-05-10
New ui startobs infiniteScroll up/down direction (B034), maintain scrollbar position (A048), continuation of #668

juzen2003:
# Note: this pull request is the continuation of #668 
For record, this is the latest comments that requires updates on codes:
https://github.com/SETI/pds-opus/pull/668#pullrequestreview-235818458


rfrenchseti:
If I drag the slider to some random position, and then click on "View Table" and "View Gallery" repeatedly, the slider moves each time. This seems to happen only with a really large browser. (Try "zooming out" in the browser to get a larger screen area if necessary).

PR #675, opened 2019-05-11
New ui s026 (widget area width)

juzen2003:
# Update
- Widget area now can shrink to a minimum width: 400px when browse window is narrower.
- Rename widget_column to op-widget-column.

### DIFF
--- a/opus/application/apps/ui/templates/base.html
+++ b/opus/application/apps/ui/templates/base.html
@@ -68,7 +68,7 @@
         {% block content %}
         <div class="tab-content container-fluid tab-full-width">
             <div id="search" class="tab-pane fade" role="tabpanel" aria-labelledby="search-tab">
-                <div class="row op-search-row">
+                <div class="row flex-nowrap">
                     <div class="col pr-0 sidebar-area border-right">
                         <div id="sidebar-container" class = "sidebar_wrapper ps-scrollbar">
                             <div id = "sidebar">
--- a/opus/application/apps/ui/templates/base.html
+++ b/opus/application/apps/ui/templates/base.html
@@ -84,7 +84,7 @@
 
                     <!-- selected widgets -->
                     <div class="col">
-                        <div id="widget-container" class = "widget_column">
+                        <div id="widget-container" class="op-widget-column">
                             <!-- note to self:    this is where the widgets will be replace by cards; should no longer by ul/li -->
                             <ul id="op-search-widgets"></ul>
                         </div>
--- a/opus/application/apps/ui/templates/base.html
+++ b/opus/application/apps/ui/templates/base.html
@@ -68,7 +68,7 @@
         {% block content %}
         <div class="tab-content container-fluid tab-full-width">
             <div id="search" class="tab-pane fade" role="tabpanel" aria-labelledby="search-tab">
-                <div class="row">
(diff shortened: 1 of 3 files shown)


opus/application/apps/ui/templates/base.html
```diff
@@ -68,7 +68,7 @@
         {% block content %}
         <div class="tab-content container-fluid tab-full-width">
             <div id="search" class="tab-pane fade" role="tabpanel" aria-labelledby="search-tab">
-                <div class="row">
+                <div class="row op-search-row">
```
dstopp: use bootstrap 4 class 'flex-nowrap' instead of opus custom class.

dstopp (changes requested):

opus/application/static_media/css/opus.css
```diff
@@ -893,21 +893,27 @@ nav.panel-heading {
...
 }
 
+/* This will prevent search widget area from being moved to the bottom of search sidebar */
+.op-search-row {
+    flex-wrap: nowrap;
+}
+
 /* This is to make sortable scrolling (when sorting happened) works with ps */
 #widget-container {
     overflow: auto !important;
```
dstopp: these overflow appear to be overlapping...

opus/application/static_media/js/search.js
```diff
@@ -21,7 +21,7 @@ var o_search = {
         minScrollbarLength: opus.minimumPSLength
     }),
     widgetScrollbar: new PerfectScrollbar("#widget-container" , {
-        suppressScrollX: true,
+        // suppressScrollX: true,
```
dstopp: do we still need this commented out code?

opus/application/static_media/js/search.js
```diff
@@ -443,15 +441,16 @@ var o_search = {
     },
 
     adjustSearchWidgetHeight: function() {
-        let containerHeight = $("#search").height() - 80;
+        let footerHeight = $(".app-footer").outerHeight();
+        let mainNavHeight = $("#op-main-nav").outerHeight();
+        let totalNonSearchAreaHeight = footerHeight + mainNavHeight;
+        let containerHeight = $("#search").height() - totalNonSearchAreaHeight;
```
dstopp, line 447: do you need to use outerHeight here as well?  <i can't tell; i suspect you are ok but wanted to check>

dstopp (changes requested):

opus/application/static_media/js/search.js
juzen2003: If we use .outerHeight() here, the scrollbar will be able to scroll all the down below the footer (a little bit covered by the footer), so we use .height() here and scrollbar looks fine.

opus/application/static_media/js/search.js
juzen2003: I put it back for now, but it shouldn't affect anything because we only allow sortable widgets to move vertically (disable horizontal movement). I was trying get the x-scrolling working for widget area only, but couldn't get it to work on Friday. When screen is narrow and we allow the whole body x-scrolling, it will look pretty ugly (plus we need another ps instance for html tag), so I was trying get the x-scrolling only for widget area in this case. I will come back to work on that later after infiniteScroll.

dstopp (approved):

PR #676, opened 2019-05-11
Fix cart

fyellin:


log_analyzer/opus/session_info.py
```diff
@@ -239,7 +239,13 @@ def __download_cart_metadata_csv(self, _query: Dict[str, str], _match: Match[str
     def __create_archive(self, query: Dict[str, str], _match: Match[str]) -> SESSION_INFO:
         self.performed_download()
         has_url = query.get('urlonly') not in [None, '0']
-        text = f'Download {"URL" if has_url else "Data"} Archive for Cart'
+
+        if self._previous_product_info_type is not None:
```
rfrenchseti: I don't understand why you are using _previous_product_info_type here. The "create archive" URLs have their own "types=XXX,YYY" field, which is what's important here, not what happened to be given for some previous status or view API call.

rfrenchseti (changes requested):

fyellin:
Because I was confused by what you wanted.  Fixed in five minutes.

rfrenchseti (approved):

PR #677, opened 2019-05-13
C035 - Add support for cart PS and infinite scroll; add table view to cart

dstopp:
c035 - add support for PS and infinite scroll to cart; added the table view whilst i was at it.


rfrenchseti:
1) (FIXED) Clicking on a table row doesn't bring up the metadata modal
2) (FIXED) Unchecking an obs in the table view doesn't actually seem to do anything
3) (FIXED) Shift+clicking on a table row doesn't start add range
4) (FIXED) If you click on the Cart tab while already in the cart view, you get a blank gallery
5) Now that we have "View Gallery/Table" and then "Select Metadata" in the nav bar, it has a very similar feel to the browse page. *Except* those items are on the right side of a divider, and the whole section is on the right side of the screen. What do you think about:
    - A) Swapping the two sides (View Table / Select Metadata / Empty Cart) and (Metadata CSV / Data Archive / URL archive) and/or
    - B) Swapping the product pane and the gallery pane. If we did this, then switching between browse and cart would seem almost seamless. The nav bar wouldn't change all that much. In fact we could rename Download CSV (all results) to Download CSV (cart) and it would also look almost the same. You could even keep the slider where it is (since we do want a slider in the cart)
6) (FIXED) There's something really broken with infinite scroll on the normal browse page. As you scroll down it keeps reloading the same page over and over so you get duplicate data, and the slider moves right and then jumps left.
7) The cart gallery runs off the bottom of the screen, so when it's full the final row of thumbnails is partially obscured.

rfrenchseti:
- (FIXED) I had to change document.ready to say:
`    opus.namespaceDict = { "search": o_browse, "cart": o_cart, "browse": o_browse };`
I'm sure this isn't the right fix, but at least it let me get things going.
- (FIXED) When you first switch to the browse page, the slider just spins forever (is fixed by one of my code review comments)
- (FIXED) If you drag the slider on the browse page, it does update and the page fetches, but the size of the slider "button" is wrong - it doesn't expand to support the width of the observation number you've set (is fixed by one of my code review comments)
- (FIXED) When you scroll down in either gallery or table view on the browse page, there is no big spinner while it fetches
- When you switch from gallery to table and back on the browse page, the slider moves (previously reported on Dave's pull #668)
- (FIXED) The first time you switch to the Cart tab, it says there are no entries in the cart. Then if you switch to it again it displays thumbnails. Although if you click it yet again the thumbnails go away and you can't get them back.
- (FIXED) Clicking on a table header (browse or cart) jumps you back to the search tab
- (FIXED) Infinite scroll on the browse tab is still completely broken - as you scroll down you keep getting copies of the same data
- (FIXED) Infinite scroll on the cart gallery doesn't do anything at all. Well, sometimes. It seems kind of random.

opus/application/static_media/js/browse.js
```diff
@@ -663,13 +660,13 @@ var o_browse = {
     updateSliderHandle: function() {
         let tab = `#${opus.prefs.view}`;
         let contentsView = o_browse.getScrollContainerClass();
-        let selector = (opus.prefs.browse === "dataTable") ? `#${opus.prefs.view} #dataTable tbody tr` : `#${opus.prefs.view} .gallery .thumbnail-container`;
+        let selector = (opus.prefs[browse] === "gallery") ?  `#${opus.prefs.view} .gallery .thumbnail-container` : `#${opus.prefs.view} .op-data-table tbody tr`;
```
rfrenchseti: I don't see how this can work. Isn't `browse` an undefined variable here? Maybe this is why the slider isn't working some of the time.

opus/application/apps/ui/templates/base.html
```diff
@@ -101,7 +101,7 @@
                                     <a href="#" class="op-browse-view nav-link"><i class="far fa-list-alt"></i></a>
                                 </li>
                                 <li class="nav-item">
-                                    <a href="#" class="metadataModal nav-link" data-toggle="modal" data-target="#metadataSelector" title="Select metadata to display or include in CSV files for each observation"><i class="fas fa-columns"></i>&nbsp;Select Metadata</a>
+                                    <a href="#" class="op-metadataModal nav-link" data-toggle="modal" data-target="#metadataSelector" title="Select metadata to display or include in CSV files for each observation"><i class="fas fa-columns"></i>&nbsp;Select Metadata</a>
```
rfrenchseti: I think we should be calling this op-metadata-modal.

opus/application/static_media/js/browse.js
```diff
@@ -1342,7 +1339,8 @@ var o_browse = {
 
     // return the infiniteScroll container class for either gallery or table view
     getScrollContainerClass: function() {
-        return (opus.prefs.browse === "gallery" ? ".op-gallery-view" : ".op-dataTable-view");
+        let browse = (opus.prefs.view === "cart_" ? "cart_browse" : "browse");
```
rfrenchseti: This should not have a "_" after cart

opus/application/static_media/js/browse.js
```diff
@@ -978,28 +976,33 @@ var o_browse = {
 
     updateBrowseNav: function() {
         let tab = `#${opus.prefs.view}`;
+        let browse = (opus.prefs.view === "browse" ? "browse" : "cart_browse");
```
rfrenchseti: This `let browse =` is duplicated many times. We should have a function for it.

opus/application/static_media/js/browse.js
```diff
@@ -978,28 +976,33 @@ var o_browse = {
...
-            $(".op-dataTable-view", "#browse").hide();
-            $(".op-gallery-view", "#browse").fadeIn();
 
-            $(".op-browse-view", "#browse").html("<i class='far fa-list-alt'></i>&nbsp;View Table");
-            $(".op-browse-view", "#browse").attr("title", "View sortable metadata table");
-            $(".op-browse-view", "#browse").data("view", "dataTable");
+        let suppressScrollY = false;
+
+        if (opus.prefs[browse] == "gallery") {
+            $(".op-dataTable-view", tab).hide();
```
rfrenchseti: Shouldn't this be `op-data-table-view` to match `op-data-table`?

opus/application/static_media/js/browse.js
```diff
@@ -978,28 +976,33 @@ var o_browse = {
 
     updateBrowseNav: function() {
         let tab = `#${opus.prefs.view}`;
```
rfrenchseti: I thought we were going to abstract this away so that opus.prefs.view was not used directly as a selector.

opus/application/static_media/js/opus.js
```diff
@@ -589,6 +596,10 @@ var opus = {
  *
  */
 $(document).ready(function() {
+    // just for browse and cart, as we need to abstract some functions easier
+    // needs to be done in the document.ready because o_cart & o_browse need to be defined first
+    opus.namespaceDict = { "cart": o_cart, "browse": o_browse };
```
rfrenchseti: If this maps "view" to namespace, I think it should have a name that reflects that. Like viewToNamespaceDict.

rfrenchseti (changes requested):

rfrenchseti:
The following bugs from earlier are now FIXED:

1) Clicking on a table row doesn't bring up the metadata modal
2) Unchecking an obs in the table view doesn't actually seem to do anything
3) Shift+clicking on a table row doesn't start add range

juzen2003:
mutationObserver.js looks fine. cartObserver will detect changes when we: 
- Switch from other tabs to cart tab 
- In gallery, click add/remove icon
- In table, click checkbox, remove/add tab in hamburger
- Switch between gallery and table
- adjustProductInfoHeight is called in cartObserver callback function to update both perfect scrollbar.
The only optimization I can see is to update the callback with: (which I didn't do it in original file)              
`let lastMutationIdx = mutationsList.length - 1;
mutationsList.forEach((mutation, idx) => {
    if (idx === lastMutationIdx) {
        adjustProductInfoHeight();
    }
});`
But overall it will work fine.

opus/application/static_media/js/browse.js
```diff
@@ -1558,26 +1563,43 @@ var o_browse = {
 
     countGalleryImages: function() {
         let tab = opus.getViewTab();
```
rfrenchseti: countGalleryImages needs to take a tab argument like everything else.

opus/application/static_media/js/browse.js
```diff
@@ -1356,14 +1359,16 @@ var o_browse = {
...
     },
 
     // Instantiate infiniteScroll
-    initInfiniteScroll: function(selector) {
-        let tab = opus.getViewTab();
-        let startObsLabel = o_browse.getStartObsLabel();
+    initInfiniteScroll: function(view, selector) {
+        let tab = `#${view}`;
+        let startObsLabel = o_browse.getStartObsLabel(view);
+        opus.getViewNamespace(view).galleryBoundingRect = o_browse.countGalleryImages();
```
rfrenchseti: Note that countGalleryImages uses the "current" opus.prefs.view tab, while the rest of this function uses the view that was passed in. countGalleryImages needs to take a tab argument as well.

opus/application/static_media/js/browse.js
```diff
@@ -1298,6 +1298,9 @@ var o_browse = {
 
         // need to add limit - getting twice as much so that the prefetch is done in one get instead of two.
         let limitNum = customizedLimitNum === undefined ? o_browse.getLimit() * 2 : customizedLimitNum;
+        if (limitNum === 0) {
```
rfrenchseti: Remove debugging code eventually.

opus/application/static_media/js/browse.js
```diff
@@ -1298,6 +1298,9 @@ var o_browse = {
```
rfrenchseti, line 1334: updateStartobsInUrl is using the current opus.prefs.view. It needs to use the passed-in tab.

opus/application/static_media/js/browse.js
```diff
@@ -1298,6 +1298,9 @@ var o_browse = {
 
         // need to add limit - getting twice as much so that the prefetch is done in one get instead of two.
         let limitNum = customizedLimitNum === undefined ? o_browse.getLimit() * 2 : customizedLimitNum;
```
rfrenchseti: getLimit() needs to take the tab as an argument

opus/application/static_media/js/browse.js
rfrenchseti: getLimit() below throughout the path function (many instances) needs to take a tab argument

rfrenchseti: also getViewNamespace needs to take a tab argument

opus/application/static_media/js/browse.js
```diff
@@ -1558,26 +1563,43 @@ var o_browse = {
...
-            xCount = Math.floor($(`${tab} .gallery-contents`).width()/o_browse.imageSize);   // images are 100px
-            // yCount = Math.ceil(($(`${tab} .gallery-contents`).height() - $(".app-footer").height()) /o_browse.imageSize);   // images are 100px
-            yCount = Math.ceil($(`${tab} .gallery-contents`).height()/o_browse.imageSize);   // images are 100px
-        } else {
-            console.log("countGalleryImages: This should never happen.");
-        }
+        let width = o_browse.calculateGalleryWidth();
+        let height = o_browse.calculateGalleryHeight();
+
+        let xCount = Math.floor(width/o_browse.imageSize);   // images are 100px
```
rfrenchseti: Probably want to kill the comment giving a hard value for the image size

opus/application/static_media/js/browse.js
```diff
@@ -1558,26 +1563,43 @@ var o_browse = {
...
         let tab = opus.getViewTab();
         let footerHeight = $(".app-footer").outerHeight();
         let mainNavHeight = $("#op-main-nav").outerHeight();
         let navbarHeight = $(".panel-heading").outerHeight();
         let totalNonGalleryHeight = footerHeight + mainNavHeight + navbarHeight;
-        let containerHeight = $(window).height()-totalNonGalleryHeight;
+        return  $(window).height()-totalNonGalleryHeight;
+    },
+
+    calculateGalleryWidth: function() {
```
rfrenchseti: calculateGalleryWidth and Height need to take tab arguments

rfrenchseti (changes requested):

opus/application/static_media/js/browse.js
```diff
@@ -672,7 +682,10 @@ var o_browse = {
         $(selector).each(function(index, elem) {
             // Fot gallery view, the topBoxBoundary is the top of .gallery-contents
             // For table view, we will set the topBoxBoundary to be the bottom of thead (account for height of thead)
-            topBoxBoundary = topBoxBoundary || (opus.prefs[browse]  === "gallery") ? $(`${tab} .gallery-contents`).offset().top : $(`${tab} .gallery-contents`).offset().top + $(`${tab} .op-data-table thead th`).outerHeight();
+            topBoxBoundary = topBoxBoundary ||
```
rfrenchseti: Add extra level of parens around expression

opus/application/static_media/js/browse.js
```diff
@@ -36,7 +36,8 @@ var o_browse = {
     galleryBegun: false, // have we started the gallery view
     galleryData: {},  // holds gallery column data
     imageSize: 100,     // default
-    maxCachedObservations: 1000,    // max number of obserations to store in cache; at some point, we can probably figure this out dynamically
+    maxCachedObservations: 750,    // max number of obserations to store in cache;
```
rfrenchseti: obserations => observations

opus/application/static_media/js/browse.js
```diff
@@ -1033,6 +1051,17 @@ var o_browse = {
         return url;
     },
 
+    updateViewInUrl: function(view, url) {
+        // remove any existing view= slug or startobs= slug
```
rfrenchseti: Remove or startobs= slug

rfrenchseti (changes requested):

rfrenchseti (approved):

PR #679, opened 2019-05-13
Fake API calls for log_analyzer

### LINKED ISSUE #620: Need to add fake API calls to better track what the user does
With the new UI, we lose some information in the Apache logs about what the user does. This makes the output of the log_analyzer less useful. We should provide fake API calls to the backend that do nothing except feed the log_analyzer. This will allow us to record any data we want on what the user is doing.

rfrenchseti:
- Prevent caching of most API calls so we get Apache log entries
- Add fake API calls when switching gallery/table, viewing metadata on an obs, and popping up metadata selector
- Update log_analyzer to better handle old log files and to clean up browse/cart gallery/data handling
- This fixes #620

### DIFF
--- a/log_analyzer/opus/query_handler.py
+++ b/log_analyzer/opus/query_handler.py
@@ -48,6 +48,7 @@ class QueryHandler:
         self._previous_sort_order = self.DEFAULT_SORT_ORDER
         self._previous_pages = ['', '']
         self._previous_startobss = ['', '']
+        self._previous_browses = ['', '']
         self._previous_state = State.RESET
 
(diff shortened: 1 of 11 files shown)


fyellin (approved):

dstopp (changes requested):
instead of having the fake url call in a lot of places, would be better to catch the 'show' event and put the fake url in there, ie.,
$("#myModalSelector").on("show.bs.modal", function(e) {
    doFakeUrl();
}

Note that there is already one for #metadataSelector that you can just add code to.

dstopp (approved):

PR #682, opened 2019-05-16
First pass JS cleanup - through opus.js

rfrenchseti:
Note this branch is based off Debby's c035_cart branch to reduce future merge problems.

- Moved all templates in apps/ui to apps/ui/templates/ui to be consistent with other apps. No templates were actually changed. Unused templates were deleted.
- Major cleanup of opus.js and associated changes to other JS files.
  - Reorg to make like functions together
  - Style changes to variable names and functions (camelCase, proper verbs in function names, etc.)
  - First pass at following JSDoc function document style (/** ... */)
  - Fixed bug in search tab spinners when tabs switched quickly
  - Removed dead deferred object code
  - Cleaned up retrieval of menu when reset buttons pressed


opus/application/static_media/js/opus.js
```diff
@@ -114,227 +141,284 @@ var opus = {
...
+         * This is the event handler for the user clicking on one of the main nav
+         * bar tabs (views).
+         */
+
+        // First hide everything and stop any interval timers
+        $("#search, #detail, #cart, #browse").hide();
         o_browse.hideMenu();
 
-        // close any open modals
+        // Close any open modals
```
dstopp, line 335: what about the metadata modal? do we need to worry about closing that here as well?  if so, do we need a "closeModals' function?

dstopp (changes requested):

opus/application/static_media/js/opus.js
rfrenchseti: There's a separate handler in browse.js for ESC on the metadata modal and also the select metadata modal. It also does things like stop the range select and, for no obvious reason, close the help pane. I think the right thing to do is to leave the metadata modal and select metadata modal in browse.js because they have other actions associated with them (like asking for confirmation closing the select metadata...eventually), and move the close help pane to opus.js. It's sort of a tricky decision because the confirm modals are scattered all over the place, so it's difficult to have a handler for each namespace separately. Putting them in opus.js seems like the best choice.

PR #683, opened 2019-05-16
New ui infinite scroll issues B087, B088, B089, B090

juzen2003:
# Update
- B087: After resizing window and toggling between table & gallery, slider value will not change now.
- B088: InfiniteScroll scroll up:
    - The previous scroll up issue:
        1. When we scroll the slider to the middle, and hold down pageUp key to scroll up.
        2. When scrollbar reaches to the up scroll threshold point, in "ps-scroll-up" event handler function, infiniteScroll loadNextPage will be triggered with the standard steps:
            a. infiniteScroll request
            b. infiniteScroll load
            c. infiniteScroll load event handler (renderGalleryAndTable is called here)
        3. And after that it will only have one loadNextPage triggered (in step 2), and scrollbar will be stuck at the top end. No more data will be prefetched unless we use the wheel or lower the scroll bar a bit and move it up again.
    - The Reason:
        In step 2, we are still holding down pageUp key, and still getting into "ps-scroll-up" event to call the handler function to perform loadNextPage for a second time. However, when we try to call loadNextPage for the 2nd time, the first trigger of loadNextPage is still in the middle of 2-a and 2-b. Therefore, the 2nd time loadNextPage will not be called because infiniteScroll thinks it's already been doing loadNextPage (in the middle of 2-a and 2-b). And the scrollbar will be stuck at the top end.
    - Now in checkScroll, we will trigger "ps-scroll-up" when scrollbar reaches to up scroll threshold point.
- B089: We set the fixed PS scrollbar length to 100 for now.
- B090: 0.5S debounce time is removed.

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -670,7 +670,7 @@ var o_browse = {
     },
 
     // find the first displayed observation index & id in the upper left corner
-    updateSliderHandle: function(resize=false) {
+    updateSliderHandle: function(browserResized=false) {
         let tab = opus.getViewTab();
         let browse = o_browse.getBrowseView();
         let selector = ((opus.prefs[browse] === "gallery") ?
@@ -708,7 +708,7 @@ var o_browse = {
                 // At this point of time, galleryBoundingRect is updated with new row size
                 // from countGalleryImages in adjustBrowseHeight.
                 let numToDelete = ((galleryBoundingRect.x - (firstCachedObs - 1) % galleryBoundingRect.x) %
-                galleryBoundingRect.x);
+                                   galleryBoundingRect.x);
 
                 let galleryObsElem = $(`${tab} .gallery [data-obs]`);
                 let tableObsElem = $(`${tab} .op-data-table-view [data-obs]`);
@@ -1651,9 +1651,6 @@ var o_browse = {
         let tab = opus.getViewTab(view);
         let footerHeight = $(".app-footer").outerHeight();
         let mainNavHeight = $("#op-main-nav").outerHeight();
-        // we need to specify the tab here
-        // otherwise in cart tab the selector will select the navbar in browse
-        // and outerHeight will be 0.
         let navbarHeight = $(`${tab} .panel-heading`).outerHeight();
         let totalNonGalleryHeight = footerHeight + mainNavHeight + navbarHeight;
         return  $(window).height()-totalNonGalleryHeight;
@@ -1672,7 +1669,7 @@ var o_browse = {
         return width;
     },
 
-    adjustBrowseHeight: function(resize=false) {
+    adjustBrowseHeight: function(browserResized=false) {
         let tab = opus.getViewTab();
         let containerHeight = o_browse.calculateGalleryHeight();
         $(`${tab} .gallery-contents`).height(containerHeight);
@@ -1683,7 +1680,7 @@ var o_browse = {
         namespace.galleryBoundingRect = o_browse.countGalleryImages();
 
         // make sure slider is updated when window is resized
-        o_browse.updateSliderHandle(resize);
+        o_browse.updateSliderHandle(browserResized);
     },
 
     adjustTableSize: function() {
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -1406,7 +1406,7 @@ var o_browse = {
         }
     },
 
-    deleteCachedObservation: function (galleryObsElem, tableObsElem, index) {
+    deleteCachedObservation: function(galleryObsElem, tableObsElem, index) {
         // don't delete the metadata if the observation is in the cart
         if (!galleryObsElem.eq(index).hasClass("in")) {
             let delOpusId = galleryObsElem.eq(index).data("id");
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -670,9 +670,8 @@ var o_browse = {
     },
 
     // find the first displayed observation index & id in the upper left corner
-    updateSliderHandle: function() {
+    updateSliderHandle: function(resize=false) {
         let tab = opus.getViewTab();
-        let contentsView = o_browse.getScrollContainerClass();
         let browse = o_browse.getBrowseView();
         let selector = ((opus.prefs[browse] === "gallery") ?
                         `${tab} .gallery .thumbnail-container` :
@@ -699,27 +698,41 @@ var o_browse = {
 
             // table: obsNum = calculatedFirstObs + number of row
             // gallery: obsNum = calculatedFirstObs + number of row * number of obs in a row
-            let obsNumDiff = ((opus.prefs[browse] === "dataTable") ?
-                              Math.round((topBoxBoundary - firstCachedObsTop)/$(`${tab} tbody tr`).outerHeight()) :
-                              Math.round((topBoxBoundary - firstCachedObsTop)/o_browse.imageSize) * galleryBoundingRect.x);
+            let obsNumDiff = ((opus.prefs[browse] === "gallery") ?
+                              Math.round((topBoxBoundary - firstCachedObsTop)/o_browse.imageSize) *
+                              galleryBoundingRect.x :
+                              Math.round((topBoxBoundary - firstCachedObsTop)/$(`${tab} tbody tr`).outerHeight()));
             let obsNum = obsNumDiff + calculatedFirstObs;
 
-            let numToDelete = ((galleryBoundingRect.x - (firstCachedObs - 1) % galleryBoundingRect.x) %
-                               galleryBoundingRect.x);
+            if (resize) {
+                let numToDelete = ((galleryBoundingRect.x - (firstCachedObs - 1) % galleryBoundingRect.x) %
+                galleryBoundingRect.x);
 
-            let galleryObsElem = $(`${tab} .gallery [data-obs]`);
-            let tableObsElem = $(`${tab} .op-data-table-view [data-obs]`);
-            // delete first "numToDelete" obs if row size is changed
-            if (numToDelete !== 0) {
-                for (let count = 0; count < numToDelete; count++) {
-                    o_browse.deleteCachedObservation(galleryObsElem, tableObsElem, count);
+                console.log("=== numToDelete ===");
+                console.log(numToDelete);
+                let galleryObsElem = $(`${tab} .gallery [data-obs]`);
+                let tableObsElem = $(`${tab} .op-data-table-view [data-obs]`);
+                console.log(galleryObsElem);
+                // delete first "numToDelete" obs if row size is changed
+                if (numToDelete !== 0) {
+                    for (let count = 0; count < numToDelete; count++) {
+                        o_browse.deleteCachedObservation(galleryObsElem, tableObsElem, count);
+                    }
                 }
             }
+            console.log(galleryBoundingRect);
+            console.log("=== updateSliderHandle firstObs ===");
+            console.log(firstCachedObs);
+            console.log(calculatedFirstObs);
+            console.log("=== Row height ===");
+            console.log(obsNumDiff);
+            console.log("=== startObs ===");
+            console.log(obsNum);
 
             // Update obsNum in both infiniteScroll instances.
             // Store the most top left obsNum in gallery for both infiniteScroll instances
             // (this will be used to updated slider obsNum).
-            if (contentsView === ".op-data-table-view") {
+            if (opus.prefs[browse] !== "gallery") {
                 obsNum = (Math.floor((obsNum - 1)/galleryBoundingRect.x + 0.0000001) *
                           galleryBoundingRect.x + 1);
             }
@@ -1047,7 +1060,7 @@ var o_browse = {
 
             $(".op-browse-view", tab).html("<i class='far fa-list-alt'></i>&nbsp;View Table");
             $(".op-browse-view", tab).attr("title", "View sortable metadata table");
-            $(".op-browse-view", tab).data("view", "dataTable");
+            $(".op-browse-view", tab).data("view", "data");
 
             suppressScrollY = false;
         } else {
@@ -1410,6 +1423,8 @@ var o_browse = {
         }
         galleryObsElem.eq(index).remove();
         tableObsElem.eq(index).remove();
+        console.log("=== galleryObsElm id being removed ===");
+        console.log(galleryObsElem.eq(index).data("obs"));
     },
 
     getBrowseView: function () {
@@ -1675,7 +1690,7 @@ var o_browse = {
         namespace.galleryBoundingRect = o_browse.countGalleryImages();
 
         // make sure slider is updated when window is resized
-        o_browse.updateSliderHandle();
+        o_browse.updateSliderHandle(true);
     },
 
     adjustTableSize: function() {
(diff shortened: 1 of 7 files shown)


opus/application/static_media/js/browse.js
```diff
@@ -14,6 +14,8 @@ const tableSortUpArrow = "fas fa-sort-up";
 const tableSortDownArrow = "fas fa-sort-down";
 const defaultTableSortArrow = "fas fa-sort";
 const infiniteScrollUpThreshold = 100;
+// Fixed scrollbar length for gallery & table view
+const galleryAndTabPSLength = 100;
```
rfrenchseti: Rename to galleryAndTablePSLength

opus/application/static_media/js/browse.js
```diff
@@ -55,7 +59,8 @@ var o_browse = {
     browseBehaviors: function() {
         // note: using .on vs .click allows elements to be added dynamically w/out bind/rebind of handler
 
-        $(".op-gallery-view, .op-dataTable-view").on("scroll", _.debounce(o_browse.checkScroll, 500));
+        // $(".op-gallery-view, .op-dataTable-view").on("scroll", _.debounce(o_browse.checkScroll, 500));
```
rfrenchseti: Remove old code

opus/application/static_media/js/browse.js
```diff
@@ -635,12 +638,13 @@ var o_browse = {
         let elem = $(`#${opus.prefs.view} .thumbnail-container[data-obs="${value}"]`);
         let startObsLabel = o_browse.getStartObsLabel();
 
+        // Update obsNum in infiniteScroll instances, and this obsNum is the first item in current page. (will be used to set scrollbar position in renderGalleryAndTable).
```
rfrenchseti: Please break the long comments across multiple lines so they don't go past the edge of a reasonable browser width.

opus/application/static_media/js/browse.js
```diff
@@ -635,12 +638,13 @@ var o_browse = {
...
+        opus.prefs[startObsLabel] = value;
+
         if (elem.length > 0) {
             o_browse.setScrollbarOnSlide(value);
-            // Update obsNum in infiniteScroll instances, and this obsNum is the first item in current page. (will be used to set scrollbar position in renderGalleryAndTable). This is for the case when above o_browse.setScrollbarOnSlide(value) trigger infiniteScroll load event.
-            $(`${tab} .op-gallery-view`).infiniteScroll({"obsNum": value});
-            $(`${tab} .op-dataTable-view`).infiniteScroll({"obsNum": value});
-            opus.prefs[startObsLabel] = value;
         } else {
             // When scrolling on slider and loadData is called, we will fetch 3 * getLimit items (one current page, one next page, and one previous page) starting from obsNum.
```
rfrenchseti: Same for this comment

opus/application/static_media/js/browse.js
```diff
@@ -666,47 +666,52 @@ var o_browse = {
...
+            let firstCachedObsTop = $(selector).first().offset().top;
             topBoxBoundary = topBoxBoundary || (opus.prefs.browse === "dataTable") ? $(`${tab} .gallery-contents`).offset().top + $(`${tab} #dataTable thead th`).outerHeight() : $(`${tab} .gallery-contents`).offset().top;
 
-            // compare the image .top + half its height in order to make sure we account for partial images
-            let topBox = $(elem).offset().top + $(elem).height()/2;
-            if (topBox >= topBoxBoundary) {
-                let obsNum = $(elem).data("obs");
+            // table: obsNum = firstCachedObs + number of row
+            // gallery: obsNum = firstCachedObs + number of row * number of obs in a row
+            let obsNumDiff = (opus.prefs.browse === "dataTable") ? Math.round((topBoxBoundary - firstCachedObsTop)/$(`${tab} tbody tr`).outerHeight()) : Math.round((topBoxBoundary - firstCachedObsTop)/o_browse.imageSize) * o_browse.galleryBoundingRect.x;
```
rfrenchseti: Break into multiple lines so it's more readable

rfrenchseti: Also, this won't work for the cart, will it? opus.prefs.browse is only for the main browse tab. Need to use Debby's new functions that retrieve the proper browse name.

rfrenchseti (changes requested):

opus/application/static_media/js/browse.js
dstopp: a better variable name might be galleryAndTablePSLength

opus/application/static_media/js/browse.js
dstopp: do we still need this?

dstopp (changes requested):

juzen2003:
# Update:
- Merge with new_ui
- Implement the method to delete first N obs when window is resized (as discussed in slack)

opus/application/static_media/js/browse.js
```diff
@@ -675,56 +674,86 @@ var o_browse = {
...
-                $(`${tab} .op-gallery-view`).infiniteScroll({"obsNum": obsNum});
-                if (contentsView === ".op-data-table-view") {
-                    obsNum = Math.floor((obsNum-1)/o_browse.gallerySliderStep+0.0000001)*o_browse.gallerySliderStep+1;
+                              (opus.prefs[browse]  === "gallery") ?
+                              $(`${tab} .gallery-contents`).offset().top :
+                              $(`${tab} .gallery-contents`).offset().top + $(`${tab} .op-data-table thead th`).outerHeight());
+
+            // table: obsNum = calculatedFirstObs + number of row
+            // gallery: obsNum = calculatedFirstObs + number of row * number of obs in a row
+            let obsNumDiff = ((opus.prefs[browse] === "dataTable") ?
```
dstopp: i don't think this is "dataTable" - I believe this is "data".

opus/application/static_media/js/browse.js
```diff
@@ -675,56 +674,86 @@ var o_browse = {
...
+                    o_browse.deleteCachedObservation(galleryObsElem, tableObsElem, count);
                 }
+            }
 
-                $(`${tab} .op-data-table-view`).infiniteScroll({"obsNum": obsNum});
-                opus.prefs[startObsLabel] = obsNum;
+            // Update obsNum in both infiniteScroll instances.
+            // Store the most top left obsNum in gallery for both infiniteScroll instances
+            // (this will be used to updated slider obsNum).
+            if (contentsView === ".op-data-table-view") {
```
dstopp: i find it confusing to check the container class vs. whether this is gallery or table view.

dstopp (changes requested):

opus/application/static_media/js/browse.js
rfrenchseti: Actually, no. The slug value in the URL is data, but the value in opus.prefs is dataTable. This is something Debby put in a while back when she didn't realize it would cause backwards compatibility problems with the URL. We fixed it in hash.js by doing the mapping when parsing/creating the URL. I hate this. And I just noticed a bug - hash.js only does the mapping for 'browse', not 'cart_browse'. See hash.js line 147. This needs to be fixed. I would MUCH rather get rid of dataTable entirely and just use data, though.

opus/application/static_media/js/browse.js
rfrenchseti: Agreed

opus/application/static_media/js/browse.js
```diff
@@ -675,56 +674,86 @@ var o_browse = {
...
+
+            let numToDelete = ((galleryBoundingRect.x - (firstCachedObs - 1) % galleryBoundingRect.x) %
+                               galleryBoundingRect.x);
+
+            let galleryObsElem = $(`${tab} .gallery [data-obs]`);
+            let tableObsElem = $(`${tab} .op-data-table-view [data-obs]`);
+            // delete first "numToDelete" obs if row size is changed
+            if (numToDelete !== 0) {
+                for (let count = 0; count < numToDelete; count++) {
+                    o_browse.deleteCachedObservation(galleryObsElem, tableObsElem, count);
```
rfrenchseti: I'm a little confused by this. When we delete obs 0, doesn't that make the next first obs also become obs 0? So we should be deleting obs 0 "numToDelete" times rather than deleting increasing numbers? I think this has the effect of deleting obs 0, 1, 3, 6, ... (original numbers before the delete). Actually the same comment applies to manageObservationCache. Am I confused or does this function do the wrong thing?

rfrenchseti (changes requested):
- (B087) I still see the slider changing when I switch between the gallery and table views with almost any size browser, even when I haven't done a resize since the last reload
- B088 appears fixed
- B089 appears fixed
- B090 appears fixed

opus/application/static_media/js/browse.js
juzen2003: All dataTable related stuff is removed, now in slug, we will have data in browse or cart_browse when we are in table view.

opus/application/static_media/js/browse.js
juzen2003: The current function will do the right thing. When we pass in the galleryObsElem and tableObsElem, they are arrays of each thumbnail container and table tr. When deleting an element in an array, the element will be removed and replaced with an undefined item like the following example:
```js
let a = [6, 7, 8, 9, 10];
delete a[0];
delete a[1];
console.log(a); // [ <2 empty items>, 8, 9, 10 ]
console.log(a.length); // 5
for (let i = 0; i < a.length; i++) {
    console.log(a[i]); 
}
// the above loop will print out the following:
// undefined
// undefined
// 8
// 9
// 10
``` 
The case you mentioned above (to always delete index 0) will work if we don't pass in galleryObsElem and tableObsElem, and always re-declare both variables inside the deleteCachedObservation function like this:
```js
deleteCachedObservation: function (galleryObsElem, tableObsElem, index) {
    let galleryObsElem = $(`${tab} .gallery [data-obs]`);
    let tableObsElem = $(`${tab} .op-data-table-view [data-obs]`);
// ...
```
This will create a new array without the undefined (previous deleted obs) items.

juzen2003:
Update: fix the gallery and table height in cart, they are not covered by footer now.

rfrenchseti:
There's still something very wrong with the slider. If I move the slider to X, it jumps to some random nearby value while the gallery or table is loading. When I'm in the table view and move the slider, the first table entry is not the same as the slider value.

rfrenchseti:
I also keep having the scrollbar in the gallery disappearing in Chrome

opus/application/static_media/js/browse.js
rfrenchseti: OK, I understand what's going on now. It works fine. Thanks.

rfrenchseti:
Gallery and table height seem to work now.
I think we have to fix the Chrome slider issue before releasing beta. It would be nice to fix the slider, but it doesn't have to delay the beta.

opus/application/static_media/js/browse.js
```diff
@@ -660,71 +664,100 @@ var o_browse = {
...
-            $(`${tab} .op-data-table-view`).infiniteScroll({"obsNum": value});
-            opus.prefs[startObsLabel] = value;
             o_browse.galleryBegun = false;
             o_browse.loadData(view, obsNum, customizedLimitNum);
         }
     },
 
     // find the first displayed observation index & id in the upper left corner
-    updateSliderHandle: function() {
+    updateSliderHandle: function(resize=false) {
```
rfrenchseti: "resize" makes this sound like something updateSlider should do itself. But this is called when a resize has already happened. How about something like "browserResized"?

opus/application/static_media/js/browse.js
```diff
@@ -660,71 +664,100 @@ var o_browse = {
...
+                              Math.round((topBoxBoundary - firstCachedObsTop)/o_browse.imageSize) *
+                              galleryBoundingRect.x :
+                              Math.round((topBoxBoundary - firstCachedObsTop)/$(`${tab} tbody tr`).outerHeight()));
+            let obsNum = obsNumDiff + calculatedFirstObs;
+
+            if (resize) {
+                // At this point of time, galleryBoundingRect is updated with new row size
+                // from countGalleryImages in adjustBrowseHeight.
+                let numToDelete = ((galleryBoundingRect.x - (firstCachedObs - 1) % galleryBoundingRect.x) %
+                galleryBoundingRect.x);
```
rfrenchseti: Indent to be past above open paren

opus/application/static_media/js/browse.js
```diff
@@ -1475,7 +1506,8 @@ var o_browse = {
                 let lastObs = $(`${tab} .thumbnail-container`).last().data("obs");
 
                 if ((firstObs === 1 && infiniteScrollData.options.loadPrevPage === true) ||
-                    (lastObs === opus.resultCount && infiniteScrollData.options.loadPrevPage === false)) {
+                    (lastObs === opus.resultCount && infiniteScrollData.options.loadPrevPage === false) ||
+                    (lastObs === o_cart.cartCount && infiniteScrollData.options.loadPrevPage === false)) {
```
rfrenchseti: This seems like it can't work. You're always comparing lastObs to BOTH the result count and the cart count, regardless of which tab you're actually dealing with. So conceivably you could be equal to the cart count while looking at the browse page, and that's going to mess things up, right?

opus/application/static_media/js/browse.js
```diff
@@ -1619,7 +1651,10 @@ var o_browse = {
         let tab = opus.getViewTab(view);
         let footerHeight = $(".app-footer").outerHeight();
         let mainNavHeight = $("#op-main-nav").outerHeight();
-        let navbarHeight = $(".panel-heading").outerHeight();
+        // we need to specify the tab here
```
rfrenchseti: Do we need this comment? We select using tab all over the place without explaining why.

opus/application/static_media/js/browse.js
```diff
@@ -1637,9 +1672,9 @@ var o_browse = {
         return width;
     },
 
-    adjustBrowseHeight: function() {
+    adjustBrowseHeight: function(resize=false) {
```
rfrenchseti: Same comment on "resize" naming as above

rfrenchseti (changes requested):
1) The Y scrollbar on the gallery is often missing from the browse tab using Chrome. I have no idea why.
2) The slider is misbehaving badly. When you drag it it doesn't stay on the value you select. The gallery and table don't go to the right obs. And when you switch back and forth between table and gallery the slider changes value.

opus/application/static_media/js/browse.js
juzen2003: lastObs is also depending on tab (the tab was defined in the callback function when we first init infiniteScroll). When it's in browse gallery view, it's trying to select the last rendered item in browse gallery. When it's in cart gallery view, it's trying to select the last rendered item in cart gallery.

rfrenchseti: But you could be on the browse tab and lastObs could equal the cart count, which would be wrong. Or vice verse. Checking for resultcount or cartcount won’t work by itself. You need to check the tab name also. Also it would be good to pull the loadPrevPage out so you aren’t checking for false in each case. Would be easier to read.

rfrenchseti:
I believe we need to make the following changes:
- setScrollbarPosition() needs to take an optional view because it is called from renderGalleryAndTable and loadData().
- manageObservationCache() needs to take an optional view because it is called from renderGalleryAndTable. Note: I think o_browse.maxCachedObservations needs to be view-specific. Right now there is just one global. Really, I think in the end we need two separate galleryData arrays.
- metadataboxHtml() needs to take an optional view because it is called from loadData. This is then passed to getNextPrevHandles, which also needs to take a view.
- The call to `let contentsView = o_browse.getScrollContainerClass();` at line 1511 needs a view since it's inside of initInfiniteScroll.
- This line: `$(selector).on("load.infiniteScroll", o_browse.infiniteScrollLoadEventListener);` which is inside of initInfiniteScroll, which takes a view, should become a closure that passes in the view along with the other normal arguments. Note that currently infiniteScrollLoadEventListener calls renderGalleryAndTable without passing a view.

I've changed items 1-4 in the merged branch. Please confirm them and I can add item #5.

rfrenchseti:
Also Dave forgot to add the following line to updateBrowseNav:
`        o_browse.fading = true;`
So fading was never being set to true anywhere. I've fixed this as well.

rfrenchseti (approved):

PR #690, opened 2019-05-21
JS cleanup of opus.js and associated files

rfrenchseti:


opus/application/static_media/js/search.js
```diff
@@ -261,7 +261,7 @@ var o_search = {
 
                // special menu behavior for surface geo, slide in a loading indicator..
                if (id == 'surfacetarget') {
-                    let surface_loading = '<li style = "margin-left:50%; display:none" class = "spinner">&nbsp;</li>';
+                    let surface_loading = '<li style = "margin-left:50%; display:none" class = "opus.spinner">&nbsp;</li>';
```
dstopp: i think "opus.spinner" is global and that should be ${opus.spinner}, no?  <of sorts>

dstopp (changes requested):

dstopp (approved):

PR #692, opened 2019-05-23
New ui provide info or warning messages to users, related to A095/A098/A096/S055

juzen2003:
# Update:
- A095:
  - If user's browser is older than the following version: Firefox (66+), Chrome (74+), Safari (12.1+), Opera (58+), and Microsoft Edge (18+), we display a modal to inform the user to update the browser version. The modal is not closable, so the user cannot proceed to use OPUS until he/she updates the browser version.
- A098:
  - If browser window width is less than 1280px, we display a modal to inform the user to resize it. The modal will not go away and user cannot click anywhere in OPUS until the browser width is resized to some values larger than 1280px.
- A096:
  - If the widget cookies is undefined, we know the user is a first time visitor. We display a welcome modal for now, but will update to display the real guide page in the future.

### DIFF
--- a/opus/application/apps/help/templates/help/about.html
+++ b/opus/application/apps/help/templates/help/about.html
@@ -177,7 +177,7 @@
 </p>
 
 <p>Please note that OPUS works best with a browser size of at least 1280 by 720. We
-	support Firefox (66+), Chrome (74+), Safari (12.1+), Opera (58+), and Microsoft Edge (18+).
+	support Firefox (59+), Chrome (56+), Safari (10.1+), and Opera (42+).
 </p>
 
 <p>Updates are announced on
--- a/opus/application/apps/help/templates/help/about.html
+++ b/opus/application/apps/help/templates/help/about.html
@@ -176,7 +176,7 @@
     We would also appreciate if you would drop us a note to let us know about it.
 </p>
 
-<p>Please note that OPUS works best with a browser size of at least 1280 by 720. We
+<p>Please note that OPUS works best with a browser size of at least 600 by 200. We
 	support Firefox (59+), Chrome (56+), Safari (10.1+), and Opera (42+).
 </p>
 
--- a/opus/application/apps/help/templates/help/about.html
+++ b/opus/application/apps/help/templates/help/about.html
@@ -176,7 +176,7 @@
     We would also appreciate if you would drop us a note to let us know about it.
 </p>
 
-<p>Please note that OPUS works best with a browser size of at least 1280 by 275. We
+<p>Please note that OPUS works best with a browser size of at least 1280 by 720. We
 	support Firefox (66+), Chrome (74+), Safari (12.1+), Opera (58+), and Microsoft Edge (18+).
 </p>
 
--- a/opus/application/apps/help/templates/help/about.html
+++ b/opus/application/apps/help/templates/help/about.html
@@ -176,7 +176,7 @@
     We would also appreciate if you would drop us a note to let us know about it.
 </p>
 
-<p>Please note that OPUS works best with a browser size of at least 1280 by 720. We
+<p>Please note that OPUS works best with a browser size of at least 1280 by 275. We
 	support Firefox (66+), Chrome (74+), Safari (12.1+), Opera (58+), and Microsoft Edge (18+).
 </p>
 
--- a/opus/application/apps/ui/templates/ui/base.html
+++ b/opus/application/apps/ui/templates/ui/base.html
@@ -356,6 +356,14 @@
             target="op-browser-size-msg"
             allow_close=False
         %}
+        {% include "ui/confirm_modal.html" with
+            id_name="op-guide"
+            confirm_title="Welcome to OPUS3"
+            confirm_msg="Please see the FAQ section from help menu or <a href='https://ringsnodesearchtool.blogspot.com/' target='_blank'>our blog</a> for details."
+            target="op-guide"
+            yes_text="OK"
+            allow_close=True
+        %}
 
         {% endblock %}
 
--- a/opus/application/apps/ui/templates/ui/base.html
+++ b/opus/application/apps/ui/templates/ui/base.html
@@ -306,6 +306,7 @@
             yes_text="Yes"
             no_text="No"
             allow_close=True
+            close_by_keyboard="esc"
         %}
         {% include "ui/confirm_modal.html" with
             id_name="op-reset-search-metadata-modal"
@@ -315,6 +316,7 @@
             yes_text="Yes"
             no_text="No"
             allow_close=True
+            close_by_keyboard="esc"
         %}
         {% include "ui/confirm_modal.html" with
             id_name="op-empty-cart"
@@ -324,6 +326,7 @@
             yes_text="Yes"
             no_text="No"
             allow_close=True
+            close_by_keyboard="esc"
         %}
         {% include "ui/confirm_modal.html" with
             id_name="op-cart-status-error-msg"
@@ -332,6 +335,7 @@
             target="op-cart-status-error-msg"
             yes_text="OK"
             allow_close=True
+            close_by_keyboard="esc"
         %}
         {% include "ui/confirm_modal.html" with
             id_name="op-update-url"
@@ -341,6 +345,7 @@
             yes_text="Keep Message"
             no_text="Discard Message"
             allow_close=True
+            close_by_keyboard="esc"
         %}
         {% include "ui/confirm_modal.html" with
             id_name="op-browser-version-msg"
@@ -363,6 +368,7 @@
             target="op-guide"
             yes_text="OK"
             allow_close=True
+            close_by_keyboard="esc"
         %}
 
         {% endblock %}
--- a/opus/application/apps/ui/templates/ui/base.html
+++ b/opus/application/apps/ui/templates/ui/base.html
@@ -343,10 +343,17 @@
             allow_close=True
         %}
         {% include "ui/confirm_modal.html" with
-            id_name="op-browser-warning"
+            id_name="op-browser-version-msg"
             confirm_title="Message About Your Browser Version"
-            confirm_msg="Please update your browser. We support Firefox (66+), Chrome (74+), Safari (12.1+), Opera (58+), and Microsoft Edge (18+)."
-            target="op-browser-warning"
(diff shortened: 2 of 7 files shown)


opus/application/static_media/js/opus.js
```diff
@@ -724,11 +734,91 @@ var opus = {
...
+         * modal to inform the user to resize the browser width.
+         */
+        if ($(window).width() < opus.browserSupport.width) {
+            $("#op-browser-size-msg").modal("show");
+        } else {
+            $("#op-browser-size-msg").modal("hide");
+        }
+    },
+
+    checkCookies: function() {
```
rfrenchseti, line 831: We're going to get rid of the widgets cookie, which doesn't currently do anything anyway. We should use some unique cookie for determining if the user is new.

rfrenchseti: That means we should also set the cookie here. Don't rely on the widget code to do it for you.

opus/application/static_media/js/opus.js
```diff
@@ -724,11 +734,91 @@ var opus = {
...
+         */
+        if ($.cookie("widgets") === undefined) {
+            $("#op-guide").modal("show");
+        }
+    }
 }; // end opus namespace
 
 $(document).ready(function() {
+    opus.checkBrowserVersion();
+    opus.checkCookies();
```
rfrenchseti: This is probably the place you need to above if checkBrowserVersion returns saying the browser is bad. Don't do anything else, including trying to initialize OPUS. Just give up.

rfrenchseti (changes requested):
- When you decrease the browser size below 1280px, and then keep decreasing it, there's a limit as to how far the dialog will move to the left. It doesn't just stay centered on the screen. Eventually you have a big space on the left and the dialog is cut off on the right. This probably won't work well on a mobile device.
- If you are both a new user and are running an old browser, you get both modal dialogs. But the new user one is on top, so you don't see the browser error. Also when you click OK on that one, both dialogs get closed so you *never* see the one about the browser version. OPUS just fails to work.
- Is there a way to tell the user which browser and version they are currently using?
- Add a catch for the browser being too short of 275px. Change the About message accordingly.

opus/application/apps/ui/templates/ui/base.html
```diff
@@ -336,6 +340,29 @@ <h3 class='op-header-text'></h3>
...
+            id_name="op-browser-version-msg"
+            confirm_title="Message About Your Browser Version"
+            confirm_msg="Please update your browser. OPUS supports Firefox (66+), Chrome (74+), Safari (12.1+), Opera (58+), and Microsoft Edge (18+)."
+            target="op-browser-version-msg"
+            allow_close=False
+        %}
+        {% include "ui/confirm_modal.html" with
+            id_name="op-browser-size-msg"
+            confirm_title="Message About Your Browser Size"
+            confirm_msg="Please resize your browser. OPUS works best with a browser width of at least 1280px."
```
rfrenchseti: Also let's spell out the word "pixels" here. Nobody knows what a "px" is. Although pixels is not really technically accurate, at least people will have heard of it.

rfrenchseti (changes requested):

juzen2003:
# Update
- In checkCookies, we don't use widgets cookies now, we check and set a new visited cookies.
- In checkBrowserSize, it will also check if the browser height is 275px now.
- In isBrowserSupported, we will also print out users' browser version if the modal is displayed.
- In document ready, OPUS init process will not be run if browser is not supported.
- Now click ok or "x" in one modal will only close that modal. Rest of modal will still exists.
- For the following issue, it's because we attach jquery-ui draggable to all modals. It's not centered because it can still be dragged.

> When you decrease the browser size below 1280px, and then keep decreasing it, there's a limit as to how far the dialog will move to the left. It doesn't just stay centered on the screen. Eventually you have a big space on the left and the dialog is cut off on the right. This probably won't work well on a mobile device.

Now we center the modal and disable draggable for the following 3 modals: #op-browser-version-msg, #op-browser-size-msg, #op-guide. They will stay at the center, and when screen gets narrower, they will always be at the center until the screen width is about the same as the modal width.

opus/application/static_media/js/opus.js
juzen2003: Now we check and set visited cookies here.

opus/application/static_media/js/opus.js
```diff
@@ -724,12 +741,108 @@ var opus = {
...
+    checkCookies: function() {
+        /**
+         * Check widgets cookie to determine if the user is a first time
+         * visitor. If so, we display a guide page.
+         * Note: we will call __help/splash.html api in the future to
+         * display the guide page. For now, we just show a modal.
+         */
+        if ($.cookie("visited") === undefined) {
+            // set the cookie for the first time user
+            $.cookie("visited", true, { expires: 28});  // days
```
rfrenchseti: So the way this is currently implemented, a user can use OPUS every day  and then on the 28th day, get the splash dialog again, right? That seems wrong. We should either make the cookie never expire, or we should reset the 28 day counter every time the user runs OPUS (set it outside of the if).

opus/application/apps/help/templates/help/about.html
```diff
@@ -176,7 +176,7 @@
     We would also appreciate if you would drop us a note to let us know about it.
 </p>
 
-<p>Please note that OPUS works best with a browser size of at least 1280 by 720. We
+<p>Please note that OPUS works best with a browser size of at least 1280 by 275. We
```
rfrenchseti: 275 is not working the way I expected. Perhaps I screwed up using the inspector tool. The idea is that in the gallery view with the metadata detail modal popped up, it should fit completely on the screen. Otherwise, if the browser is too short, a bogus scrollbar shows up. So figure out how tall that is, round it to something pretty, and use that value.

rfrenchseti (changes requested):

opus/application/static_media/js/opus.js
```diff
@@ -646,13 +659,17 @@ var opus = {
         $(document).on("keydown click", function(e) {
             if ((e.which || e.keyCode) == 27) {
                 // ESC key - close modals and help panel
-                $(".op-confirm-modal").modal('hide');
+                // But don't close "#op-browser-version-msg" and "#op-browser-size-msg"
+                if (!$(".op-confirm-modal").is("#op-browser-version-msg") &&
```
dstopp: i wonder if it would be useful in the future to have a class or data element on the modal that identifies it as "close-able on esc", instead of having to check the id of the modal.    Then you could just do something like:
if ($(".op-confirm-modal").data("action") === "esc")    more or less, doesn't have to be exactly this.

opus/application/static_media/js/opus.js
```diff
@@ -667,7 +684,7 @@ var opus = {
                             o_cart.emptyCart();
                             break;
                     }
-                    $(".modal").modal("hide");
+                    $(`#${target}`).modal("hide");
```
dstopp, line 688: don't you need the same fix for the "cancel"...?

opus/application/static_media/js/opus.js
```diff
@@ -646,13 +659,17 @@ var opus = {
...
+                    !$(".op-confirm-modal").is("#op-browser-size-msg")) {
+                    $(".op-confirm-modal").modal("hide");
+                }
                 opus.hideHelpPanel();
             }
         });
 
         // Handle the Submit or Cancel buttons for the various confirm modals we can pop up
-        $(".op-confirm-modal").on("click", ".btn", function() {
+        $(".op-confirm-modal").on("click", ".btn", function(event) {
```
dstopp: while i think it's good form to add the parameter event, i'm not sure i see where you are using it...  I also think it might be good to standardize on 'e' vs 'event' as a name, we use both.

opus/application/static_media/js/opus.js
```diff
@@ -724,12 +741,112 @@ var opus = {
...
+            browserName = "Chrome";
+            browserVersion = matchObj[1];
+        } else if (userAgent.indexOf("Version") > -1) {
+            // userAgent example output:
+            // Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_4) AppleWebKit/605.1.15
+            // (KHTML, like Gecko) Version/12.1 Safari/605.1.15
+            matchObj = userAgent.match(/Version\/(\d+.\d+)/);
+            browserName = "Safari";
+            browserVersion = matchObj[1];
+        }
```
dstopp, line 787: what if it's none of the "else if" group....?

dstopp (changes requested):

rfrenchseti:
I did a new investigation of browser features, and I think we can support:

FF 59+
Chrome 56+
Safari 10.1+
Opera 42+

We don't support Edge at all

rfrenchseti:
Also for right now let's make the required browser size more like what we eventually intend to support (like 600x200). It's really annoying to debug OPUS otherwise.

dstopp (approved):

rfrenchseti (approved):

opus/application/static_media/js/opus.js
```diff
@@ -794,10 +786,16 @@ var opus = {
...
-        let modalMsg = (`Your current browser is ${browserName} ${browserVersion}.
-                        Please update the browser. OPUS supports Firefox (66+),
-                        Chrome (74+), Safari (12.1+), Opera (58+), and Microsoft Edge
-                        (18+).`);
+        let browser = `${browserName} ${browserVersion}`;
+        if (browserName === "unsupported") {
+            browser = browserName;
+        }
+        let modalMsg = (`Your current browser is ${browser}.
+                        Please update the browser. OPUS supports
```
rfrenchseti: One more change here. Right now it will say "Your current browser is unsupported. Please update the browser." That doesn't make much sense. Move the "Please update the browser." into the let statement at 789 and then it will work better.

Also let's change it to "Please update your browser."

opus/application/apps/help/templates/help/about.html
```diff
@@ -177,7 +177,7 @@
 </p>
 
 <p>Please note that OPUS works best with a browser size of at least 1280 by 720. We
```
rfrenchseti: This message needs to be changed to the new sizes, doesn't it?

rfrenchseti (changes requested):

rfrenchseti (approved):

PR #695, opened 2019-05-25
B081 adjust cache size

dstopp:
update the default cache size of 1000 to something based on screen size/number of observations viewable on the browser tab.

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -40,6 +40,7 @@ var o_browse = {
     galleryBegun: false, // have we started the gallery view
     galleryData: {},  // holds gallery column data
     imageSize: 100,     // default
+    cachedObservationFactor: 4,     // this is the factor times the screen size to determine cache size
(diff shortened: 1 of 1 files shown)


rfrenchseti (commented):
So right now when we need to remove observations from the cache because of a page load, we only remove as many as we are adding. That means that if you start with a large browser and then shrink it, the cache will always contain a number of observations much larger than the current maximum, right? At least until resetData() gets called at some point. Do you think we should clean up the cache here when the max cache size shrinks?

dstopp:
On any resize event, adjustBrowseHeight calls updateSliderHandle, which checks the size of the cached observations and adjusts or retrieves as needed, so I don't believe we need any new code to handle that case.

rfrenchseti:
I don't see anything in updateSliderHandle that reduces the cache size back down (or loads data), other than removing a few observations from the first cached row to make sure the row alignment is correct. Perhaps you're thinking of onUpdateSlider, which isn't called on a resize. I'm going to open a different issue related to memory leaks in galleryData.

PR #696, opened 2019-05-28
startobs is getting set to 1 in two places and ignoring the hash

dstopp:
-commented out on the init
-initialized lastSelector before it is used on opus init

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -1177,8 +1177,17 @@ var o_browse = {
                 }
                 $(".gallery", tab).html(galleryHtml);
             } else {
-                // we've hit the end of the infinite scroll.
-                $(".op-page-loading-status > .loader").hide();
+                if (data.start_obs > data.result_count) {
+                    // handle a corner case where a user has changed the startobs to be greater than result_count
+                    // just reset back to 1 and get a new page
+                    opus.prefs[o_browse.getStartObsLabel()] = 1;
+                    o_hash.updateHash();
+                    $("#galleryViewContents").addClass("op-disabled");
+                    $(`${tab} ${contentsView}`).infiniteScroll("loadNextPage");
+                } else {
+                    // we've hit the end of the infinite scroll.
+                    $(".op-page-loading-status > .loader").hide();
+                }
                 return;
             }
         } else {
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -37,6 +37,7 @@ var o_browse = {
 
     galleryBegun: false, // have we started the gallery view
     galleryData: {},  // holds gallery column data
+    totalObsCount : undefined,
     imageSize: 100,     // default
     cachedObservationFactor: 4,     // this is the factor times the screen size to determine cache size
     maxCachedObservations: 1000,    // max number of observations to store in cache, will be updated based on screen size
@@ -138,7 +139,7 @@ var o_browse = {
             if (selectionsHashStr !== "") {
                 selectionsHashStr += "&";
             }
-            let resultCountStr = opus.resultCount.toString();
+            let resultCountStr = o_browse.totalObsCount.toString();
             let orderStr = opus.prefs.order.join(",");
             let csvLink = `/opus/__api/data.csv?${selectionsHashStr}cols=${colStr}&limit=${resultCountStr}&order=${orderStr}`;
             $(this).attr("href", csvLink);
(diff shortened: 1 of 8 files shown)


dstopp:
-Add check for result_count vs. startobs on return of data from dataimages.json; if the startobs > result_count, then change startobs to 1 and retrieve the data again.  This will handle any corner case if the user changed the startobs to something out of range.

opus/application/static_media/js/browse.js
```diff
@@ -1177,8 +1177,17 @@ var o_browse = {
...
             } else {
-                // we've hit the end of the infinite scroll.
-                $(".op-page-loading-status > .loader").hide();
+                if (data.start_obs > data.result_count) {
+                    // handle a corner case where a user has changed the startobs to be greater than result_count
+                    // just reset back to 1 and get a new page
+                    opus.prefs[o_browse.getStartObsLabel()] = 1;
+                    o_hash.updateHash();
+                    $("#galleryViewContents").addClass("op-disabled");
+                    $(`${tab} ${contentsView}`).infiniteScroll("loadNextPage");
```
rfrenchseti, line 1181: I think you need to add a .loader hide() here. If you drag the slider to the very end, the page loads and then the spinner doesn't go away.

rfrenchseti (changes requested):

rfrenchseti:
Something really weird is happening with these changes. If you are on the BROWSE tab, drag the slider all the way to the end, let the page load, and then drag the slider somewhere else, you get a flash of data and then a your CART is empty message filling the screen.

opus/application/static_media/js/cart.js
```diff
@@ -16,7 +16,7 @@ var o_cart = {
     maxCachedObservations: 1000,    // max number of observations to store in cache, will be updated based on screen size
     lastRequestNo: 0,
     downloadInProcess: false,
-    cartCount: 0,
+    totalObsCount : undefined,
```
rfrenchseti: I tried to group the variables that are common to o_browse and o_cart at the top of the var declaration. I think it would be nice to keep them together. Also could add a comment saying these are in common, and another comment marking the start of the ones that are not in common.

opus/application/static_media/js/opus.js
```diff
@@ -713,6 +712,8 @@ var opus = {
         opus.prefs.widgets = [];
         o_widgets.updateWidgetCookies();
 
+        // set these to the current hash on opus init
+        [opus.lastSelections, opus.lastExtras] = o_hash.getSelectionsExtrasFromHash();
```
rfrenchseti, line 719: Could you add the assignment we discussed making force_load = true so it's obvious what's going on?

opus/application/static_media/js/browse.js
```diff
@@ -1486,8 +1498,12 @@ var o_browse = {
...
                         }
                     }
-                    let path = o_browse.getDataURL(view, obsNum, customizedLimitNum);
-                    return path;
+                    // if totalObsCount is not yet defined, we still need to init the infinite scroll...
+                    if (viewNamespace.totalObsCount === undefined || obsNum <= viewNamespace.totalObsCount) {
+                        let path = o_browse.getDataURL(view, obsNum, customizedLimitNum);
+                        return path;
+                    }
+                    // returning no value indicates end of infinite scroll...
```
rfrenchseti: When you hit the end of the observations, the console shows this error:
`infinite-scroll.pkgd.js:1265 GET http://127.0.0.1:8000/undefined 404 (Not Found)`
From that point on infinite scroll fails to work. I think Dave discovered that the returning of "undefined" doesn't work as documented. That's why we were still retrieving extra data even at the end, causing multiple API calls. He hadn't figured out a way around it. The fix you put in breaks scrolling entirely. It would be better to just have the multiple API calls for now.

rfrenchseti (changes requested):

opus/application/static_media/js/opus.js
dstopp: added w/comment

opus/application/static_media/js/browse.js
dstopp: hmm.  I can see that this funciton from infinite-scroll:
function request( url, responseType, onLoad, onError, onLast ) {
  var req = new XMLHttpRequest();
  req.open( 'GET', url, true );
appears to not bail when url is empty, thus being an issue.
I have posted a note to GIT regarding this; i can remove and note but this will also cause side effects.

PR #697, opened 2019-05-28
set the cartChange flag on sort order change.

dstopp:


rfrenchseti:
- Remove instances of `o_browse.galleryBegun = false;` in functions that call `renderSortedDataFromBeginning`

dstopp:
galleryBegun removed in 3 places.

rfrenchseti (changes requested):
So now we have the opposite problem - if you change the sort order while looking at the cart, the browse page doesn't get updated.

The more I look at this code, the more I see cases where galleryBegun is just plain doing the wrong thing. I think there are a lot more bugs hiding in here. Sadly, I think it may be time to just bite the bullet and clean this up. We would have a single variable per browse/cart namespace called something like "reloadData".

rfrenchseti:
This pull has been subsumed into another one.

PR #705, opened 2019-05-30
New ui responsive issues

juzen2003:
# Update
- Update css to have metadata selector modal responsive to window resizing. 
- Item in progress:
  - Make metadata dialog modal responsive.
- Note: in chrome, the minimum height we can shrink to is 296px.


juzen2003:
- Update css to have metadata display modal responsive to window resizing.

juzen2003:
- Update css to have search page responsive to window resizing.

rfrenchseti:
I tried out your new search changes. Something really weird happens when the browser is wider than the minimum required for the widgets to be at full width. As you make the browser even wider, the widgets actually *shrink*. Also the text inside of them does odd things. Also there's a big jump that happens when you pass 850px.

juzen2003:
The weird shrinking behavior when window gets wider is removed by removing the % of padding left/right of the widget area. This behavior could happen if we set a border-box element's padding with %. There are 2 break points (@media) at 860px and 970ox. At 860px, the  % of sidebar-area is properly tuned now, so there is no big jump. At 970px, this is the point when the q-type option is about to or just move to the 2nd line. And we are setting the % for text input at this point so that each widget can properly shrink when window gets narrow. Please note that each browser has slightly different default font size so the point where q-type option moves to the 2nd line is a bit different, but it's all around 970px range. (verified on Chrome, FF, Safari, and Opera).

rfrenchseti:
Two remaining problems:
1) When the left side start to shrink the "blue text" found under the ring and surface geometry constraints categories doesn't shrink and the words just get cut off
2) The "Reset Search and Metadata" button disappears entirely

juzen2003:
More break points are set in search page, when screen width is getting close to 600px:
- reset buttons will stay in one line.
- search menu will shrink and won't be cut off by the scrollbar.

juzen2003:
- Update to have browse tab responsive.
  - The 1st (top) navbar:
    - At 992px or less:
      - "Recent Announcements", "Questions/Feedback", and "Help" will be merged into the hamburger. 
      - Rest of tab options will stay at the same location.
      - When hamburger is open, 3 options inside will stay in one line.
  - The 2nd (bottom) navbar:
    - At 1000px or less: 
      - Slider will be moved to the 2nd line and take up all the width, and it will shrink proportionally when screen is getting narrower. 
      - The rest of 3 options will take up the width evenly at the 1st line, and they will shrink proportionally when screen is getting narrower.

juzen2003:
- Browse tab:
  - When slider is at the 2nd line, reduce its width to 75%. This will ensure the slider bar won't be cut off by the screen when it's moved to the end.
- Update to have detail tab responsive:
  - When screen is getting narrow:
    - Left pane will stop shrinking when it reaches to 500px, and detail image will keep shrinking. At 767px screen width (or less), the detail image is fairly small and will be moved from the right of the screen to the top left pane.

juzen2003:
- Update to have cart tab responsive to window resizing.
- In browse tab, when the slider is moved to the 2nd line, rest of 3 options in 2nd navbar will stay the same without shrinking or growing when screen width is changed.

opus/application/static_media/js/browse.js
```diff
@@ -267,7 +267,7 @@ var o_browse = {
 
         $(".modal-dialog").draggable({
             handle: ".modal-content",
-            cancel: ".metadata",
+            cancel: ".contents",   // Change from ".metadata" to ".contents"
```
rfrenchseti: You don't need to include the comment here - I was just telling you which line had to change

rfrenchseti (changes requested):
A few things it would be nice to fix before deployment:

- General:
  - The message that says the browser "works best with" should really say "requires" since we don't even let the user use OPUS below that size.

- Cart:
  - When the browser is fairly small (but still larger than 600x200) and you refresh or switch to the Cart tab, the dataimages.json API gets a call for a *negative* limit.
  - The "Empty Cart" button has moved down further than it was. It should be aligned with the top row of the secondary nav items (View Table, etc.) and stay up there even if the nav bar goes to two lines.
  - When the secondary nav bar becomes two lines, there is a big vertical gap between the first and second lines, much larger than between the second line and the sort order. This gap should be much smaller.
  - The "Download Options" button should have a "click here" mouse pointer (like the other nav bar options) and the text should not be selectable. Right now you can drag-select the text in the button.
  - When the browser gets very narrow, and the secondary nav bar breaks into two lines again, the "Download Options" button stays to the right of all of them, giving you weird things like "Data Archive   URL Archive   <- Download Options". The Download Options should always be on the left.
  - I think I'd rather see the Download Options be below the sort order to better separate it from the nav bar. This would fix the previous issue as well.
  - Also I think maybe it would be nice to make Download Options actually look like a button, or at least to something to bring attention to it.

- Detail:
  The responsiveness isn't working well for some image sizes. For example, look at co-vims-v1484510044_ir. The image and text below it will be partially cut off before it switches to being on top of the left column

A few things that could use tuning but we can do after deployment:

- General:
  - When the browser is narrow enough that the primary nav bar has a hamburger, there is slightly different spacing between the remaining nav items than there was without a hamburger.
  - When the browser is narrow enough that the primary nav bar has a hamburger, click on the hamburger, then Help. The (?) icon next to Help will not have any margin to its left.

- Search:
  - As the browser narrows and the left pane starts to shrink, the horizontal lines under each category run into the vertical divider instead of maintaining some kind of margin. They also run under the PS.
  - In the widgets, the vertical divider on the right and the > and X end up in a weird place if the widget shrinks such that they can't fit on the same line as the widget title.

- Browse:
  - Make the slider fill the available space when the browser is wide.

- Cart:
  - When the browser hits 700px width, the font size of the secondary nav bar changes slightly

- Detail
  - When the browser is very small, the image takes up the whole space and you can no longer see the actual metadata. The image should shrink further in this case.

juzen2003:
- Update items to be fixed before deployment.
- Change supporting browser size to 600 x 350 pixels.

PR #745, opened 2019-06-04
changed from box-shadow to border and adjusted margins

dstopp:
-changed the css to border 
-added some small margins for space between the images

### DIFF
--- a/opus/application/static_media/css/opus.css
+++ b/opus/application/static_media/css/opus.css
@@ -187,9 +187,10 @@ ul.searchMenu {
     display: inline-block;
     float: left;
     position: relative;
-    margin: .15em;
+    margin: .10em;
     min-width: 100px;
     min-height: 100px;
+    border: medium solid transparent;
 }
 
 .thumbnail-container .img-thumbnail {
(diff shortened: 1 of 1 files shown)


juzen2003 (approved):

rfrenchseti (changes requested):
If you look carefully at the new box layout, you'll see that the thumbnail itself is 94x94, then there's a 3px border on each side (giving a total of 100x100px), but then there's an additional 1.4px (0.1em) margin on each side. That means when we divide the width of the browse by 100 to get the number of items in a row (or similarly for the height), the computation will be off. I suggest changing the code as follows to maintain a nice integer 100x100 thumbnail size:

```
.thumbnail-container {
    display: inline-block;
    float: left;
    position: relative;
    margin: 1px;     // HERE
    min-width: 98px;   // HERE
    min-height: 98px;   // HERE
    border: medium solid transparent;
}
```

rfrenchseti:
Also, there is a bug in calculateGalleryWidth(). The width returned is not the actual width that triggers a thumbnail reflow as the browser changes size. It's off by a few pixels, and I can't figure out where they're coming from. You can easily see this by very slowly changing the browser width, and seeing when the slider updates its step size. The problem can be fixed by adding this line to the bottom of calculateGalleryWidth:

```
        width -= (tab === "#cart" ? 4 : 6);
        return width;
```

PR #765, opened 2019-06-08
Various bug fixes

rfrenchseti:
- I broke parts of the metadata selector when I changed the location of the tooltips in the previous pull. Discard changes didn't work, among other things. I fixed the problems, eliminated redundant code, and did general cleanup.
- I improved tooltips in many places, including giving instructions on how to use ctrl+click, etc.
- There was an obscure bug where if you were sorting on a field that was no longer in your selected metadata (and thus you couldn't get the sort label from the table), when you clicked on a sort pill or table header it would momentarily double-up the text in each sort pill. This was because there are two .sort-container now, one for browse and one for cart, and when we retrieved the label from the existing pill we got two copies. Fixed by just taking the element from #browse, since they will always be identical.
- Unified handling of the help pane and cart overlay pane. This means that ESC will now close whichever one is open. To do this I got rid of the opus.helpPanelOpen and instead store an "active" class on the appropriate #container. Note there are now *two* active classes - one on the op-overlay div for the CSS to enable its viewing, and one on the div for the appropriate pane. The latter is so we can tell which pane is open to get the slide direction right. I could've added a second global variable instead but this seemed cleaner.
- I fixed an obscure bug on the detail page where text wrap wasn't as pretty as it could be for very narrow browsers. The metadata now wraps like the filenames in the product type list do.
- I renamed various classes and did other cosmetic cleanup. There's much more than needs to be done.

### DIFF
--- a/opus/application/apps/results/templates/results/detail_metadata_internal.html
+++ b/opus/application/apps/results/templates/results/detail_metadata_internal.html
@@ -1,18 +1,20 @@
 {% for cat, d in data.items %}
-    <ul id="detail__data_{{ cat.lower.split|join:"_" }}" class="op-detail-category-metadata">
+    <ul id="detail__data_{{ cat.lower.split|join:"_" }}" class="op-detail-list">
         {% for param, val in d.items %}
-            <li>
+            <li class="op-detail-entry">
                 {% for name, param_info in all_info.items %}
                     {% if name == param %}
                         {% if param_info.get_tooltip_results %}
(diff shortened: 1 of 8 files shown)


dstopp:
Only comment I had is likely related to an issue w/focus; when you close the cart download options panel w/the ESC key, there is a border around the button that is not there when you use the X.  To be dealt with in separate bug fix.

dstopp (approved):

PR #771, opened 2019-06-10
758-disable slider when the browse results are empty

dstopp:
-remove spinner from slider and set html to "?"
-fix the width of the slider to the size of the "?" when empty
-disable the slider when browse results empty
-fix the bug where sometimes the cart message would appear
-change the slider number to a class instead of id for future use on the cart


rfrenchseti:
How to reproduce the "You have no results" message appearing after a gallery containing actual results:
- Start with a fresh OPUS
- Select something like Earth to give zero results
- Go to the Browse tab
- Go to the Search tab
- Select something like Mars to give a small but not zero number of results
- Go to the Browse tab
- You can scroll down below the gallery and see the "you have no results" message

rfrenchseti (changes requested):
As long as you're fixing problems with "empty results" showing up when it shouldn't, do you want to go ahead and fix the "browse" version above?

rfrenchseti:
OK found another problem - when there are no results, the gallery still has a scrollbar that lets you scroll quite a lot. Every time you scroll it does an API call. You can actually scroll the "no results" message off the top of the screen. Oddly enough, this doesn't happen in the table view. There is no scrollbar there. But there's also no message saying there are no results. It seems to me the two should be consistent - both say there are no results, and neither have a scrollbar.

rfrenchseti:
- Start fresh
- Go to browse tab
- Select table view
- Go to search tab
- Select something with zero results
- Go to browse tab
- Go to search tab
- Select something with some results
- Go to browse tab
- The browse tab will show gallery thumbnails, even though we were previously in table mode. Also the navbar item says "View Gallery" even though we're looking at a gallery. Note that browse=data in the URL.

rfrenchseti:
Actually it's worse than that. Go to browse with lots of results. View the gallery. Then view table. Drag the slider a long distance. You will see the thumbnails but you're still in table mode.

rfrenchseti:
- I just got the cart page to be empty when I switched to it after adding observations. Not sure how I did it.
- Scrolling in the cart is broken. Put a lot of observations in. Go to the cart tab. Use the mouse wheel to scroll down as fast as you can. The scrolling will get stuck and not continue to infinite scroll/load new pages.

rfrenchseti (approved):

PR #772, opened 2019-06-10
S108 move slider on collapse search menu

dstopp:
-Move the scrollbar if the opened menu goes below viewable area
-Don't move the scrollbar if the menu is closed
-Note that if the scrollbar position is adjusted on close and there would have been white space, there appears to be a bit of a 'jump' because the scrollbar adjusts the entire page to remove any whitespace at the bottom of the menu. Not sure if that is ok.


rfrenchseti:
The behavior when opening a menu appears to be correct. The jump when closing a menu is no worse than it is now, although it's a little annoying. I'm not sure there's any easy way to fix it, though, since there is no whitespace in the DOM at the bottom of the menu for the PS to use.

opus/application/static_media/js/mutationObserver.js
```diff
@@ -69,7 +69,21 @@ var o_mutationObserver = {
                     if (idx === lastMutationIdx) {
                         if (mutation.target.classList.value.match(/collapse/)) {
                             // If there is a collapse/expand happened (attribute changes), we only update ps at the last mutation when the animation finishes and class/style are finalized
-                            adjustSearchSideBarHeight();
+                            o_search.adjustSearchSideBarHeight();
```
juzen2003, line 72: Just a note, we can call the the function directly like what Debby did here if we only call it once at the last mutation (inside if (idx === lastMutationIdx) block) in mutationObserver. The debounced version is used when the function is called at multiple mutations in mutationObserver.

opus/application/static_media/js/mutationObserver.js
```diff
@@ -69,7 +69,21 @@ var o_mutationObserver = {
...
+
+                            // if the collapse opens below the viewable area, move the scrollbar
+                            // only move the scrollbar if we open the menu item
+                            let lastElement = $(mutation.target).children().last();
+                            if (mutation.target.classList.value.match(/show/) &&
+                               !lastElement.isOnScreen("#sidebar-container")) {
```
juzen2003: Just minor styling issue, please add one more space before "!" so the if condition expression is aligned in inside the outer "()", like what we did in other places:
```js
if (mutation.target.classList.value.match(/show/) &&
    !lastElement.isOnScreen("#sidebar-container")) {
```

dstopp: done

opus/application/static_media/js/mutationObserver.js
dstopp: regarding the jump when closing a menu, i wonder if we could find a smooth animation that would gently scroll vs. jump... just a thought.

rfrenchseti (approved):

Issue #762, opened 2019-06-08
Infinite scroll in the table view is spastic


rfrenchseti:
- In the table view, when you scroll the slider does not move by the gallery step size, but instead moves by units of 1
- When cache filling is required, the table jumps a long distance (so does the slider)
- For awhile we had the behavior that the slider would *always* step by the gallery row size. I believe this is the desired behavior, even though it looks weird in table view. I don't know what broke this.

PR #777, opened 2019-06-11
Fix infiniteScroll in tableView 

juzen2003:
- This is related to #762 
  - When scrolling slider in table view, the slider has the step size equal to gallery row size.
  - When scrolling ps in table view, the slider will be updated base on gallery row size.
  - When switching between browse tab and other tabs, the slider will stay at the same value.
  - When we move the slider to the end (max value), scroll down the ps to the loadNext threshold point (near the bottom of ps), and switching between gallery and table views, the slider value will stay the same.

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -729,10 +729,8 @@ var o_browse = {
             // Update obsNum in both infiniteScroll instances.
             // Store the most top left obsNum in gallery for both infiniteScroll instances
             // (this will be used to updated slider obsNum).
-            if (o_browse.isGalleryView()) {
-                obsNum = (o_utils.floor((obsNum - 1)/galleryBoundingRect.x) *
-                          galleryBoundingRect.x + 1);
-            }
+            obsNum = (o_utils.floor((obsNum - 1)/galleryBoundingRect.x) *
+                      galleryBoundingRect.x + 1);
 
             if (maxSliderVal >= obsNum) {
                 $(`${tab} .op-gallery-view`).infiniteScroll({"obsNum": obsNum});
@@ -1504,7 +1502,7 @@ var o_browse = {
 
                         // Update the obsNum in infiniteScroll instances with the first obsNum of the row above current last page
                         // This will be used to set the scrollbar position later
-                        if (infiniteScrollData) {
+                        if (infiniteScrollData && obsNum <= viewNamespace.totalObsCount) {
                             $(`${tab} .op-gallery-view`).infiniteScroll({"obsNum": scrollbarObsNum});
                             $(`${tab} .op-data-table-view`).infiniteScroll({"obsNum": scrollbarObsNum});
                             opus.prefs[startObsLabel] = scrollbarObsNum;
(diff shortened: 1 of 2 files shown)


opus/application/static_media/js/browse.js
```diff
@@ -674,6 +674,11 @@ var o_browse = {
 
     // find the first displayed observation index & id in the upper left corner
     updateSliderHandle: function(browserResized=false) {
+        // Only update the slider & obSnum in infiniteScroll instances when the user
```
rfrenchseti, line 683: What is the point of this new function? Doesn't it do exactly the same thing as getViewTab below?

rfrenchseti (changes requested):
The scrollbar still doesn't work well in the table view. Sometimes when you scroll down, the new page loads, and the scrollbar jumps back to the middle, the slider also jumps (as well as the top row of the table jumping back). You can probably reproduce this by sliding to a place so you clear the cache, then slowly pressing the down arrow to move one row at a time. When you get to the place where the scrollbar resets to the middle, the slider will move back substantially as well.

I've also seen cases where the top of the table is several gallery rows off from what the slider says.

opus/application/static_media/js/browse.js
juzen2003: This is directly reading from opus.prefs.view and it will have possible returned value equal to "search", "browse", "cart", and "detail". It will also let us know where users are at right now, for example, it will return "search" if the user is at search page. getViewTab only returns "#cart" or "#browse", and if the user is at search page, the function still returns "#browse".  The reason of using this function is to make sure updateSliderHandle only gets called when we select browse tab (or we are at browse tab). This is to avoid updateSliderHandle being called from adjustBrowseHeight when mutationObserver detects a tab changes. For example, when we are in browse view and click search nav-link, mutationObserver will detect a change and call adjustBrowseHeight even though we are getting out of browse view. By adding the check here, updateSliderHandle will not be called when we are exiting out the browse tab, so the startObs stored in the infiniteScroll instances and startobs in url will not get messed up (because the calculation in updateSliderHandle is based on the existing elements in browse tab, variables like topBoxBoundary will get messed up in above example).

juzen2003:
- In table view, when infiniteScroll load event is triggered, the scrollbar and slider location will be at the correction places without jumping backward.

opus/application/static_media/js/browse.js
rfrenchseti: Ah, I see. I think we might want to rename "getViewTab" to "getViewBrowseCartTab" or something to avoid confusion.

rfrenchseti:
Table scrolling still doesn't work right. The slider doesn't jump any more, but the actual table contents still does. Do the same thing as before - use the down arrow to move one table row at a time. Wait until infinite scroll load is triggered. The table will now be back on a gallery row boundary even though it wasn't before. For example, if the row size is 10, you get sliders at 1, 11, 21, 31, 41, 51, 61. Scroll down so the top left element is, say 55, and moving to 56 will trigger the load. Go to 56, but now the table top will be back at 51.

juzen2003:
- In table view, when infiniteScroll load is triggered, the table top will not jump back to the obsNum that matches the slider number. It will now stay at the same obsNum that trigger the infiniteScroll load.

rfrenchseti (approved):

Issue #761, opened 2019-06-08
Infinite scroll realigns rows after filling cache


rfrenchseti:
Use the up or down arrows to slowly scroll the browse page. The page will scroll normally, and the top row of icons may be partially above the screen. The slider value will update based on the center point of the top row. This is all fine.

Now use the arrows to get to the point where a cache fill is needed. The above behavior will happen, then the loading spinner will come on, and then when loading is finished the browse window will realign itself so that the top row is even with the top of the browser. This causes a jump in the thumbnails and also possibly a jump in the slider position, both of which are disorienting to the user and unnecessary.

PR #786, opened 2019-06-12
Metadata selector fixes and other general cleanup

### LINKED ISSUE #767: Select Metadata should scroll to see field just added
When adding fields with Select Metadata, if there is a scrollbar on the right side the new fields might not be visible. We should scroll the right side so that the bottom field is visible.

### LINKED ISSUE #783: Changing metadata selection resets startobs to 1
When you update the selected metadata via the Select Metadata dialog and "Save Changes", the startobs resets to 1 instead of staying where it was. It has no reason to reset, and we already fixed this once.

rfrenchseti:
- Fixed duplicate IDs on left search pane and metadata selector that was causing things the user did in one place to affect the other. Now IDs have a prefix.
- Fixed a bug where when new categories were added or removed from the search tab, the metadata selector wasn't updated.
- In the metadata selector, when a category is expanded scroll so that the new category is visible just like we do on the search tab. Fixes #767.
- In the metadata selector, when a new item it added to the right pane, make sure it is visible.
- When a metadata column is added, make the (i) appear in the correct place relative to ones that were loaded through menu.html.
- When removing a metadata column by clicking on the left pane, the right <li> would fade out, but if you clicked on the trashcan it would just disappear. Now it fades out in both cases.
- Fixed improper font naming of fa-info-circle.
- Recent changes to consolidate code broke what happens when the metadata selector exits with "Save". Now it keeps the slider and the scroll position as it was, rather than resetting back to obs 1. For the table this is basically seamless. For the gallery it will realign the top row with the top of the browser (see issue #761). Fixes #783.
- Make sure the bottom of a recently exposed category is *entirely* on the screen, not just half of it.
- When you click on a table header or sort pill to change the sort order, the table headers compress and then expand (there's no space between the table header label and the sort arrow, then there is). This was caused by a missing class. I've added it and now sorting the table is much smoother.
- General code cleanup.

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -539,7 +539,7 @@ var o_browse = {
             // let headers = $(`.op-data-table-view th a:not([data-slug="opusid"], [data-slug=${slug}])`).find("span:last");
             let headers = $(`.op-data-table-view th a:not([data-slug=${slug}])`).find("span:last");
             headers.data("sort", "none");
-            headers.attr("class", defaultTableSortArrow);
+            headers.attr("class", `column_ordering ${defaultTableSortArrow}`);
         }
 
         // Re-render each pill
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -957,7 +957,7 @@ var o_browse = {
             // update the data table w/the new columns
             if (!o_utils.areObjectsEqual(opus.prefs.cols, currentSelectedMetadata)) {
                 let tab = opus.getViewTab();
-                o_browse.clearObservationData();
+                o_browse.clearObservationData(true); // Leave startobs alone
                 o_hash.updateHash(); // This makes the changes visible to the user
                 o_browse.initTable(tab, opus.colLabels, opus.colLabelsNoUnits);
                 o_browse.loadData(opus.prefs.view);
@@ -1956,9 +1956,15 @@ var o_browse = {
         o_browse.metadataboxHtml(opusId);
     },
 
-    clearObservationData: function() {
-        opus.prefs.startobs = 1; // reset startobs to 1 when data is flushed
-        opus.prefs.cart_startobs = 1;
+    clearObservationData: function(leaveStartObs) {
+        if (!leaveStartObs) {
+            // Normally when we delete all the data we want to force a return to the top because
+            // we don't know what data is going to be loaded. But in some circumstances
+            // (like updating metadata columns) we know we're going to reload exactly the same
+            // data so we can keep our place.
+            opus.prefs.startobs = 1; // reset startobs to 1 when data is flushed
+            opus.prefs.cart_startobs = 1;
+        }
         o_cart.reloadObservationData = true;  // forces redraw of cart tab
         o_cart.observationData = {};
         o_browse.reloadObservationData = true;  // forces redraw of browse tab
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -923,7 +923,7 @@ var o_browse = {
 
         let label = $(menuSelector).data("qualifiedlabel");
         let info = `<i class="fas fa-info-circle" title="${$(menuSelector).find('*[title]').attr("title")}"></i>`;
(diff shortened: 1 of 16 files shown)


opus/application/static_media/js/mutationObserver.js
```diff
@@ -37,9 +37,9 @@ var o_mutationObserver = {
             subtree: true,
         };
 
-        let adjustSearchSideBarHeight = _.debounce(o_search.adjustSearchSideBarHeight, 200);
-        let adjustSearchWidgetHeight = _.debounce(o_search.adjustSearchHeight, 200);
-        let adjustSearchHeight = _.debounce(o_search.adjustSearchHeight, 200);
+        let searchSideBarHeightChanged = _.debounce(o_search.searchSideBarHeightChanged, 200);
+        let searchWidgetHeightChanged = _.debounce(o_search.searchHeightChanged, 200);
```
juzen2003: This is my bad, it should be a debounced version of original adjustSearchWidgetHeight.
But it's still working fine without issues because adjustSearchWidgetHeight is called inside adjustSearchHeight.
To be more cleaner, it should be like this:
``` js
let searchWidgetHeightChanged = _.debounce(o_search.searchWidgetHeightChanged, 200);
```

juzen2003 (commented):
- In tool 8080 server, there is a little white space above the footer in browse tab. But It doesn't exist when running at localhost.
- This is just a note for me, in detail page, when the screen is narrow and image is on top of the left pane, ps can scroll all the way down and get out of the browser window. This issue is fixed in previous responsive branch before merging. I'll look into this later.

opus/application/static_media/js/mutationObserver.js
rfrenchseti: Fixed

rfrenchseti:
This also fixed #788

juzen2003 (approved):

opus/application/static_media/js/browse.js
```diff
@@ -1947,12 +1956,29 @@ var o_browse = {
...
+            opus.prefs.startobs = 1; // reset startobs to 1 when data is flushed
+            opus.prefs.cart_startobs = 1;
+        }
         o_cart.reloadObservationData = true;  // forces redraw of cart tab
         o_cart.observationData = {};
         o_browse.reloadObservationData = true;  // forces redraw of browse tab
         o_browse.observationData = {};
     },
+
+    clearBrowseObservationDataAndEraseDOM: function(leaveStartObs=false) {
```
dstopp, line 1974: why don't you need to do this on the cart for similar reasons as well....?

opus/application/static_media/js/mutationObserver.js
```diff
@@ -140,18 +139,35 @@ var o_mutationObserver = {
...
+                        // If a collapse/expand happened (attribute changes),
+                        // we only update ps at the last mutation when the animation
+                        // finishes and class/style are finalized.
+                        // Note we call the original version here, not the debounced
+                        // version, because we need the PS to be visible instantly in
+                        // order for scrollTop to work below.
+                        o_browse.adjustMetadataSelectorMenuPS();
+                        // if the collapse opens below the viewable area, move the scrollbar
+                        // only move the scrollbar if we open the menu item
+                        let lastElement = $(mutation.target).children().last();
```
dstopp, line 160: line 186 et all are exactly the same - function?

dstopp (changes requested):

opus/application/static_media/js/browse.js
rfrenchseti: clearBrowseObservationDataAndEraseDOM is called from the main OPUS UI when the user changes the search parameters. It makes sure that a data load will be triggered in the browse tab, even if the user switches tabs at an inopportune time (like we didn't actually send out for a result_count yet). But the cart isn't affect by the user's search, so there's no point in clearing it. Since there are now two separate caches for data, there's no overlap between the tabs.

opus/application/static_media/js/mutationObserver.js
rfrenchseti: Actually there are quite a few differences - one is doing things to the left side pane (with collapsing categories) and the other is doing things to the right pane (list of selected metadata). Although the code looks similar from a distance, almost every line is different. I looked to see if it would be possible to refactor, but I think it would just make the code more complicated.

PR #806, opened 2019-06-18
Fix infinite scroll realigns rows issue #761

juzen2003:
- This is related to issue #761 
- Local branch: infiniteScroll_issue_761
- Now the scrollbar position will not be re-aligned after infiniteScroll load is triggered.

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -609,7 +609,7 @@ var o_browse = {
         }
     },
 
-    setScrollbarPosition: function(galleryObsNum, tableObsNum, view) {
+    setScrollbarPosition: function(galleryObsNum, tableObsNum, view, offset=0) {
         let tab = opus.getViewTab(view);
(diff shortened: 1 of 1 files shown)


opus/application/static_media/js/browse.js
```diff
@@ -749,17 +754,37 @@ var o_browse = {
...
+            }
+
             if (maxSliderVal >= obsNum) {
                 // "sliderObsNum" will be the startObs.
                 // "scrollbarObsNum" will be the obsNum at current scrollbar location.
                 // In table view, it's the top obsNum. In gallery view, it's one of obsNums in the top row.
                 $(`${tab} .op-gallery-view`).infiniteScroll({
                     "sliderObsNum": obsNum,
-                    "scrollbarObsNum": tableCurrentObsNum
+                    "scrollbarObsNum": tableCurrentObsNum,
```
rfrenchseti, line 781: I'm not sure I understand this code, but why it scrollbarObsNum being set to tableCurrentObsNum here instead of obsNum? I would think we would want the gallery scrollbar to always be on a row boundary.

rfrenchseti (changes requested):

Issue #807, opened 2019-06-18
Switching between table and gallery view moves the slider


rfrenchseti:
- Open a fresh OPUS
- Go to the browse tab
- Move the slider to somewhere random in the middle
- Go to the table view
- The slider will have moved 1 row earlier
- Switch back to gallery
- Switch back to table
- The slider will have moved another 1 row earlier

PR #809, opened 2019-06-19
Slider Jumping issue #807

juzen2003:
- This is related to #807 
- Local branch: slider_issue_807
- The reason for slider jumping:
  - In updateSliderHandle, we are calculating the startObs and table row height is used in the calculation ($(`${tab} tbody tr`).outerHeight()). However when issue happened, the first row height will be 1px (coming from a 1px top border, and it's not in css file. This only happened in my firefox and I didn't see it in my chrome) larger than rest of row height, and this will messed up the calculation. That's why slider always jump to one row ahead when switching from gallery to table view.  Now we usd the 2nd row height (same as rest of table tr's height), please let me know if we prefer to do something like: ``` Math.min($(`${tab} tbody tr`).outerHeight(), $(`${tab} tbody tr`).eq(1).outerHeight())```
- Now we break down steps in updateSliderHandle into multiple functions and there are comments.

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -711,10 +711,14 @@ var o_browse = {
 
             // table: obsNum = calculatedFirstObs + number of row
             // gallery: obsNum = calculatedFirstObs + number of row * number of obs in a row
+            // Note: in table view, we divide by 2nd table tr's height because in some corner
+            // cases, the first table tr's height will be 1px larger than rest of tr, and this
+            // will mess up the calculation.
             let obsNumDiff = (o_browse.isGalleryView() ?
                               Math.round((topBoxBoundary - firstCachedObsTop)/o_browse.imageSize) *
                               galleryBoundingRect.x :
-                              Math.round((topBoxBoundary - firstCachedObsTop)/$(`${tab} tbody tr`).outerHeight()));
+                              Math.round((topBoxBoundary - firstCachedObsTop)/
+                              $(`${tab} tbody tr`).eq(1).outerHeight()));
 
             let obsNum = obsNumDiff + calculatedFirstObs;
             if (browserResized) {
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -696,94 +696,19 @@ var o_browse = {
             let viewNamespace = opus.getViewNamespace();
             let galleryBoundingRect = viewNamespace.galleryBoundingRect;
 
-            // this will get the top left obsNum for gallery view or the top obsNum for table view
-            let firstCachedObs = $(selector).first().data("obs");
-            let firstCachedObsTop = $(selector).first().offset().top;
-            let calculatedFirstObs = (o_utils.floor((firstCachedObs - 1)/galleryBoundingRect.x) *
-                                      galleryBoundingRect.x + 1);
-            // For gallery view, the topBoxBoundary is the top of .gallery-contents
-
-            // For table view, we will set the topBoxBoundary to be the bottom of thead
-            // (account for height of thead)
-            let topBoxBoundary = (o_browse.isGalleryView() ?
-                              $(`${tab} .gallery-contents`).offset().top :
-                              $(`${tab} .gallery-contents`).offset().top + $(`${tab} .op-data-table thead th`).outerHeight());
-
-            // table: obsNum = calculatedFirstObs + number of row
-            // gallery: obsNum = calculatedFirstObs + number of row * number of obs in a row
-            // Note: in table view, we divide by 2nd table tr's height because in some corner
-            // cases, the first table tr's height will be 1px larger than rest of tr, and this
-            // will mess up the calculation.
-            let obsNumDiff = (o_browse.isGalleryView() ?
-                              Math.round((topBoxBoundary - firstCachedObsTop)/o_browse.imageSize) *
-                              galleryBoundingRect.x :
-                              Math.round((topBoxBoundary - firstCachedObsTop)/
-                              $(`${tab} tbody tr`).eq(1).outerHeight()));
-
-            let obsNum = obsNumDiff + calculatedFirstObs;
-            if (browserResized) {
-                // At this point of time, galleryBoundingRect is updated with new row size
-                // from countGalleryImages in adjustBrowseHeight.
-                let numToDelete = ((galleryBoundingRect.x - (firstCachedObs - 1) % galleryBoundingRect.x) %
-                                   galleryBoundingRect.x);
-
-                let galleryObsElem = $(`${tab} .gallery [data-obs]`);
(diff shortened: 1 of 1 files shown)


opus/application/static_media/js/browse.js
```diff
@@ -696,90 +696,19 @@ var o_browse = {
...
-            $(`${tab} .op-slider-nav`).removeClass("op-button-disabled");
-            // update startobs in url when scrolling
-            o_hash.updateHash(true);
+            // When we update the slider, the following steps are run:
+            // 1. Get the current startObs and top obsNum (top item in table view and one of
+            //    top row items in gallery view) that scrollbar is at.
+            // 2. Get the max value for slider.
+            // 3. Update the slider value and data in infiniteScroll instances and URL using
+            //    the data obtained from above 2 steps.
+            let [obsNum, currentScrollObsNum] = o_browse.getCurrentStartObs(selector, tab,
```
rfrenchseti, line 705: I think getCurrentStartObs is a confusing name, because it does so much more (like delete items from the DOM). How about something like realignDOMAndGetStartObsAndObsNum.

opus/application/static_media/js/browse.js
```diff
@@ -796,6 +725,134 @@ var o_browse = {
...
+        /**
+         * get the current startObs (for slider number) and the top item's
+         * obsNum in table view, this item will also be at the top row in
+         * gallery view (for setting scrollbar location)
+         */
+
+        // this will get the top left obsNum for gallery view or the top obsNum for table view
+        let firstCachedObs = $(selector).first().data("obs");
+        let firstCachedObsTop = $(selector).first().offset().top;
+        let calculatedFirstObs = (o_utils.floor((firstCachedObs - 1)/galleryBoundingRect.x) *
```
rfrenchseti, line 738: This would make more sense called something like "alignedCachedFirstObs".

opus/application/static_media/js/browse.js
```diff
@@ -796,6 +725,134 @@ var o_browse = {
...
+                                  galleryBoundingRect.x + 1);
+        // For gallery view, the topBoxBoundary is the top of .gallery-contents
```
rfrenchseti, line 740: This comment should be swapped with the blank line below it

opus/application/static_media/js/browse.js
```diff
@@ -796,6 +725,134 @@ var o_browse = {
...
+                          galleryBoundingRect.x :
+                          Math.round((topBoxBoundary - firstCachedObsTop)/
+                          tableRowHeight));
+
+        let obsNum = obsNumDiff + calculatedFirstObs;
+
+        o_browse.deleteCachedObs(browserResized, tab, viewNamespace, galleryBoundingRect,
+                                 firstCachedObs);
+
+        // Update obsNum in both infiniteScroll instances.
```
rfrenchseti, line 765: This comment doesn't make sense anymore, since we aren't actually updating the infiniteScroll instances here. Instead the comment should say something about what obsNum is and why it's calculated that way.

opus/application/static_media/js/browse.js
```diff
@@ -796,6 +725,134 @@ var o_browse = {
...
+        // (this wll be used to updated the table view scrollbar location after initInfiniteScroll
+        //  load is triggered).
+        let currentScrollObsNum = obsNum;
+        obsNum = (o_utils.floor((obsNum - 1)/galleryBoundingRect.x) *
+                  galleryBoundingRect.x + 1);
+
+        return [obsNum, currentScrollObsNum];
+    },
+
+    deleteCachedObs: function(browserResized, tab, viewNamespace ,galleryBoundingRect, firstCachedObs) {
```
rfrenchseti, line 778: There's really no reason to pass in browserResized here, since the function does nothing if it wasn't resized. Instead, place this browserResize test back in updateSliderHandle and only call deleteCachedObs if it's true.

opus/application/static_media/js/browse.js
```diff
@@ -796,6 +725,134 @@ var o_browse = {
...
+        let topBoxBoundary = (o_browse.isGalleryView() ?
+                          $(`${tab} .gallery-contents`).offset().top :
+                          $(`${tab} .gallery-contents`).offset().top + $(`${tab} .op-data-table thead th`).outerHeight());
+
+        // table: obsNum = calculatedFirstObs + number of row
+        // gallery: obsNum = calculatedFirstObs + number of row * number of obs in a row
+        // Note: in table view, we divide by the 2nd table tr's height because in some corner
+        // cases, the first table tr's height will be 1px larger than rest of tr, and this
+        // will mess up the calculation.
+        let tableRowHeight = $(`${tab} tbody tr`).eq(1).outerHeight();
```
rfrenchseti, line 753: So what happens here if there's only one row in the table and we ask for info on the second row? OPUS continues to work, probably because the slider ends up being disabled anyway, but I'm guessing it returns "Undefined" here and then the math below returns something bogus.

opus/application/static_media/js/browse.js
```diff
@@ -796,6 +725,134 @@ var o_browse = {
...
+        let obsNumDiff = (o_browse.isGalleryView() ?
+                          Math.round((topBoxBoundary - firstCachedObsTop)/o_browse.imageSize) *
+                          galleryBoundingRect.x :
+                          Math.round((topBoxBoundary - firstCachedObsTop)/
+                          tableRowHeight));
+
+        let obsNum = obsNumDiff + calculatedFirstObs;
+
+        o_browse.deleteCachedObs(browserResized, tab, viewNamespace, galleryBoundingRect,
```
rfrenchseti, line 762: I find it confusing to have deleteCachedObs right here because it's between the computations needed for obsNum and the actual computation of obsNum but it doesn't have anything to do with those computations. Suggest moving it earlier in the function perhaps right after the definition of firstCachedObs.

opus/application/static_media/js/browse.js
```diff
@@ -796,6 +725,134 @@ var o_browse = {
...
+        return [obsNum, currentScrollObsNum];
+    },
+
+    deleteCachedObs: function(browserResized, tab, viewNamespace ,galleryBoundingRect, firstCachedObs) {
+        /**
+         * If browser window is resized, delete some cached observations
+         * to make sure row boundary is correct.
+         */
+
+        if (browserResized) {
```
rfrenchseti, line 784: It would make more sense to put this if outside the function in getCurrentStartObs and don't pass it as an argument since this function does nothing unless it's true.

rfrenchseti (changes requested):
This does seem to fix the same problem I see in Chrome

opus/application/static_media/js/browse.js
rfrenchseti: In an ideal world these two functions would be completely independent and simply called in order here. But I looked at the code and there seems to be too much in common to easily separate them so I think we're stuck for now.

PR #811, opened 2019-06-21
Slider jumping issue (#807) and infinite scroll realigns rows issue #(761)

juzen2003:
- I accindently kill https://github.com/SETI/pds-opus/pull/809 when renaming my remote branch, so I recreate this pull request that combines previous two pull requests (since they are modifying the similar function and area).
- This pull request is related to the following issues: #807 and #761 
- This pull request also combines the following two old pull requests: 
  - https://github.com/SETI/pds-opus/pull/809
  - https://github.com/SETI/pds-opus/pull/806
- local branch: slider_issue_807_and_infiniteScroll_issue_761
- Now the scrollbar position will not be re-aligned after infiniteScroll load is triggered.
- Now when scrolling in table view, switch to gallery view, and go back to table view again, the scrollbar position at table view will still be at the same observation number as previous one if the startObs is not changed in gallery view. For example: (row boundary: 15)
  - The current slider number is 8206 in table view. We scroll down to 8218 as the top item in table view. Switch to gallery view, and switch back to table view again. 8218 will still be the top item in table view because the startObs is not changed in gallery view.
  - The current slider number is 11026 in table view. We scroll down to 11032 as the top item in table view. Switch to gallery view and scroll down so that 11041 is the new startObs. Switch back to table view, the top item will be 11041 because the startObs is changed in gallery view and slider is updated to 11041.

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -787,7 +787,7 @@ var o_browse = {
         let nextObsNum = obsNum + galleryBoundingRect.x;
         let contentsView = o_browse.getScrollContainerClass();
         let previousScrollObsNum = $(`${tab} ${contentsView}`).data("infiniteScroll").options.scrollbarObsNum;
-        let maxSliderVal = o_browse.getSliderMaxValue(viewNamespace, galleryBoundingRect)
+        let maxSliderVal = o_browse.getSliderMaxValue(viewNamespace, galleryBoundingRect);
         obsNum = Math.min(obsNum, maxSliderVal);
         if ((previousScrollObsNum > obsNum && previousScrollObsNum < nextObsNum) ||
             (obsNum === maxSliderVal && previousScrollObsNum > maxSliderVal)) {
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -702,8 +702,12 @@ var o_browse = {
             let galleryBoundingRect = viewNamespace.galleryBoundingRect;
 
             // When we update the slider, the following steps are run:
-            // 1. Get the current startObs and top obsNum (top item in table view and one of
-            //    top row items in gallery view) that scrollbar is at.
+            // 1. Get the current startObs (top left item in gallery view in current page)
+            //    and top obsNum (top item in table view and one of top row items in gallery
+            //    view) that scrollbar is at.
+            //    - If browser is resized, realign doms by deleting some cached obSnum to make
+            //      sure first cached obs are at the correct row boundary. If we don't do so,
+            //      all rendered obs will be with the wrong row boundary after browser resizing.
             // 2. Get the max value for slider.
             // 3. Get the scrollbar offset for both gallery and table view.
             // 4. Update the slider value and data in infiniteScroll instances and URL using
@@ -733,16 +737,24 @@ var o_browse = {
 
     realignDOMAndGetStartObsAndScrollbarObsNum: function(selector, browserResized) {
         /**
-         * get the current startObs (obsNum, for slider number) and the
+         * Get the current startObs (obsNum, for slider number) and the
          * top item's obsNum in table view, this item will also be at the
          * top row in gallery view (currentScrollObsNum, for setting
-         * scrollbar location).
+         * scrollbar location). If browser is resized, realign doms by
+         * deleting some observations from the beginning. This will make
+         * sure the first cached obs is at the correct row boundary
+         * after browser resizing. (which means (first obs - 1) % row size
+         * is equal to 0). If we don't realign doms, first rendered obs
+         * will be at the wrong row boundary, and all rendered obs will
+         * be with the wrong row boundary ((first obs - 1) % row size
+         * will not be 0).
          */
         let tab = opus.getViewTab();
         let viewNamespace = opus.getViewNamespace();
         let galleryBoundingRect = viewNamespace.galleryBoundingRect;
         // this will get the top left obsNum for gallery view or the top obsNum for table view
         let firstCachedObs = $(selector).first().data("obs");
+
         if (browserResized) {
             o_browse.deleteObsToCorrectRowBoundary(tab, firstCachedObs);
         }
@@ -808,7 +820,15 @@ var o_browse = {
     deleteObsToCorrectRowBoundary: function(tab, firstCachedObs) {
         /**
          * If browser window is resized, delete some cached observations
-         * to make sure row boundary is correct.
+         * to make sure row boundary is correct. When browser is resized,
+         * the first cached obs will always be rendered as the first item
+         * of the top row of all rendered obs. If we don't delete some
+         * observations from the beginning, the first cached obs will be
+         * at the wrong row boundary. And all rendered obs will be at the
+         * wrong row boundary ((first obs - 1) % row size will not
+         * be 0). So we delete some observations and that way the first
+         * cached obs is at the correct row boundary after resizing (
+         * (first obs - 1) % row size is equal to 0).
          */
         let viewNamespace = opus.getViewNamespace();
         let galleryBoundingRect = viewNamespace.galleryBoundingRect;
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -711,10 +711,14 @@ var o_browse = {
 
             // table: obsNum = calculatedFirstObs + number of row
             // gallery: obsNum = calculatedFirstObs + number of row * number of obs in a row
+            // Note: in table view, we divide by 2nd table tr's height because in some corner
+            // cases, the first table tr's height will be 1px larger than rest of tr, and this
+            // will mess up the calculation.
             let obsNumDiff = (o_browse.isGalleryView() ?
                               Math.round((topBoxBoundary - firstCachedObsTop)/o_browse.imageSize) *
                               galleryBoundingRect.x :
-                              Math.round((topBoxBoundary - firstCachedObsTop)/$(`${tab} tbody tr`).outerHeight()));
+                              Math.round((topBoxBoundary - firstCachedObsTop)/
+                              $(`${tab} tbody tr`).eq(1).outerHeight()));
 
             let obsNum = obsNumDiff + calculatedFirstObs;
             if (browserResized) {
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -757,6 +757,9 @@ var o_browse = {
 
         if (browserResized) {
             o_browse.deleteObsToCorrectRowBoundary(tab, firstCachedObs);
+            // update firstCachedObs after first couple observations are deleted so that slider
+            // will be upddated to the correct value when browser is resized
+            firstCachedObs = $(selector).first().data("obs");
         }
 
         let firstCachedObsTop = $(selector).first().offset().top;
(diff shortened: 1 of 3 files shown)


opus/application/static_media/js/browse.js
```diff
@@ -796,6 +731,176 @@ var o_browse = {
...
+
+        let firstCachedObsTop = $(selector).first().offset().top;
+        let alignedCachedFirstObs = (o_utils.floor((firstCachedObs - 1)/galleryBoundingRect.x) *
+                                     galleryBoundingRect.x + 1);
+
+        // For gallery view, the topBoxBoundary is the top of .gallery-contents
+        // For table view, we will set the topBoxBoundary to be the bottom of thead
+        // (account for height of thead)
+        let topBoxBoundary = (o_browse.isGalleryView() ?
+                              $(`${tab} .gallery-contents`).offset().top :
```
dstopp: might want to consider creating a variable for the gallery-contents   .top value, as it seems to be used in multiple places

opus/application/static_media/js/browse.js
```diff
@@ -796,6 +731,176 @@ var o_browse = {
...
+        obsNum = Math.min(obsNum, maxSliderVal);
+        if ((previousScrollObsNum > obsNum && previousScrollObsNum < nextObsNum) ||
+            (obsNum === maxSliderVal && previousScrollObsNum > maxSliderVal)) {
+            currentScrollObsNum = o_browse.isGalleryView() ? previousScrollObsNum : currentScrollObsNum;
+        }
+
+        return [obsNum, currentScrollObsNum];
+    },
+
+    deleteCachedObs: function(tab, viewNamespace ,galleryBoundingRect, firstCachedObs) {
```
dstopp: is this different than the function:
    deleteCachedObservation: function(galleryObsElem, tableObsElem, index, viewNamespace) {
the names are so similar...

opus/application/static_media/js/browse.js
```diff
@@ -696,90 +701,20 @@ var o_browse = {
...
+            // 2. Get the max value for slider.
+            // 3. Get the scrollbar offset for both gallery and table view.
+            // 4. Update the slider value and data in infiniteScroll instances and URL using
+            //    the data obtained from above 2 steps.
+            let [obsNum, currentScrollObsNum] = o_browse.realignDOMAndGetStartObsAndScrollbarObsNum(selector,
+                                                tab, viewNamespace, galleryBoundingRect, browserResized);
+            let maxSliderVal = o_browse.getSliderMaxValue(viewNamespace, galleryBoundingRect);
+            let [galleryOffset, tableOffset] = o_browse.getScrollbarOffset(tab, obsNum, currentScrollObsNum);
+
+            o_browse.updateSliderValue(tab, galleryBoundingRect, startObsLabel, maxSliderVal,
```
dstopp: was curious why you made this a function when it is only used once and it takes 7 input parameters.  I can see that it shortens the original function, is that the reason?  I'm on the fence about that sort of function creation.

opus/application/static_media/js/browse.js
```diff
@@ -796,6 +731,176 @@ var o_browse = {
...
+         * Get galleryOffset and tableOffset, these values will be used
+         * in the calculation in setScrollbarPosition and provide smooth
+         * scrolling when infiniteScroll load event is triggered.
+         */
+        let galleryTarget = $(`${tab} .op-thumbnail-container[data-obs="${obsNum}"]`);
+        let tableTarget = $(`${tab} .op-data-table tbody tr[data-obs='${currentScrollObsNum}']`);
+        let galleryOffset = 0;
+        let tableOffset = 0;
+        if (galleryTarget.length && tableTarget.length) {
+           let galleryTargetTopPosition = galleryTarget.offset().top;
```
dstopp: note that the tab space here is 3 :)  just an observation...

opus/application/static_media/js/browse.js
```diff
@@ -1295,7 +1401,7 @@ var o_browse = {
         //   it for the case when cached data is removed so that scrollbar position is always correct (and never reaches to the
         //   end until it reaches to the end of the data)
         o_browse.setScrollbarPosition(infiniteScrollData.options.sliderObsNum,
```
dstopp: I think you can simplify the code a little bit here if you make infiniteScrollData.options a variable.

opus/application/static_media/js/browse.js
```diff
@@ -1541,12 +1647,10 @@ var o_browse = {
...
-                        // startObs values right before infiniteScroll load event is triggered. This will be used
-                        // to properly set the scrollbar and slider location in table view after new data is loaded.
-                        let scrollbarObsNum = (o_browse.isGalleryView() ? (obsNum - o_browse.getLimit(view) -
-                                               opus.getViewNamespace(view).galleryBoundingRect.x) :
-                                               opus.prefs[startObsLabel]);
+                        // We will update the obsNum in InfiniteScroll instances with the startObs values
+                        // right before infiniteScroll load event is triggered. This will be used to properly
+                        // set the scrollbar and slider location after new data is loaded.
+                        let scrollbarObsNum = opus.prefs[startObsLabel];
                         scrollbarObsNum = Math.max(scrollbarObsNum, 1);
```
dstopp: can scrollbarObsNum ever be less than 1?

dstopp (changes requested):

opus/application/static_media/js/browse.js
```diff
@@ -696,90 +701,20 @@ var o_browse = {
...
-                $("#op-observation-slider").slider({
-                    "value": obsNum,
-                    "step": o_browse.gallerySliderStep,
-                    "max": maxSliderVal,
-                });
-            }
-            $(`${tab} .op-slider-nav`).removeClass("op-button-disabled");
-            // update startobs in url when scrolling
-            o_hash.updateHash(true);
+            // When we update the slider, the following steps are run:
```
dstopp, line 705: I think renaming the variable "numThumbnails" to numObservations <or similar> would help the code at line 699 be clearer because 'selector' may refer to the number of thumbnails on the gallery page OR the number of observations in the viewable part of the table,

opus/application/static_media/js/browse.js
rfrenchseti: Agreed. This function needs a better name.

opus/application/static_media/js/browse.js
rfrenchseti: I haven't looked at the code carefully here, but I think Debby thought that most of these passed parameters could actually be computed inside the called function since they were only used for that one purpose. That would reduce the number of passed arguments and make it look cleaner.

dstopp: I think that the only thing you should actually pass in is obsNum , currentScrollObsNum, galleryOffset & tableOffset.  The var maxSliderVale is only used by this function and can be retrieved in the function; the others can just be retrieved inside the function as it is not time consuming or cycle-intensive to just get those values again. 
In addition, getScrollbarOffset should return an object vs. a list of parameters <see interesting discussion in slack>. This will avoid potential future issues w/parameter ordering, etc.

opus/application/static_media/js/browse.js
juzen2003: Because scrollbarObsNum is the top item in table view (or one of the top row items in gallery view), the minimum number will be 1.

juzen2003:
- Update based on comments

opus/application/static_media/js/browse.js
rfrenchseti: I think Debby's point was that if opus.prefs[startObsLabel] can never be less than 1, then there's no point in doing the max() and it can be removed.

opus/application/static_media/js/browse.js
rfrenchseti: I think this has not been resolved.

opus/application/static_media/js/browse.js
```diff
@@ -691,99 +696,30 @@ var o_browse = {
...
+            // 1. Get the current startObs and top obsNum (top item in table view and one of
+            //    top row items in gallery view) that scrollbar is at.
```
rfrenchseti: Need to add a comment here about getting the current startObs etc. also doing a DOM realignment, and why that's necessary

opus/application/static_media/js/browse.js
```diff
@@ -691,99 +696,30 @@ var o_browse = {
...
+            // 3. Get the scrollbar offset for both gallery and table view.
+            // 4. Update the slider value and data in infiniteScroll instances and URL using
+            //    the data obtained from above 2 steps.
+            let obsNumObj = o_browse.realignDOMAndGetStartObsAndScrollbarObsNum(selector, browserResized);
+            let obsNum = obsNumObj.startObs;
+            let currentScrollObsNum = obsNumObj.scrollbarObsNum;
+            let maxSliderVal = o_browse.getSliderMaxValue();
+            let scrollbarOffset = o_browse.getScrollbarOffset(obsNum, currentScrollObsNum);
+            let galleryOffset = scrollbarOffset.galleryOffset;
+            let tableOffset = scrollbarOffset.tableOffset;
```
rfrenchseti: I don't think there's really much point to having all these extra "let" statements here when the variable is only used once. Just put obsNumObj.scrollbarObsNum or whatever directly in the function call argument list.

opus/application/static_media/js/browse.js
```diff
@@ -796,6 +732,187 @@ var o_browse = {
         }
     },
 
+    realignDOMAndGetStartObsAndScrollbarObsNum: function(selector, browserResized) {
+        /**
```
rfrenchseti, line 746: Add comment about what we're doing to the DOM

opus/application/static_media/js/browse.js
```diff
@@ -796,6 +732,187 @@ var o_browse = {
...
+            currentScrollObsNum = o_browse.isGalleryView() ? previousScrollObsNum : currentScrollObsNum;
+        }
+
+        return {"startObs": obsNum, "scrollbarObsNum": currentScrollObsNum};
+    },
+
+    deleteObsToCorrectRowBoundary: function(tab, firstCachedObs) {
+        /**
+         * If browser window is resized, delete some cached observations
+         * to make sure row boundary is correct.
```
rfrenchseti: This is such a weird thing to do that I think it would be helpful to have a much more descriptive comment so future developers will understand why we're doing this. Say something about the automatic wrapping around of the thumbnails and how when the browser is resized the first element in the cache won't necessarily be on an appropriate row boundary anymore, and we have to delete obs to make that true, because we always want the cache to be on a row boundary.

rfrenchseti (changes requested):
- Choose a search that only gives you less than one page of results such that the slider is disabled; look at the URL and notice that startobs is negative! Then when you change your search to be larger startobs still stays negative and the slider has a negative value.
- Look at table view. Move the slider somewhere. Note the top line of the table. Resize the browser. The top line of the table will be different. It seems to me like this is unnecessary. Given the control we have over scrollbarObsNum and startObsNum we should be able to put the top row of the table back to the same obs.
- When you resize the browser, the slider jumps to one number, and then quickly to another number.

rfrenchseti:
Another resize problem.
- Start with a relatively small number of observations that will fit in a single browser screen.
- Make the browser smaller so the slider is active.
- Move the slider to the right end.
- Enlarge the browser so that all obs fit on the screen.
- The slide will not become inactive until you try to scroll.

juzen2003:
- Remove min-height css property for #browse .gallery so that there is no extra space for scrolling at the bottom when thumbnail container height is less than 700px.
- Add defensive programming to make sure startObs is never less than 1.
- Slider will be moved to the most left when being disabled.
- When all obs fit on the screen, the slider will always stay inactive.

Issue #798, opened 2019-06-15
Selecting additional search checkboxes causes entire window to reload


rfrenchseti:
On the search tab, if you select a checkbox, and then select a second checkbox for the same widget, the entire window will reload instead of just updating the search.

rfrenchseti:
I can't get this to happen anymore. Will close until it's reproduced.

rfrenchseti:
This is happening regularly for me again. Start with a fresh OPUS. Select Instrument Name. Click on Cassini ISS. Wait. Click on Voyager ISS. The window will reload.

Issue #719, opened 2019-06-03
Safely escape input fields before inserting into URL


rfrenchseti:
A061: The user can enter a raw URL string like "&observationduration1=10" into any range or string field and it gets pasted into the URL verbatim, allowing people to really mess with the URL. These kinds of strings need to be escaped for safety.

juzen2003:
NOTE: In pull request https://github.com/SETI/pds-opus/pull/812, we encode characters like "&" in hash before updating URL or calling API. So whenever the string field, numerical field, or check boxes contain special characters, most characters will be properly encoded before passing into API calls as a search value and Django will handle the decoding at the backend automatically.  "," is the only exception and causes issues. When "," exists in a string or numerical field value, either it's passed into API calls as "%2C" or comma, Django will decode it as ",". And we can't tell if "," is part of the actual value for a slug key or is used to separate multiple values for a slug key. We will have search Django document and see if there is a proper way to control the decoding in Django when revisiting this issue later.

rfrenchseti:
I am deferring this for now because it doesn't matter until we eventually permit the use of multiple values for numeric or string inputs.

rfrenchseti:
This is no longer relevant because we implemented multiple search terms using separate clauses.

PR #812, opened 2019-06-22
Issue related to window reload caused by selecting search checkboxes (#798)

juzen2003:
- This pull request is related to issue #798 
- Local branch: search_reload_issue_798
- Previous window reload issue is caused by the difference between opus.selections and selections obtained from getSelectionsExtrasFromHash. Previously, in opus.selections, if selected value has white space, the white space will be stored as white space. But in selections, the white space is stored as %20 (URL encoding for white space). This difference will cause reload in opus.load function. Now all white space will be stored as %20 so opus.selections and selections are the same and window won't reload.

### DIFF
--- a/opus/application/static_media/js/hash.js
+++ b/opus/application/static_media/js/hash.js
@@ -22,7 +22,7 @@ var o_hash = {
         let hash = [];
         for (let param in opus.selections) {
             if (opus.selections[param].length) {
-                hash.push(param + "=" + opus.selections[param].join(",").replace(" ", "%20"));
+                hash.push(param + "=" + opus.selections[param].join(",").replace(/ /g, "%20"));
             }
         }
 
--- a/opus/application/static_media/js/hash.js
+++ b/opus/application/static_media/js/hash.js
@@ -22,7 +22,7 @@ var o_hash = {
         let hash = [];
         for (let param in opus.selections) {
             if (opus.selections[param].length) {
-                hash.push(param + "=" + opus.selections[param].join(",").replace(/ /g, "%20"));
+                hash.push(param + "=" + opus.selections[param].join(","));
             }
         }
 
--- a/opus/application/static_media/js/hash.js
+++ b/opus/application/static_media/js/hash.js
@@ -159,7 +159,9 @@ var o_hash = {
                     // each qtype will only have one value at a time
                     extras[slug] = [value];
                 } else {
-                    // If value contains ", ", it's one value from string input.
+                    // If value contains ", ", it's one value from string input. This will make sure "," in
+                    // slug value gets encoded in "%2C" in URL, and store as decoded "," in opus internal variables
+                    // selections.
                     if (value.includes(", ")) {
                         selections[slug] = [value];
                     } else {
@@ -237,7 +239,14 @@ var o_hash = {
                     }
                 } else {
                     // these are search params/value!
-                    opus.selections[slug] = value.split(',');
+                    // If value contains ", ", it's one value from string input. This will make sure "," in
+                    // slug value gets encoded in "%2C" in URL, and store as decoded "," in opus internal variables
+                    // opus.selections.
+                    if (value.includes(", ")) {
+                        opus.selections[slug] = [value];
+                    } else {
+                        opus.selections[slug] = value.split(',');
+                    }
                 }
             }
         });
--- a/opus/application/static_media/js/hash.js
+++ b/opus/application/static_media/js/hash.js
@@ -21,18 +21,15 @@ var o_hash = {
         let hash = [];
         for (let param in opus.selections) {
             if (opus.selections[param].length) {
-                let encodedSelectionValues = o_hash.encodeSlugValue(opus.selections[param]);
+                let encodedSelectionValues = o_hash.encodeSlugValues(opus.selections[param]);
                 hash.push(param + "=" + encodedSelectionValues.join(","));
             }
         }
 
         for (let key in opus.extras) {
-            let encodedExtraValues = o_hash.encodeSlugValue(opus.extras[key]);
-            try {
+            if (opus.extras[key].length) {
+                let encodedExtraValues = o_hash.encodeSlugValues(opus.extras[key]);
                 hash.push(key + "=" + encodedExtraValues.join(","));
-            } catch(e) {
-                // oops not an arr
-                hash.push(key + "=" + encodedExtraValues);
             }
         }
         $.each(opus.prefs, function(key, value) {
@@ -46,20 +43,20 @@ var o_hash = {
         return hash.join("&");
     },
 
-    encodeSlugValue: function(slugValueArray) {
+    encodeSlugValues: function(slugValueArray) {
         /**
          * Take in a slug value array (like opus.selections, each element
-         * will be a value for the slug) and encode all values in the
+         * will be a list of values for the slug) and encode all values in the
          * array. Return an array that contains encoded values for the
          * slug. This function will be called in updateHash to make sure
          * slug values in the hash are all encoded before updating the URL.
          */
         let slugValue = [];
-        $.each(slugValueArray, function(index, val) {
+        for (let val of slugValueArray) {
             let value = encodeURIComponent(val);
             value = value.replace(/\%20/g, "+");
             slugValue.push(value);
-        });
+        }
 
         return slugValue;
     },
@@ -68,41 +65,68 @@ var o_hash = {
         /**
          * Take in a hash array (each element will be "slug=value") and
          * encode the "value" of each element. Return a hash array that
-         * is consist of "slug=encodedValue". This function will be called
+         * consists of "slug=encodedValue". This function will be called
          * in opus.normalizedURLAPICall to make sure slug values from
-         * new URL are decoded.
+         * new URL are encoded.
          */
         let hash = [];
-        $.each(hashArray, function(index, pair) {
-            let slug = pair.split('=')[0];
-            let value = pair.split('=')[1];
-            value = encodeURIComponent(value);
-            value = value.replace(/\%20/g, "+");
+        for (let pair of hashArray) {
+            // Get the first idx of "=" sign and take all of string after the first
+            // "=" as the value (or a list of values) for the slug.
+            let idxOfFirstEqualSign = pair.indexOf("=");
+            let slug = pair.slice(0, idxOfFirstEqualSign);
+            let value = pair.slice(idxOfFirstEqualSign + 1);
+
+            let valueArray = value.split(",");
+            valueArray = o_hash.encodeSlugValues(valueArray);
+            value = valueArray.join(",");
+
             hash.push(`${slug}=${value}`);
-        });
+        }
 
         return hash;
+    },
+
+    decodeSlugValues: function(slugValueArray) {
+        /**
+         * Take in a slug value array (like opus.selections, each element
+         * will be a list of values for the slug) and decode all values in the
+         * array. Return an array that contains decoded values for the
+         * slug. This function will be called in decodeHashArray to make sure
+         * slug values in the hash are all decoded.
+         */
+        let slugValue = [];
+        for (let val of slugValueArray) {
+            let value = val.replace(/\+/g, "%20");
+            value = decodeURIComponent(value);
+            slugValue.push(value);
+        }
+
+        return slugValue;
     },
 
     decodeHashArray: function(hashArray) {
         /**
          * Take in a hash array (each element will be "slug=value") and
          * decode the "value" of each element. Return a hash array that
-         * is consist of "slug=decodedValue". This function will be called
+         * consists of "slug=decodedValue". This function will be called
          * in getSelectionsExtrasFromHash to make sure slug values in
          * selections and extras are decoded. And it will also be called
          * in initFromHash to make sure slug values in opus.selections are
          * decoded.
          */
-        // console.log(`=== hash from decodeHashArray ===`);
         let hash = [];
-        $.each(hashArray, function(index, pair) {
-            let slug = pair.split('=')[0];
-            let value = pair.split('=')[1];
-            value = value.replace(/\+/g, "%20");
-            value = decodeURIComponent(value);
+        for (let pair of hashArray) {
+            let idxOfFirstEqualSign = pair.indexOf("=");
+            let slug = pair.slice(0, idxOfFirstEqualSign);
+            let value = pair.slice(idxOfFirstEqualSign + 1);
+
+            let valueArray = value.split(",");
+            valueArray = o_hash.decodeSlugValues(valueArray);
+            value = valueArray.join(",");
+
             hash.push(`${slug}=${value}`);
-        });
+        }
 
         return hash;
     },
@@ -151,22 +175,22 @@ var o_hash = {
         let extras = {}; // store qtype from url
 
         $.each(hash, function(index, pair) {
-            let slug = pair.split('=')[0];
-            let value = pair.split('=')[1];
+            let idxOfFirstEqualSign = pair.indexOf("=");
+            let slug = pair.slice(0, idxOfFirstEqualSign);
+            let value = pair.slice(idxOfFirstEqualSign + 1);
 
             if (!(slug in opus.prefs) && value) {
                 if (slug.startsWith("qtype-")) {
                     // each qtype will only have one value at a time
                     extras[slug] = [value];
                 } else {
-                    // If value contains ", ", it's one value from string input. This will make sure "," in
-                    // slug value gets encoded in "%2C" in URL, and store as decoded "," in opus internal variables
-                    // selections.
-                    if (value.includes(", ")) {
-                        selections[slug] = [value];
-                    } else {
-                        selections[slug] = value.split(",");
-                    }
+                    // Leave comments here, need to revisit this later. We have find
+                    // a better way to tell if the slug value is coming from string input.
+                    // if ($(`input[name="${slug}"]`).hasClass("STRING")) {
+                    //     selections[slug] = [value];
+                    // } else {
+                    selections[slug] = value.split(",");
+                    // }
                 }
             }
         });
@@ -203,8 +227,9 @@ var o_hash = {
         hash = hash.split('&');
         hash = o_hash.decodeHashArray(hash);
         $.each(hash, function(index, pair) {
-            let slug = pair.split('=')[0];
-            let value = pair.split('=')[1];
+            let idxOfFirstEqualSign = pair.indexOf("=");
+            let slug = pair.slice(0, idxOfFirstEqualSign);
+            let value = pair.slice(idxOfFirstEqualSign + 1);
             if (value) {
                 if (slug.match(/qtype-.*/)) {
                     // range drop down, add the qtype to the global extras array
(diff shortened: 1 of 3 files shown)


opus/application/static_media/js/search.js
```diff
@@ -237,31 +237,31 @@ var o_search = {
...
 
-        $('#search').on("change", 'input.multichoice, input.singlechoice', function() {
+        $("#search").on("change", "input.multichoice, input.singlechoice", function() {
            // mult widget gets changed
-           let id = $(this).attr("id").split('_')[0];
-           let value = $(this).attr("value").replace(/\+/g, '%2B');
-
-           if ($(this).is(':checked')) {
+           let id = $(this).attr("id").split("_")[0];
+           let value = $(this).attr("value").replace(/\+/g, "%2B");
```
dstopp: i'm sort of confused why there is a two-step replace/regexp here.  If it cannot be combined, perhaps a comment indicating what's going on would be useful.

dstopp (changes requested):

rfrenchseti (changes requested):
So in the currently deployed implementation (which is how things have worked for years), a "+" in the URL means "space" and a "%3B" in the URL means "+". These appear for instrument names like "Cassini ISS" and the Cassini ISS filters like "BL1+GRN", respectively. It's important for this to still be supported for legacy URLs, and I believe these changes break that backwards compatibility.

The whole reason that "+" is replaced with "%3B" is to avoid it conflicting with the "+" that meant "space". It's an annoying hacked workaround that is now getting us in trouble. Honestly I don't see a clean fix to this. We probably have to take advantage of the fact that COISSfilter is the only field that has ever had "+" in the values, and it doesn't ever have spaces.

Here's what I suggest:

- Modify normalizeurl so that when a URL comes in, any existing "+" is replaced with "%20" for any slug value EXCEPT COISSfilter (or its old-slug equivalent). For COISSfilter, leave "+" as-is (this will be needed for the future to avoid continually rewriting this field; see below)
- Modify normalizeurl so that when a URL comes in, any existing "%3B" is replaced with "+" (in that order, of course) (this is safe for both old and new URLs)
- Going forward, leave "+" as-is in the URL - don't encode it as "%3B" anymore
- Going forward, encode "space" as "%20"
- Since replacing "space" with "%20" is really just an escaping of "dangerous" characters in the URL, we should do that with all dangerous characters (like "&"). See issue #719 for details on this. I think solving #719 will automatically solve the "space" issue here as well.
- Make sure all this is carefully commented so we understand it in the future

See JavaScript's "encodeURI" and "decodeURI" for how to do this cleanly.

rfrenchseti:
BTW this will require quite a few new test cases for normalizeurl, both to check that it’s making the right subs and that it isn’t making them when it shouldn’t.

juzen2003:
- Update to make URL have %20 as space and %2B as "+". (This will work for legacy URL)
- Here is the summary and 2 actions added:
    - When URL is pasted in the browser:
        - legacy url:  
            - "+" in url => means space in value
            - "%2B" in url => means "+" in value
        - new url:
            - "%20" in url => means space in value
            - "%2B" in url => means "+" in value

    - When above URL is passed to normalizedURLAPICall:
        - in api call:
            - pass in "+" => space in return value
            - pass in "%20" => space in return value
            - pass in "%2B" => "+" in return value
            - Note: if "+" is passed in with Cassini filter manually (like BL1+GRN), it will complain about the filter had an illegal value
        - after api call (convert new url returned from api call): (**ACTION ADDED**)
            - space => convert to "%20"
            - "+" => convert to "%2B"
            - After above conversion, URL will always have "%20" as space and "%2B" as "+". Since they are all encoded, whenever getHash is called (from getSelectionsExtrasFromHash), values in selections in load (in opus.js) will also be encoded.

    - Selecting checkboxes:   
        - in search selection (on change event hander when selecting checkboxes): (**ACTION ADDED**)
            - select a white space => opus.selections will store it as "%20"
            - select a "+" => opus.selections will store it as "%2B"
            - Values stored in opus.selections will be used to update URL later on, and they are all encoded.

rfrenchseti:
When you take a URL from the existing system that has %2B in it and paste it into the new one, you get a normalizeurl error and the COISSfilter field is deleted entirely.

127.0.0.1:8000/opus/#/instrument=Cassini+ISS&COISSfilter=BL1%2BGRN,BL1%2BIR3,BL1%2BMT3&cols=opusid,instrument,planet,target,time1,observationduration&widgets=COISSfilter,instrument,planet,target&order=time1,opusid&view=search&browse=gallery&cart_browse=gallery&startobs=1&cart_startobs=1&detail=

This appears to be a bug in the current version, too, although the symptoms are different. Try creating a search with a COISS filter that has a + in the name, then paste that URL into a new tab, and watch OPUS die.

juzen2003:
> When you take a URL from the existing system that has %2B in it and paste it into the new one, you get a normalizeurl error and the COISSfilter field is deleted entirely.
> 
> 127.0.0.1:8000/opus/#/instrument=Cassini+ISS&COISSfilter=BL1%2BGRN,BL1%2BIR3,BL1%2BMT3&cols=opusid,instrument,planet,target,time1,observationduration&widgets=COISSfilter,instrument,planet,target&order=time1,opusid&view=search&browse=gallery&cart_browse=gallery&startobs=1&cart_startobs=1&detail=
> 
> This appears to be a bug in the current version, too, although the symptoms are different. Try creating a search with a COISS filter that has a + in the name, then paste that URL into a new tab, and watch OPUS die.

The above example is getting error from normalizedURL api call because in Cassini filter, we don't have "BL1+IR3" and "BL1+MT3" as options in checkboxes area in test environment. Please see this api call from localhost:
http://localhost:8000/opus/__normalizeurl.json?instrument=Cassini+ISS&COISSfilter=BL1%2BGRN,BL1%2BIR3,BL1%2BMT3&cols=opusid,instrument,planet,target,time1,observationduration&widgets=COISSfilter,instrument,planet,target&order=time1,opusid&view=search&browse=gallery&cart_browse=gallery&startobs=1&cart_startobs=1&detail=

In tools, we have these options "BL1+IR3" and "BL1+MT3", please see the same api call from tools:
https://tools.pds-rings.seti.org/opus/__normalizeurl.json?instrument=Cassini+ISS&COISSfilter=BL1%2BGRN,BL1%2BIR3,BL1%2BMT3&cols=opusid,instrument,planet,target,time1,observationduration&widgets=COISSfilter,instrument,planet,target&order=time1,opusid&view=search&browse=gallery&cart_browse=gallery&startobs=1&cart_startobs=1&detail=

The failed case we see from tools: 
URL is like this when we just select those options in checkbox and copy it: (note that all "+" is "%2B" in URL)
https://tools.pds-rings.seti.org/#/instrument=Cassini+ISS&COISSfilter=BL1%2BGRN,BL1%2BIR3,BL1%2BMT3&cols=opusid,instrument,planet,target,time1,observationduration&widgets=COISSfilter,instrument,planet,target&order=time1,opusid&view=search&browse=gallery&cart_browse=gallery&startobs=1&cart_startobs=1&detail=

When we paste it to a new tab, in the step of "normalizedURLAPICall", we didn't replace "+" to "%2B" in returned new_url. And it becomes this:  (note that "+" is in all COISSfilter options)
https://tools.pds-rings.seti.org/#/COISSfilter=BL1+GRN,BL1+IR3,BL1+MT3&instrument=Cassini+ISS&cols=opusid,instrument,planet,target,time1,observationduration&widgets=COISSfilter,instrument,planet,target&order=time1,opusid&view=search&browse=gallery&cart_browse=gallery&startobs=1&cart_startobs=1&detail=

With new code, the above case with COISSfilter=BL1+GRN,BL1+IR3,BL1+MT3 will always be COISSfilter=BL1%2BGRN,BL1%2BIR3,BL1%2BMT3 and OPUS will be running fine.

rfrenchseti:
I don't understand your comment here: "With new code, the above case with COISSfilter=BL1+GRN,BL1+IR3,BL1+MT3 will always be COISSfilter=BL1%2BGRN,BL1%2BIR3,BL1%2BMT3 and OPUS will be running fine." What new code are you talking about? This problem still happens in the current code for this pull.

rfrenchseti:
Here's another problem I just found. Some Voyager ISS "Image ID" have "+" in the string. We need to be able to search on that "+" but right now when you try normalizeURL throws an error and you get a message about it. The reason is that the "+" is being converted to a space before normalizeURL is called. The same is true for string search choices.

rfrenchseti:
Observations:

1) RFC-1866 (HTML 2.0) spec has the following to say about form submission values:

    https://tools.ietf.org/html/rfc1866#section-8.2

```
    1. The form field names and values are escaped: space
        characters are replaced by `+', and then reserved characters
        are escaped as per [RFC-1738]; that is, non-alphanumeric
        characters are replaced by `%HH', a percent sign and two
        hexadecimal digits representing the ASCII code of the
        character. Line breaks, as in multi-line text field values,
        are represented as CR LF pairs, i.e. `%0D%0A'.
```

This covers the `?slug=val&slug=val...` part of the Django API string. It does not cover the path part of the URL (so, for example, `.json` does not have the "." converted to a hex string). Django handles the decoding of this automatically. For any API call, "+" is replaced with space, and then "%xx" is replaced with the appropriate character. So the Python code receives slugs and values with the decoding already done. The Python code never needs to know any kind of substitution was performed.

2) This means the legacy behavior of using "+" for space in things like the instrument name is correct and we shouldn't change it. However, it is *also* OK to use "%20" for a space. For cleanliness, though, I think it's prettier to keep using "+" and it agrees with past use.
3) Issue #719 already notes that we don't properly escape other things, like "&" in a search string. This is also now screwing us up because if you try to search on a string with a "+" in it, that "+" gets into the URL literally and then gets converted to a space in the Django, resulting in a search different from what you expected. I also note now that searching on strings containing "#" or "?" also get things very confused.
4) There is a JavaScript call `encodeURIComponent()` that solves this problem (except for the "+" meaning space). Also, `encodeURIComponent()` converts spaces into "%20", not "+".

    https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

    Note this needs to be run on *each value*, not the string as a whole. We want to leave the slug names and the "=" between slug and value and the "&" between slug/value pairs alone.

5) `normalizeURL.json` returns the new URL and new values *unencoded*. Thus they need to be encoded before they can be put into the URL.

So, here's my suggestion:

1) Create two new functions that encode and decode values in the way we want.
  A) The encode function should call `encodeURIComponent()` and then replace "%20" with "+"
  B) The decode function should replace "+" with "%20" and then call `decodeURIComponent()`
2) Now there's a choice. I don't have an immediate idea of which is better.
  A) Keep all values internal to OPUS (e.g. in "selections") as plain un-encoded strings, and then do the encoding every time you create a hash string
  B) Keep all the values internal to OPUS as encoded strings, and then convert back and forth whenever a selection changes (checkbox is changed, search string is changed).
  I'm going to guess than "A" is the right thing to do, and is basically what we do now except without the proper encode/decode functions.

juzen2003:
- Updates:
- In URL: (hash will be encoded before updating URL)
  - space will be "+"
  - rest of characters will be encoded by encodeURIComponent
- In internal variables like opus.selections, selections, etc: (hash obtained from URL will be decoded before constructing those variables)
  - space will be " "
  - rest of characters will be decoded by decodeURIComponent
- After normalizedURL api call, the return new URL will be encoded before updating URL.

opus/application/static_media/js/hash.js
```diff
@@ -18,20 +18,21 @@ var o_hash = {
...
             if (opus.selections[param].length) {
-                hash.push(param + "=" + opus.selections[param].join(",").replace(/ /g,"+"));
+                let encodedSelectionValues = o_hash.encodeSlugValue(opus.selections[param]);
+                hash.push(param + "=" + encodedSelectionValues.join(","));
             }
         }
 
         for (let key in opus.extras) {
+            let encodedExtraValues = o_hash.encodeSlugValue(opus.extras[key]);
             try {
```
rfrenchseti: Since encodeSlugValue always returns an array now, we probably don't need this try anymore.

opus/application/static_media/js/hash.js
```diff
@@ -45,6 +46,67 @@ var o_hash = {
         return hash.join("&");
     },
 
+    encodeSlugValue: function(slugValueArray) {
```
rfrenchseti: This probably should be called encodeSlugValues since it can encode more than one value

opus/application/static_media/js/hash.js
```diff
@@ -45,6 +46,67 @@ var o_hash = {
...
+        /**
+         * Take in a slug value array (like opus.selections, each element
+         * will be a value for the slug) and encode all values in the
```
rfrenchseti: will be a "list of values" for the slug

opus/application/static_media/js/hash.js
```diff
@@ -45,6 +46,67 @@ var o_hash = {
...
+         * array. Return an array that contains encoded values for the
+         * slug. This function will be called in updateHash to make sure
+         * slug values in the hash are all encoded before updating the URL.
+         */
+        let slugValue = [];
+        $.each(slugValueArray, function(index, val) {
```
rfrenchseti: Given ES6, Debby and I talked about the use of $.each. It's really not needed for such a simple case and has excess overhead. It might be more readable to use the new "for...of" construct:
`for (const val of slugValueArray) {`

opus/application/static_media/js/hash.js
```diff
@@ -45,6 +46,67 @@ var o_hash = {
...
+        });
+
+        return slugValue;
+    },
+
+    encodeHashArray: function(hashArray) {
+        /**
+         * Take in a hash array (each element will be "slug=value") and
+         * encode the "value" of each element. Return a hash array that
+         * is consist of "slug=encodedValue". This function will be called
```
rfrenchseti: "that consists of"

opus/application/static_media/js/hash.js
```diff
@@ -45,6 +46,67 @@ var o_hash = {
...
+         * in opus.normalizedURLAPICall to make sure slug values from
+         * new URL are decoded.
```
rfrenchseti: Do you mean "are ENcoded"?

opus/application/static_media/js/hash.js
```diff
@@ -45,6 +46,67 @@ var o_hash = {
...
+         */
+        let hash = [];
+        $.each(hashArray, function(index, pair) {
```
rfrenchseti: Same use of `for...of` here

opus/application/static_media/js/hash.js
```diff
@@ -45,6 +46,67 @@ var o_hash = {
...
+            let slug = pair.split('=')[0];
+            let value = pair.split('=')[1];
```
rfrenchseti: It's possible for a search string to have an "=" in it. For example, try searching on "a=b". Because the URL will say "searchslug=a=b", the split will break this into 3 parts, and we will only take "a" as the value. You can easily try this by searching a string field on "a=b" and notice that the URL will only say "a". It would be better to take value = ALL of the string after the first "=".

opus/application/static_media/js/hash.js
```diff
@@ -45,6 +46,67 @@ var o_hash = {
...
+            value = encodeURIComponent(value);
```
rfrenchseti: So we have this legacy code everywhere that assumes values (and qtypes) are *lists*, not just single values. This is because the old version of OPUS allowed multiple search values for a given field, e.g. "observationduration1=1,2,3" and "observationduration2=4,5,6". We might actually want to allow this again at some point, so we might as well support it everywhere. It seems like the better thing to do here, rather than just do a single encodeURIComponent, is to split the value at ",", then call encodeSlugValue, and then join the results back with "," again. This also means the encode is only done in one place, so we don't have duplicate code that can diverge over time.

opus/application/static_media/js/hash.js
```diff
@@ -45,6 +46,67 @@ var o_hash = {
...
+        });
+
+        return hash;
+    },
+
+    decodeHashArray: function(hashArray) {
+        /**
+         * Take in a hash array (each element will be "slug=value") and
+         * decode the "value" of each element. Return a hash array that
+         * is consist of "slug=decodedValue". This function will be called
```
rfrenchseti: "that consists of"

opus/application/static_media/js/hash.js
```diff
@@ -45,6 +46,67 @@ var o_hash = {
...
+         * in getSelectionsExtrasFromHash to make sure slug values in
+         * selections and extras are decoded. And it will also be called
+         * in initFromHash to make sure slug values in opus.selections are
+         * decoded.
+         */
+        // console.log(`=== hash from decodeHashArray ===`);
```
rfrenchseti: Remove comment

opus/application/static_media/js/hash.js
```diff
@@ -45,6 +46,67 @@ var o_hash = {
...
+        let hash = [];
+        $.each(hashArray, function(index, pair) {
```
rfrenchseti: `for...of`

opus/application/static_media/js/hash.js
```diff
@@ -45,6 +46,67 @@ var o_hash = {
...
+            let slug = pair.split('=')[0];
+            let value = pair.split('=')[1];
```
rfrenchseti: Same deal with split taking all the remaining parts of the string

opus/application/static_media/js/hash.js
```diff
@@ -96,7 +159,14 @@ var o_hash = {
                     // each qtype will only have one value at a time
                     extras[slug] = [value];
                 } else {
-                    selections[slug] = value.replace("+", " ").split(",");
+                    // If value contains ", ", it's one value from string input. This will make sure "," in
+                    // slug value gets encoded in "%2C" in URL, and store as decoded "," in opus internal variables
+                    // selections.
+                    if (value.includes(", ")) {
```
rfrenchseti: Try to look for the widget to see if it's a string instead of using includes.

opus/application/static_media/js/hash.js
```diff
@@ -169,7 +239,14 @@ var o_hash = {
                     }
                 } else {
                     // these are search params/value!
-                    opus.selections[slug] = value.replace(/\+/g, " ").split(',');
+                    // If value contains ", ", it's one value from string input. This will make sure "," in
+                    // slug value gets encoded in "%2C" in URL, and store as decoded "," in opus internal variables
+                    // opus.selections.
+                    if (value.includes(", ")) {
```
rfrenchseti: Also look for the widget here

opus/application/static_media/js/opus.js
```diff
@@ -558,8 +558,23 @@ var opus = {
...
 
-            // Update URL in browser
-            window.location.hash = "/" + normalizeURLData.new_url.replace(" ", "+");
+            // Get the newURLHash array from new_slugs (data returned from api call).
+            // The reason we don't use new_url directly is because some slug values might
+            // contain "&", and we can't just do .split("&"). This is a more proper way
+            // to get newURLHash array.
+            let newSlugArr = normalizeURLData.new_slugs;
+            let newURLHash = [];
+            for (let idx in newSlugArr) {
```
rfrenchseti: New ES6 uses "for...of"

rfrenchseti (changes requested):

opus/application/static_media/js/hash.js
juzen2003: Using $(`input[name="${slug}"]`).hasClass("STRING") to look for widget with string input will work, but when opus is reload, the page will keep reloading if there is an input with "," in string field. The reason is because  in opusInitialization, the selector will return false because the widget is not rendered yet at this point. And at the same time, both selections (from getSelectionsExtrasFromHash) and opus.selections (from initFromHash) are build. After opusInitialization, load is called and selections will get updated correctly (from getSelectionsExtrasFromHash) and it will be different from opus.selections, which will cause page reload again. (This will happen over and over).

So for now, we didn't update this part and when there a "," in string input, the page will reload one time. And if there is no "," in the string field, page will not reload and search will be performed directly.

opus/application/static_media/js/hash.js
```diff
@@ -45,6 +43,94 @@ var o_hash = {
...
+    encodeSlugValues: function(slugValueArray) {
+        /**
+         * Take in a slug value array (like opus.selections, each element
+         * will be a list of values for the slug) and encode all values in the
+         * array. Return an array that contains encoded values for the
+         * slug. This function will be called in updateHash to make sure
+         * slug values in the hash are all encoded before updating the URL.
+         */
+        let slugValue = [];
+        for (let val of slugValueArray) {
```
rfrenchseti, line 55: For this and all other "for (let..." statements we should use "for (const..." instead assuming the value isn't going to be modified in the loop.

rfrenchseti (approved):

Issue #814, opened 2019-06-25
Make View Gallery and View Table buttons fixed-width


rfrenchseti:
I thought there was already an issue on this but I can't find it...

We should make the "View Gallery" and "View Table" buttons have the same width, so when you switch back and forth the remainder of the nav bar and slider don't jump left and right.

Issue #770, opened 2019-06-10
Cosmetic improvements to Download Options panel


dstopp:
On the cart page, narrow the screen so that the download options are a panel.  Open the panel.
Two things:
 1- if you widen the screen while the panel is open, there is a remnant bit of the panel that remains on the screen **THIS IS NOW FIXED**
 2- the scrollbar on the panel needs to move right to the edge of the panel.  Also, the 'close' button at the top of the panel needs to be adjusted so that it is in a symmetric location.  Also, the gray header is not as tall as the header on the help panel; i think they should be the same. **THIS IS STILL AN OPEN ISSUE**

A suggestion may be that when the download options become a panel, that the header displays the text "Download Options" instead of having that be in the content area.  I also think that would make more sense in the panel context. **THIS IS A GOOD IDEA**

This should make it more consistent w/the other panels/modals that we have.

Issue #708, opened 2019-06-03
The Cart tab needs a slider like the Browse tab


rfrenchseti:
C097: Cart tab needs a slider

PR #815, opened 2019-06-25
708 add slider to cart

dstopp:
Add slider to the cart view + responsiveness.
-Slider behaviour on cart should mimic behaviour on browse tab
-Slider should use max count from number of items in the cart
-Slider should gray out when there are not enough items on the page to slider


opus/application/static_media/js/opus.js
```diff
@@ -196,7 +196,7 @@ var opus = {
         $("#op-result-count").html(opus.spinner).parent().effect("highlight", {}, 500);
 
         // Start the observation number slider spinner - no point in doing a flash here
-        $("#browse .op-observation-number").html(opus.spinner);
+        $(".op-observation-number").html(opus.spinner);
```
rfrenchseti: Are you sure you want to do this for both sliders? Just because the search has changed doesn't mean we need a spinner on the cart. They are unrelated.

rfrenchseti (changes requested):
- Browse slider is very broken. It's short, fat, has no space above it, and no horizontal line.
- On the cart tab, there's no vertical space between the slider and the sort pills when the slider is on the second row.

rfrenchseti (changes requested):
Bug #1:
- Put enough observations in the cart that you can enlarge your browser and not quite fill the page
- Note that the cart slider is greyed out
- Gradually shrink the browser. At some point the slider will no longer be greyed out, and will say "1" in a narrow slider handle (narrower than the max number of obs in the cart). The slider can't be moved.
- Narrow the slider further. At some point the handle will widen to the proper width to hold the total number of obs in the cart, and the slider will be moveable.
- This implies to me that the logic to disable the slider is not working properly.

Bug #2:
- When the slider is on the second row, there is no space between the slider handle and the sort pills. There should be the same space as on the browse tab.

dstopp:
bug#1 - this is on the main branch; i think it needs to be resolved separately from this pull request, as it also happens in the browse tab.  
bug#2 - resolved.

rfrenchseti (changes requested):
There is still no space between the slider handle and the sort order pills on the cart page. It doesn't look like there has been a new push in the past week.

Also there is a responsiveness problem on the cart page. At width 1190px - 1210px there is an extra blank line added between the slider and the sort pills.

opus/application/static_media/js/opus.js
dstopp: good catch, updating to just set the spinner on the #browse tab.

rfrenchseti:
This will fix:
- #821 (naming of download buttons and put them on a separate line)
- #814 (view gallery/table fixed width buttons)
- #708 (add slider to cart page)
- #770 (download options empty when browse widened)

rfrenchseti:
- Go to the cart table view. Drag the slider all the way to the right. There will still be two table rows not visible off the bottom of the screen. This works properly on the browse tab.
- Lines 720-724 in browse.js need to be deleted in your most recent merge (the if checking for the browse tab)
- For browse widths between 900 and 930, if you press the "download options" button the pane slights out but has no contents
- For browse widths around 1100+ and 1301+ the empty cart button is too close to (even on top of) the slider

rfrenchseti:
Current bugs:
- Go to the cart table view. Drag the slider all the way to the right. There will still be two table rows not visible off the bottom of the screen. This works properly on the browse tab.
- For browse widths around 1100+ and 1301+ the empty cart button is too close to (even on top of) the slider. It would be best if the empty cart button maintained a constant distance from the right end of the slider. Make sure it has space for at least 9999 items in the cart.
- We had talked about increasing the breakpoint from 930 to 940 so that the Download buttons don't wrap under Linux, but they are still 930. Need to change opus.css line 1119:
`@media (max-width: 940px) {`
and opus.css line 1152:
`@media (min-width: 941px) {
`

dstopp:
-changed the break to 740 for linux
-set the size of the cart nav container such that the slider will always be at least a certain distance apart from the empty cart button

Issue #699, opened 2019-05-29
Add ability to create new search for all multiple-choice and string metadata fields


rfrenchseti:
On May 29, 2019, at 7:35 AM, Ben Knowles <ben.cisscal@gmail.com> wrote:
> Hey Mike and Matt,
>
> A wee suggestion for your search interface (which I generally think is far superior to the Planetary Image Atlas): a "View all images in this observation" button/link on the Observation Detail page that automatically pulls up a search for all images sharing that Observation_ID. This is something I asked for with the IMDB and never got. I don't know if this would be useful for you, but it would be for me. Just a thought.
>
> Cheers,
> Ben

Hi Rob,

Sure, that could be useful. You mean on the Observation Detail page, on each metadata line, next to the info icon? Another option (instead of or in addition to) would be on the table view.

-B

On 5/29/19 10:55 AM, Rob French wrote:
> Hi Ben -
>
> Matt forwarded your suggestion about OPUS to me. First, thank you very
> much for your idea. We always appreciate feedback, suggestions, and
> questions. I'm the lead developer on OPUS, and you are more than welcome
> to contact me directly in the future with any needs you may have.
>
> If I understand your suggestion, you are looking for a button on the
> "Detail" tab for an observation that would create a new browser tab
> containing a new instance of OPUS, with the search automatically set to
> look for the Cassini Observation_ID of the one you were looking at
> previously, but with no other search terms set. You would still have the
> previous instance of OPUS up and running with no changes. Is that correct?
>
> We generally try to make OPUS generic, rather than specific to a
> particular mission or instrument, so I would be hesitant to do this only
> for Cassini Observation_ID. However, your suggestion opens up a complete
> new feature: We could place a button of some kind next to every line of
> metadata on the Detail tab that represents a string value or a
> multiple-choice value (it wouldn't make much sense for numeric values).
> If you click this button you would get a new OPUS window that searched
> for that value in that field. This would also allow people to do easy
> searches on Volume ID, Data Set ID, Note, etc.
>
> Would that be useful to you?

PR #816, opened 2019-06-25
#699 add searches to detail

rfrenchseti:
This adds a search icon next to all appropriate fields on the Details page

### DIFF
--- a/opus/application/apps/results/templates/results/detail_metadata_internal.html
+++ b/opus/application/apps/results/templates/results/detail_metadata_internal.html
@@ -22,7 +22,7 @@
                                         {% endif %}"
                                         target="_blank">
                                     <i class="fas fa-search"
-                                        title="Open a new OPUS that searches for {{ param_info.label_results }} equal to {{ val }}"></i>
+                                        title='Open a new browser tab containing OPUS that searches for "{{ param_info.label_results }}" equal to "{{ val }}"'></i>
                                 </a>
                             {% endif %}
                         </div>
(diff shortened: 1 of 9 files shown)


opus/application/apps/results/templates/results/detail_metadata_internal.html
```diff
@@ -14,6 +15,16 @@
...
                                 {{ param_info.label_results }}: {{ val }}
                             {% endif %}
+                            {% if param_info.is_string_or_mult %}
+                                <a href="/opus/#/{{ param_info.slug }}={{ val|encode_value }}
+                                        {% if param_info.is_string %}
+                                            &qtype-{{ param_info.slug }}=matches
+                                        {% endif %}"
+                                        target="_blank">
+                                    <i class="fas fa-search"
+                                        title="Open a new OPUS that searches for {{ param_info.label_results }} equal to {{ val }}"></i>
```
dstopp: You have "Open a new OPUS that searches for..." ...       I kind of want there to be some indication of "want to open this in a new tab" sort of thing... 

maybe "Open a new OPUS tab that searches for ..."   or something like that...
I also think that "for Planet equal to Saturn" - {{ param_info.label_results }} should be distinguished from {{ val }} somehow, quotes, bold, something.   Might make it more obvious regarding the search parameters.

dstopp (changes requested):

opus/application/apps/results/templates/results/detail_metadata_internal.html
rfrenchseti: Unfortunately there's no way to put formatting like bold in a standard tooltip. Perhaps some day when we use a fancier library. In the meantime I've changed the wording to:

`Open a new browser tab containing OPUS that searches for "{{ param_info.label_results }}" equal to "{{ val }}"`

Issue #733, opened 2019-06-03
Allow browser search to work properly with perfect scrollbar on detail page


rfrenchseti:
Request from Matt Tiscareno:

A105: When you use Ctrl+F in the browser to search for a text string and the found location is off the bottom of the screen, the scrollbar does not move to make it visible. It worked in the old OPUS.

juzen2003:
Ctrl + F search scroll will not work properly with perfect scrollbar. Please see the followings:
https://github.com/mdbootstrap/perfect-scrollbar/issues/467
https://github.com/mdbootstrap/perfect-scrollbar/wiki/Caveats

If we add the workaround mentioned in the the issue page:
```/* detail tab */
.op-detail-metadata {
    /* setting container overflow to scroll and hide webkit scrollbar is
    the workaround for ctrl + F search scroll to work with perfect scrollbar */
    overflow-y: scroll !important;
}

/* As mentioned above, this is part of steps to make ctrl + F search scroll
to work with perfect scrollbar */
.op-detail-metadata::-webkit-scrollbar {
    display: none;
}
```

Ctrl + F search scroll will work fully in Chrome, but partially in FF and Opera.
In FF, we will see the overlapped scrollbar (default one and ps) even if we set display none for webkit scrollbar. (-webkit-scrollbar is not supported for FF: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar)
In Opera, scrolling won't work if we want to drag and scroll.

The change is updated in my repo for experiment: https://github.com/juzen2003/pds-opus/ 
Branch: search_in_detail_page_issue_733

rfrenchseti:
The final fix doesn't work for Opera or FF under Linux but seems to work for all others.

ArturGudiev:
> Ctrl + F search scroll will not work properly with perfect scrollbar. Please see the followings:
> [mdbootstrap/perfect-scrollbar#467](https://github.com/mdbootstrap/perfect-scrollbar/issues/467)
> https://github.com/mdbootstrap/perfect-scrollbar/wiki/Caveats
> 
> If we add the workaround mentioned in the the issue page:
> 
> ```
> .op-detail-metadata {
>     /* setting container overflow to scroll and hide webkit scrollbar is
>     the workaround for ctrl + F search scroll to work with perfect scrollbar */
>     overflow-y: scroll !important;
> }
> 
> /* As mentioned above, this is part of steps to make ctrl + F search scroll
[...]

PR #822, opened 2019-07-18
Fix Ctrl + F search scroll issue in detail page (issue #733)

juzen2003:
- Fix issue #733 (Work for Chrome, Firefox, Safari, but not Opera)
- Local branch: search_in_detail_page_issue_733


rfrenchseti:
Just for the record, it doesn't work for Linux Chrome, but it now works for Linux FF and Windows Chrome/FF.

PR #824, opened 2019-07-20
Select metadata headers cleanup

### LINKED ISSUE #781: Provide basic instructions in Metadata Selector
OPUS 1 (RIP) included some information on what to actually do in the Metadata Selector ("Drag from here" and "Drop here"). It's not a bad idea to do something like that now. I would suggest putting a header on the left pane "Click to select or deselect metadata" and the right pane "Selected metadata\n(Used for table view, slideshow, and CSV files).

rfrenchseti:
- Fixes #781 
- Clean up select metadata code
- Rename metadata selector to select metadata everywhere
- Add headers to select metadata dialog with instructions
- Implement dialog responsiveness in JS instead of CSS for more flexibility
- Rename #searchMenu to op-search-menu and various other cleanup

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -45,7 +45,7 @@ var o_browse = {
     imageSize: 100,     // default
 
     metadataDetailOpusId: "",
-    metadataSelectorDrawn: false,
+    selectMetadataDrawn: false,
 
     tempHash: "",
     onRenderData: false,
@@ -91,7 +91,7 @@ var o_browse = {
             }
         });
 
-        $("#op-metadata-selector").modal({
+        $("#op-select-metadata").modal({
             keyboard: false,
             backdrop: 'static',
             show: false,
@@ -475,7 +475,7 @@ var o_browse = {
 
             if ((e.which || e.keyCode) == 27) { // esc - close modals
                 o_browse.hideGalleryViewModal();
-                $("#op-metadata-selector").modal('hide');
+                $("#op-select-metadata").modal('hide');
                 // reset range select
                 o_browse.undoRangeSelect();
             }
@@ -849,7 +849,6 @@ var o_browse = {
         // 2. In table view, the top obs will stay the same.
         if (browserResized) {
             currentScrollObsNum = previousScrollObsNum;
-            let scrollbarOffset = o_browse.getScrollbarOffset(obsNum, currentScrollObsNum);
             let infiniteScrollDataObj = $(`${tab} ${contentsView}`).data("infiniteScroll").options;
             let offset = infiniteScrollDataObj.scrollbarOffset;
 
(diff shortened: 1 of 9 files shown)


opus/application/static_media/js/browse.js
```diff
@@ -1325,20 +1294,170 @@ var o_browse = {
...
+           So we use a default value that will cover the common case of a dialog wide
+           enough to cause excessive header word wrap.
+        */
+        if (modalHeaderHeight === 0) {
+            modalHeaderHeight = 57;
+            modalFooterHeight = 68;
+            selectMetadataHeadersHeight = 122;
+            selectMetadataHeadersHRHeight = 1;
+
+        }
```
juzen2003, line 1394: We can remove this extra line above in line 1394.

opus/application/static_media/js/browse.js
```diff
@@ -1325,20 +1294,170 @@ var o_browse = {
...
+        let totalNonGalleryHeight = (footerHeight + mainNavHeight + modalHeaderHeight +
+                                     modalFooterHeight + selectMetadataHeadersHeight +
+                                     selectMetadataHeadersHRHeight);
```
juzen2003: Do we want to rename totalNonGalleryHeight? This is the height of non-scrollable area (height of everything other than .op-no-select) when select metadata modal is open.

opus/application/static_media/js/opus.js
```diff
@@ -617,8 +617,9 @@ var opus = {
         let adjustProductInfoHeightDB = _.debounce(o_cart.adjustProductInfoHeight, 200);
         let adjustDetailHeightDB = _.debounce(o_detail.adjustDetailHeight, 200);
         let adjustHelpPanelHeightDB = _.debounce(opus.adjustHelpPanelHeight, 200);
-        let adjustMetadataSelectorMenuPSDB = _.debounce(o_browse.adjustMetadataSelectorMenuPS, 200);
-        let adjustSelectedMetadataPSDB = _.debounce(o_browse.adjustSelectedMetadataPS, 200);
+        let adjustSelectMetadataHeightDB = _.debounce(o_browse.adjustSelectMetadataHeight, 200);
```
juzen2003: Just a note: we can use non-debounced version ```let adjustSelectMetadataHeightDB = o_browse.adjustSelectMetadataHeight;``` to remove the jumping when browser width is changing around 1000px & 1001px, 775px & 774px, and 652px & 653px.

juzen2003 (changes requested):
- Just two minor changes needed.
- Responsiveness works nicely with adjustSelectMetadataHeight. A note of observation, we will see a little height jumping when text in .op-select-metadata-headers are wrapping up to a new line or shrinking to reduce one line (when browser width is changing around 1000px & 1001px, 775px & 774px, and 652px & 653px). To remove that jumping while adjustSelectMetadataHeight is running the calculation, we can use the non-debounced version in "resize" event handler (line 620 in opus.js).

opus/application/static_media/js/browse.js
rfrenchseti: Done

opus/application/static_media/js/browse.js
rfrenchseti: Rename to totalNonScrollableHeight

opus/application/static_media/js/opus.js
rfrenchseti: Changed

Issue #810, opened 2019-06-19
Make Download Options  and Download Size always visible


rfrenchseti:
On the Cart tab, it would be helpful if the Download Options header and the Download Size were always visible, and the scrollbar only affected the product types. That way as you're selecting/deselecting product types, you can immediately see the effect on download size.

This will be a problem if there is a list of zip files present. Frankly I've never liked where those filenames go. To accomplish this we probably need to figure out a new place to put them.

Issue #764, opened 2019-06-08
It's not obvious how to download files


rfrenchseti:
When you have a very narrow browser and the Cart "download options" slide-over is active, when you actually download a ZIP file the URL goes into the slide-over, and the user doesn't necessarily know it's there.

rfrenchseti:
There is actually a bigger problem here in that it's not obvious the download links will show up in the "download options" area even when it's visible. The links aren't "options" after all. We need to figure out a better place to put the links.

rfrenchseti:
See #821 as well

rfrenchseti:
We have decided to put the download links in a separate popout window (like a support chat window) in the lower right. It will pop up when a download finishes and can be closed. A pretty button in the footer will open it again.

Issue #728, opened 2019-06-03
Better error message when downloading too many selections


rfrenchseti:
C092: Need better error message when trying to download too many selections. This should pop up a dialog instead of putting an entry in the zip file list.

Issue #388, opened 2018-07-18
Add total product sizes and number of products to cart download options tab


rfrenchseti:
Each of the checkbox-product-type lines should also include the number of products that will be included if the product type is selected, along with the number of files that will be downloaded and the total  size of those files. The data is already being returned by the API. It just needs to be displayed and formatted.

PR #826, opened 2019-07-23
Add select/deselect all buttons in download options panel in cart tab (Issue #267). 

### LINKED ISSUE #267: Download options tab should have select all / deselect all for file types
For images with lots of associated product types, it takes a lot of effort to check or uncheck all the file type boxes. We need a separate "select/deselect all" for each product type category, as well as one global select/deselect all for all product types.

juzen2003:
- Fixes #267, #770, #810, #388, #764, #728 
- Local branch: issues_in_product_type_selections


opus/application/static_media/js/cart.js
```diff
@@ -74,17 +74,51 @@ var o_cart = {
...
-            let add_to_url = o_cart.getDownloadFiltersChecked();
-            o_cart.lastRequestNo++;
-            let url = "/opus/__cart/status.json?reqno=" + o_cart.lastRequestNo + "&" + add_to_url + "&download=1";
-            $.getJSON(url, function(info) {
-                if (info.reqno < o_cart.lastRequestNo) {
-                    return;
-                }
-                o_cart.hideDownloadSizeSpinner(info.total_download_size_pretty);
-            });
+        $("#cart").on("click","#download_options input", function(e) {
```
rfrenchseti: Add a space after "click"

opus/application/static_media/js/cart.js
```diff
@@ -74,17 +74,51 @@ var o_cart = {
...
+            let productList = $(e.target).closest("ul");
+            let productInputs = productList.find("input");
+            let prodTypeSelectAllBtn = productList.prev("div").children().first();
+            let prodTypeDeselectAllBtn = productList.prev("div").children().last();
```
rfrenchseti: This whole thing seems very fragile to me. A slight change to the HTML will break it. Instead, since we have a category on the "ul" and a category on each of the select/deselect all buttons, why not retrieve the category from the "ul" and use it to search by "data" tag and class name for the two buttons?

opus/application/apps/cart/views.py
```diff
@@ -653,11 +654,12 @@ def _get_download_info(product_types, session_id):
             pretty_name = 'Diagram Products'
         else:
             pretty_name = category + '-Specific Products'
+            category_name = 'specific'
```
rfrenchseti: I don't understand the need for a separate "category_name" here. First of all, it's possible to have more than one "X-Specific Products" in the list, like if you have both Voyager and Cassini observations in the cart at the same time. Your current code will use the same category_name for these two cases, which I assume could cause problems later. Instead, the "pretty_name" is already unique, and you could use it instead. If you need it to be a valid URL component/slug name/whatever (without spaces, "-", etc), then in the Django template you can just use "product_cat|slugify".

opus/application/static_media/js/cart.js
```diff
@@ -112,6 +146,79 @@ var o_cart = {
...
+             $(target).hasClass("op-cart-select-all-btn"))) {
+            for (let checkbox of checkboxesOptions) {
+                $(checkbox).prop("checked", true);
+            }
+        } else if (($(target).hasClass("op-cart-deselect-btn") ||
+                    $(target).hasClass("op-cart-deselect-all-btn"))) {
+            for (let checkbox of checkboxesOptions) {
+                $(checkbox).prop("checked", false);
+            }
+        }
```
rfrenchseti, line 235: else.... log an error?

opus/application/static_media/js/cart.js
```diff
@@ -112,6 +146,79 @@ var o_cart = {
...
+        $(target).prop("disabled", true);
+        $(target).siblings().prop("disabled", false);
+        o_cart.updateDownloadFileInfo();
+    },
+
+    updateSelectDeselectBtn: function(isAllSelected, isAllDeselected, selectBtn, deselectBtn) {
+        /**
+         * Enable/disable select all and deselect all buttons based on
+         * parameters allSelected & allDeselected passed in.
```
rfrenchseti: Actually the parameters are named isAllSelected etc. Also more clear to say "Enable/disable one pair of select all and deselect all buttons..."

opus/application/apps/cart/templates/cart/cart.html
```diff
@@ -6,14 +6,22 @@ <h1>Zipped File Links</h1>
...
         <h1>Download Options</h1>
+        <div>
+            <button class="op-cart-select-all-btn" type="button">Select all options</button>
+            <button class="op-cart-deselect-all-btn" type="button">Deselect all options</button>
+        </div>
         <!-- <p>Total Files: <span id="total_files">{{ download_count }}</span></p> -->
         <!-- <p>Total Size (before zip): <span id="op-total-download-size">{{ total_download_size_pretty }}</span></p> -->
         <p class="op-total-size">Total Size (before zip): <span class="spinner">&nbsp;</span><span id="op-total-download-size">{{ total_download_size_pretty }}</span></p>
 
         <div id="download_options">
```
rfrenchseti: Now would probably be a fine time to rename this to "op-download-options" since we're using it everywhere.

opus/application/apps/cart/templates/cart/cart.html
```diff
@@ -6,14 +6,22 @@ <h1>Zipped File Links</h1>
             <span class="spinner">&nbsp;</span>
         </div>
         <h1>Download Options</h1>
+        <div>
+            <button class="op-cart-select-all-btn" type="button">Select all options</button>
+            <button class="op-cart-deselect-all-btn" type="button">Deselect all options</button>
+        </div>
         <!-- <p>Total Files: <span id="total_files">{{ download_count }}</span></p> -->
         <!-- <p>Total Size (before zip): <span id="op-total-download-size">{{ total_download_size_pretty }}</span></p> -->
         <p class="op-total-size">Total Size (before zip): <span class="spinner">&nbsp;</span><span id="op-total-download-size">{{ total_download_size_pretty }}</span></p>
```
rfrenchseti: I think it would help to add a phrase like "Select which product types to include in downloads:" after the size and before the "Select all product types" buttons.

opus/application/static_media/js/cart.js
```diff
@@ -74,17 +74,51 @@ var o_cart = {
...
+
+            let allCheckboxesOptions = $("#download_options input");
+            let isAllOptionsChecked = o_cart.isAllOptionStatusTheSame(allCheckboxesOptions);
+            let isAllOptionsUnchecked = o_cart.isAllOptionStatusTheSame(allCheckboxesOptions, false);
+
+            o_cart.updateSelectDeselectBtn(isAllOptionsChecked, isAllOptionsUnchecked,
+                                           ".op-cart-select-all-btn", ".op-cart-deselect-all-btn");
+        });
+
+        // Event handler when clicking "Select all options" and "Deselect all options" buttons.
```
rfrenchseti: Make sure to change comment based on new button names

opus/application/static_media/js/cart.js
```diff
@@ -112,6 +146,79 @@ var o_cart = {
...
+    },
+
+    updateCheckboxes: function(target, checkboxesOptions) {
+        /**
+         * Update checkboxes and disable/enable select all & deselect all
+         * buttons correspondingly.
+         */
+        if (($(target).hasClass("op-cart-select-btn") ||
+             $(target).hasClass("op-cart-select-all-btn"))) {
+            for (let checkbox of checkboxesOptions) {
```
rfrenchseti: Since "checkboxesOptions" is the result of a jQuery find(), can't you just do `checkboxesOptions.prop("checked", true);` and it will automatically loop through them all? If not, change the "let" to "const".

opus/application/static_media/js/cart.js
```diff
@@ -112,6 +146,79 @@ var o_cart = {
...
+         * Check if all download options of a product type are the same,
+         * return true if all options are selected (when checked is true)
+         * or if all options are deselected (when checked is false). This
+         * needs to be executed after updateCheckboxes (wait until all
+         * checkboxes properties are updated)
+         */
+        if (productList.length === 1) {
+            return checked === productList.is(":checked");
+        }
+        for (let productOption of productList) {
```
rfrenchseti: Use const

rfrenchseti (changes requested):
- The "Select all options" and "Deselect all options" buttons need to be below the "Total Size" line. They should also be called "Select all product types".
- I think it would look better if the Select all/Deselect all buttons were a little to the right, perhaps lined up with the (i) Text instead of with the checkbox.

rfrenchseti (changes requested):
Good start!

- The "Total files" shouldn't say "(before zip)" because the number of files doesn't change with zipping. It would also be nice to have the total size and total files on the same line to save vertical space.
- I think the buttons for "select all product types" and "deselect all product types" are pretty ugly. There's no reason to have them as wide as the whole options pane. Please come up with some ideas to make them look smaller and more aesthetically pleasing.
- The columns for # Obs, # Files, and Size should have their contents right justified
- The values for # Obs and # Files should have thousand-commas added
- I don't think there should be so much space between the checkmark, X, and product category. I also think the icons might look better smaller, but I'd have to see them to be sure. It would be more intuitive if the checkmark and X were basically over the checkbox column, rather than off to the side. Given the checkbox and the space to the right of the checkboxes, there should be enough space in the product category header to have both a checkmark and X over that region.
- I don't think the product names need to be in bold
- Now that we have only the table scrollable, it's possible to put in enough zip files that you can't see the products at all, and there's no way to scroll down to see them. I think with this design we have to implement the "Download Links" pop-out window at the same time (#764).
- The table scrolling includes the table header. The header should stay put and the scrolling should only affect the product types and counts.
- Make the heading "# Obs" and "# Files" so it fits on one header line.
- When the browser is narrow and the download options pane is a slide-out, its width is responsive to the browser width. This wasn't a big deal before, but with the new table layout it can look really bad. The pane slide-out should be a constant width, or at least have a maximum width that's the same as the non-slide-out pane.
- I haven't reviewed any of the code changes.

juzen2003:
- Update based on comments (except the following item is still in progress).
- Item in progress: #764

rfrenchseti:
- The breakpoint for making the download options a slide-out needs to be at 1160px now. I don't know why but Linux fonts are a different size from Mac/Windows.

juzen2003:
- Update based on comments and add a clear all button in the popover window of download links history.

rfrenchseti:
- I think the download history button should go away when the list is empty. Right now it isn't there when OPUS starts up and only shows up when the first file is downloaded. Or perhaps the button should always be there but just greyed out when there are no download links present. Either way it should be consistent.
- Instead of "Zipped File Links" how about "Download Archive Links"?
- At least under Linux Chrome, when you make the browser too short (but not as short as we allow) - about 395 px - the pop up stops looking good. I think it needs to be just a little shorter.
- The modal dialog for "browser is too small" is too far back in the Z hierarchy. It shows up behind the cart options pane and behind the download popout.
- I know this is unrelated, but I just noticed that our footer says "Planetary Ring-Moon Systems Node". I have no idea where that came from. Can you change it to say "PDS" instead of "Planetary"?

juzen2003:
- Update based on comments.

opus/application/apps/cart/templates/cart/cart.html
```diff
@@ -1,41 +1,70 @@
...
+{% load humanize %}
+<div id="cart-sidebar" class="w-100">
     <div id="cart_summary" class="well well-lg">
-        <div id="op-download-links">
-            <h1>Zipped File Links</h1>
-            <ul class="zippedFiles"></ul>
-            <span class="spinner">&nbsp;</span>
+        <div class="op-download-options-header border-bottom border-light">
+            <h1>Download Options</h1>
+            <!-- <p>Total Files: <span id="total_files">{{ download_count }}</span></p> -->
```
rfrenchseti: Remove comments

opus/application/static_media/css/opus.css
```diff
@@ -368,10 +368,10 @@ ul.op-search-menu {
     text-align: center;
 }
 
-#cart .fa-times {
+/* #cart .fa-times {
```
rfrenchseti: Remove comments

opus/application/static_media/css/opus.css
```diff
@@ -889,8 +889,45 @@ a:-webkit-any-link, a::any-link {
...
+#cart .op-cart-select-btn, #cart .op-cart-deselect-btn {
+    width: 4%;
+    outline: none;
+}
+
+/* Remove the focus/hover background for "v" & "x" icon in download options
+panel */
+#cart .op-cart-select-btn:hover, #cart .op-cart-deselect-btn:hover,
+#cart .op-cart-select-btn:focus, #cart .op-cart-deselect-btn:focus {
+    background-color: #F0F3F5 !important;
```
rfrenchseti: Is this really the best way to do this? Isn't there some way to simply disable the hover action entirely? Setting the background to a hardcoded color seems like a hack.

opus/application/static_media/js/cart.js
```diff
@@ -8,7 +8,31 @@
...
+// .popover-body and failed to attached the ps to the actual popover window. (Note: there
+// is no proper way to distinguish between them because the 2nd one is the duplicate of
+// the 1st one).
+const downloadLinksPopoverTemplate = "<div class='popover' role='tooltip'>" +
+                                     "<div class='arrow'></div>" +
+                                     "<h3 class='popover-header'></h3>" +
+                                     "<div class='popover-body'></div>" +
+                                     "<div class='popover-footer'>" +
+                                     "<button class='op-clear-history-btn btn btn-sm btn-secondary'" +
+                                     "type='button' title='Clear all history' disabled>Clear all</button></div>" +
```
rfrenchseti: In other dialogs (like Select Metadata), every word in the button is capitalized, so to be consistent this should be "Clear All"

opus/application/static_media/js/cart.js
```diff
@@ -110,6 +117,161 @@ var o_cart = {
...
+         * buttons correspondingly.
+         */
+        if (($(target).hasClass("op-cart-select-btn") ||
+             $(target).hasClass("op-cart-select-all-btn"))) {
+            $(checkboxesOptions).prop("checked", true);
+        } else if (($(target).hasClass("op-cart-deselect-btn") ||
+                    $(target).hasClass("op-cart-deselect-all-btn"))) {
+            $(checkboxesOptions).prop("checked", false);
+        } else {
+            opus.logError(`Target button in download options left pane has the wrong class.`);
```
rfrenchseti: Should use "..."

opus/application/static_media/js/cart.js
```diff
@@ -110,6 +117,161 @@ var o_cart = {
...
+         }
+    },
+
+    isAllOptionStatusTheSame: function(productList, checked=true) {
+        /**
+         * Check if all download options of a product type are the same,
+         * return true if all options are selected (when checked is true)
+         * or if all options are deselected (when checked is false). This
+         * needs to be executed after updateCheckboxes (wait until all
+         * checkboxes properties are updated)
```
rfrenchseti: Add period at end

opus/application/static_media/js/cart.js
```diff
@@ -155,49 +326,98 @@ var o_cart = {
...
-                    $(`<li>${data.error}</li>`).hide().prependTo("ul.zippedFiles", "#cart_summary").slideDown("fast");
+                    // hide the spinner and display error message in an open modal
+                    $(".op-download-links-contents .spinner").hide();
+                    $(".app-footer .op-download-links-btn").popover("update");
+                    $("#op-download-links-error-msg .modal-body").text(data.error);
+                    $("#op-download-links-error-msg").modal("show");
                 } else {
-                    $(`<li><a href = "${data.filename}" download>${data.filename}</a></li>`).hide().prependTo("ul.zippedFiles", "#cart_summary").slideDown("slow");
+                    // To dynamically update and display contents of an open popover, we have to make sure html
+                    // in both (1) #op-download-links ( content when popover is initialized) and (2) .popover-body
```
rfrenchseti: Remove space after (

opus/application/static_media/js/cart.js
```diff
@@ -155,49 +326,98 @@ var o_cart = {
...
+                    // (when popover is open) are synced up with updates. To achieve this, we have to call show
+                    // method from popover to update content, and make sure the selector managing DOM are selecting
+                    // the same elements in both #op-download-links and .popover-body. (lenght === 2).
```
rfrenchseti: length

rfrenchseti (changes requested):
- The horizontal line under the table header goes all the way to the right edge of the pane, while the line above the header stops at the end of the words. I think the top line has the proper behavior.
- Also how about changing the tooltip to "Select/Deslect all in category" to make it clear how it's different from the bit select/deselect all buttons (edited)

opus/application/static_media/css/opus.css
juzen2003: In our case, we don't want to disable the entire hover & focus effect. We only want to remove the outline and background color when we mouse over to "v" and "x". The background color of the hover effect is coming from button.bg-light:hover (from btn-light:hover if we change the class in html), and the easiest way to bypass this is to customize css for these two buttons. I've modified the styling to "transparent" (not a specific hardcoded color now) so the hover effect background color will just match to its parent container which is still a .bg-light from table sub-header.

juzen2003:
Update based on latest comments.

rfrenchseti (approved):

PR #830, opened 2019-07-25
Fix the issue of the missing (i) icon next to qtype in search widgets (issue #827)

### LINKED ISSUE #827: The tooltip for Any/All/Only has gone away
There used to be an (i) next to the any/all/only dropdown that showed a pop-out with some basic information. See #196 for an example (the text one, not the graphics one). Sometime with all of our modifications this went away, and it needs to be brought back.

juzen2003:
- Fixes #827 
- Local branch: tooltip_issue_for_any_all_only_827
- Use the same popover method for the hover over event on (i) icon next to qtype.

### DIFF
--- a/opus/application/apps/ui/templates/ui/base.html
+++ b/opus/application/apps/ui/templates/ui/base.html
@@ -453,7 +453,7 @@
     <div class="op-overlay"></div>
 
     {% include "ui/footer.html" %}
-    <div id="op-range-qtype-helper-html">
+    <div id="op-qtype-tooltip">
         <p><strong>Any</strong> = The image contains at least a portion of the region bounded by the given min and max.</p>
 
         <p><strong>All</strong> = The image contains the entire requested range from min to max.</p>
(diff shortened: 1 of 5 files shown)


opus/application/static_media/js/widgets.js
```diff
@@ -409,6 +409,27 @@ var o_widgets = {
                 o_hash.updateHash();
             }
 
+            // If we have a widget with qtype options any/all/only, we will append the (i) icon.
+            if ($(`#widget__${slug} select[name="${qtype}"]`).length !== 0) {
```
dstopp, line 406: What do you think about combining the previous 'if' block and just do all the "${qtype}" business together?

opus/application/apps/ui/templates/ui/base.html
```diff
@@ -453,7 +453,7 @@ <h3 class='op-header-text'></h3>
     <div class="op-overlay"></div>
 
     {% include "ui/footer.html" %}
-    <div id="range_query_widget_helper_html">
+    <div id="op-range-qtype-helper-html">
```
dstopp: What do you think of renaming the id to "op-qtype-tooltip" instead?  I'm not clear that we need 'helper' or 'html' in that name and my thought is that this is just the tooltip for the qtype.

dstopp (changes requested):

juzen2003:
- Update based on Debby's comments.

dstopp (approved):

Issue #739, opened 2019-06-03
Removing items from cart page screws up caching


rfrenchseti:
If you remove an item from the cart on the Cart tab and then scroll down enough that a new page gets fetched, the starting obs# retrieved will be incorrect because the backend knows the removed item is gone, but the JS is still holding a place in the DOM for it. This means there will be missing thumbnails.

rfrenchseti:
Suggested solution: Have a "recycle bin" for removed selections. When a selection is removed on the cart page, take it out of the DOM instantly but put it in the recycle bin. The recycle bin would probably be a new tab.

When an item is removed from the cart, the product info will have to be updated because the specific product types might change. This will wreck havoc with checkboxes unless we're careful to preserve what the user previously selected.

rfrenchseti:
Also, we will need to implement the recycle bin in the back end. This could be done in multiple ways, but the easiest might be to add a field to the existing "cart" table that says whether the items is in the recycle bin or not. The various JOINs would have to be modified to select the proper value for that flag. This has the advantage of making things like dataimage.json and the __cart operations easy to modify, simply by including a &recyclebin=0/1 flag in the URL.

dstopp:
-the tooltip icons at the bottom of the image on the cart will need to be modified such that delete an image goes to the recycle pile, vs just setting a class on the image <today>.
-when an observation is deleted from the cart, that observation should be deleted from the cart DOM in both browse and table views to avoid the need to retrieve the cart data and flash the user.  In addition, the observations in the cart page will need to be renumbered in the DOM bits where the observation number is kept.

rfrenchseti:
This will be fixed by this pull: https://github.com/SETI/pds-opus/pull/835
The recycle bin idea is moved to a separate issue

Issue #738, opened 2019-06-03
Allow selection of ranges across cache boundaries


rfrenchseti:
If you select a starting observation for a range, then scroll enough that the cache gets flushed, the starting observation is lost. This prevents the user from selecting ranges larger than 4X the current browser window size.

rfrenchseti:
When the start of the range is on the current browser page, the flashing green square makes it obvious where the start is. But once you scroll away, there's no easy way to know you are even in the middle of selecting a range. It might be nice as part of this fix to include some kind of indication that a range has been started and what the opus_id of the starting observation is.

rfrenchseti:
- Put an overlay in the bottom left corner saying something like "Range Select Start at OPUSID XXXX (ESC to cancel)". It should disappear whenever range select is turned off (like switching tabs)
- Store both opusid (for addrange.json) and obsnum (for DOM highlighting)
- Handle the case where the starting obsnum isn't in the cache (either before or after what's cached)

PR #835, opened 2019-07-30
738 allow selection across cache boundaries

### LINKED ISSUE #787: Race conditions with cart operations
There are a variety of race conditions in the cart handling code.

- API operations can happen out of order depending on network delays. Thus it's possible to add an obs and then remove it, while the backend sees a remove followed by an add. The UI will think it knows the state of the obs and highlight it appropriately, but it will disagree with the actual database status.
- reset.json doesn't currently take a reqno, so its return can be out of order with add/remove calls.
- A slow reset call could be an even bigger problem, because the UI thinks the entire cart is empty and does a reload, but when the reload happens the cart may or may not actually be empty.

dstopp:
-on start of select range, pop up overlay with range start information
-allow selection of observations in range to span across cached data
-make sure range selection works in both directions

Fixes: #787, #738, #739


rfrenchseti (changes requested):
- I think the popup should distinguish between "Add range to cart starting at observation #x: opusid" and "Remove range from cart starting at observation #x: opusid". Just plain "Range select start" doesn't explain that it's a cart operation in progress. The wording given here matches that in the hamburger menu, and the format #x: yyy matches the hover text.
- When you start an add range and then drag the slider to a different cache region, the new thumbnail hamburger menus say "start add range" instead of "end add range".
- Now that we can add lots of observations, several bugs have been revealed that were probably there before but need to be fixed:
  - If adding the range fails because it exceeds the maximum number, the UI still selects the thumbnails even though the backend won't have added them to the cart. We need to wait for the addrange.json to return with a success code before changing the thumbnails. This will also solve the "no spinner" problem commented on around line 315 (which is not the same spinner problem mentioned below here).
  - When add.json or addrange.json returns with a failure code and an error dialog box pops up, the browse tab slider position is reset to 1 for no obvious reason.
  - When there are a lot of items in the cart and you switch to the cart tab, there is no spinner while it's loading the data until the very end (there's a long pause before the spinner starts)
- Another bug I noticed at the same time if you feel like fixing it here: If you pull down a hamburger menu and then move the slider, the hamburger menu stays present on the screen. The same is true of scrolling. Once this happens things get very confused if you start a range - the slider jumps.

opus/application/apps/ui/templates/ui/base.html
```diff
@@ -350,6 +356,8 @@ <h3 class='op-header-text'></h3>
                 </div>
             </div>
         </div>
+
```
rfrenchseti: Does this have the wrong indent level?

opus/application/static_media/js/cart.js
```diff
@@ -631,17 +548,73 @@ var o_cart = {
...
+        } else {
+            // note - doing it this way handles the obs on the browse tab at the same time
+            let fromElem = o_browse.getGalleryElement(fromOpusId);
+            fromIndex = $(`${tab} .op-thumbnail-container`).index(fromElem);
+            action = (fromElem.hasClass("op-in-cart") ? "remove" : "add");
+        }
+
+        let url = o_cart.getEditURL(opusIdRange, action);
+
+        // just so we don't have to continually checke for both addrange and add...
```
rfrenchseti: "checke"

opus/application/static_media/js/cart.js
```diff
@@ -631,17 +548,73 @@ var o_cart = {
...
+            let toObsNum = o_browse.getGalleryElement(toOpusId).data("obs");
+
+            // reorder if need be
+            if (fromObsNum > toObsNum) {
+                [fromObsNum, toObsNum] = [toObsNum, fromObsNum];
+                [fromOpusId, toOpusId] = [toOpusId, fromOpusId];
+            }
+            opusIdRange = `${fromOpusId},${toOpusId}`;
+            let fromElem = $(`${tab} .op-gallery-view`).find(`[data-obs=${fromObsNum}]`);
+            let toELem = $(`${tab} .op-gallery-view`).find(`[data-obs=${toObsNum}]`);
```
rfrenchseti: ELem should not have a capital L

opus/application/static_media/js/cart.js
```diff
@@ -631,17 +548,73 @@ var o_cart = {
...
 
         o_cart.showCartCountSpinner();
         o_cart.showDownloadSpinner();
+        $(`.op-page-loading-status > .loader`).show();
 
-        o_cart.lastRequestNo++;
-        $.getJSON(url  + add_to_url + "&reqno=" + o_cart.lastRequestNo, function(statusData) {
-            // display modal for every error message returned from api call
+        // handle it as range
+        if (toOpusId != undefined) {
```
rfrenchseti: !==

opus/application/static_media/js/cart.js
```diff
@@ -631,17 +548,73 @@ var o_cart = {
...
+            // note that only one of the range endpoints can potentially be not present in the DOM,
+            // as we know the user just clicked on one of them to get here.
+            if (fromElem.length > 0) {
+                length = (toELem.length === 0 ? elementArray.length - $(`${tab} .op-thumbnail-container`).index(fromIndex) : length);
+                fromIndex = $(`${tab} .op-thumbnail-container`).index(fromElem);
+            } else {
+                length = $(`${tab} .op-thumbnail-container`).index(toELem);
+            }
+        } else {
+            // note - doing it this way handles the obs on the browse tab at the same time
```
rfrenchseti: What does "this way" mean?

opus/application/static_media/js/cart.js
```diff
@@ -631,17 +548,73 @@ var o_cart = {
...
+            let fromElem = o_browse.getGalleryElement(fromOpusId);
+            fromIndex = $(`${tab} .op-thumbnail-container`).index(fromElem);
+            action = (fromElem.hasClass("op-in-cart") ? "remove" : "add");
+        }
```
rfrenchseti, line 612: I think it would be more clear what's going on if we set "length = 1" here rather than using the default at the top of the function.

opus/application/static_media/js/cart.js
```diff
@@ -631,17 +548,73 @@ var o_cart = {
...
+            let toELem = $(`${tab} .op-gallery-view`).find(`[data-obs=${toObsNum}]`);
+            length = toObsNum - fromObsNum + 1;
+            // note that only one of the range endpoints can potentially be not present in the DOM,
+            // as we know the user just clicked on one of them to get here.
+            if (fromElem.length > 0) {
+                length = (toELem.length === 0 ? elementArray.length - $(`${tab} .op-thumbnail-container`).index(fromIndex) : length);
+                fromIndex = $(`${tab} .op-thumbnail-container`).index(fromElem);
+            } else {
+                length = $(`${tab} .op-thumbnail-container`).index(toELem);
+            }
```
rfrenchseti, line 606: I think it would be more clear what's going on if we set "fromIndex = 0" here rather than using the default at the top of the function.

opus/application/static_media/js/cart.js
```diff
@@ -631,17 +548,73 @@ var o_cart = {
...
+    },
+
+    sendEditRequest: function(url) {
+        o_cart.lastRequestNo++;
+        return $.getJSON(url + "&reqno=" + o_cart.lastRequestNo);
+    },
+
+    toggleInCart: function(fromOpusId, toOpusId) {
+        let tab = opus.getViewTab();
+        let length = 1;
```
rfrenchseti: See comments below...could set length and fromIndex to undefined or null or something here to enforce their setting below and avoid potential bugs.

rfrenchseti (changes requested):
- When adding a range the little window as "Add range from cart" instead of "Add range to cart". The hamburger menu says this too.
- The "loading" spinner needs to have a delay on it just like the cart badge spinners do (see showCartCountSpinner). Right now every time you click on a trashcan/cart icon, you get a flicker of the big spinner even if it returns really fast.
- You need to look at updateBadges. I think its whole purpose is gone now that we don't need to add items to the cart one at a time when doing an addrange.

opus/application/apps/ui/templates/ui/base.html
dstopp: ok, looks like my editor is dropping tabs instead of spaces or something.  fixed.

opus/application/static_media/js/cart.js
dstopp: done

opus/application/static_media/js/cart.js
dstopp: at least it was consistent :)  

done.

opus/application/static_media/js/cart.js
dstopp: done

opus/application/static_media/js/cart.js
dstopp: removed comment, no longer applicable

opus/application/static_media/js/cart.js
dstopp: ok

opus/application/static_media/js/cart.js
dstopp: ok

opus/application/static_media/js/browse.js
```diff
@@ -54,6 +54,7 @@ var o_browse = {
     onRenderData: false,
     fading: false,  // used to prevent additional clicks until the fade animation complete
 
+    pageLoaderSpinnerTimer: null,   // used to apply a small delay to the loader on cart edit
```
rfrenchseti: This delay occurs in lots of places, not just on cart edit

rfrenchseti:
The following comments have not been addressed:

- When adding a range the little window as "Add range from cart" instead of "Add range to cart". The hamburger menu says this too.
- You need to look at updateBadges. I think its whole purpose is gone now that we don't need to add items to the cart one at a time when doing an addrange.

opus/application/static_media/js/browse.js
```diff
@@ -1128,16 +1131,16 @@ var o_browse = {
...
         // use the state of the current selected observation to set the icons if one has been selected,
         // otherwise use the state of the current observation - this will identify what will happen to the range
-        let tab = opus.getViewTab();
-        let selectedElem = $(tab).find(".selected");
-        if (selectedElem.length != 0) {
-            inCart = (o_cart.isIn($(selectedElem).data("id")) ? "" : "in");
+        let rangeSelected = o_browse.isRangeSelectEnabled(tab);
+        let rangeText = "";
+        if (rangeSelected !== undefined) {
+            let addRemoveText = (rangeSelected === "removerange" ? "remove range from" : "add range from");
```
rfrenchseti: Here is where it should say "add range to"

opus/application/static_media/js/browse.js
```diff
@@ -1128,16 +1131,16 @@ var o_browse = {
...
+            rangeText = `<i class='fas fa-sign-out-alt fa-rotate-180'></i>End ${addRemoveText} cart here`;
+        } else {
+            let addRemoveText = (inCart !== "in" ? "remove range from" : "add range from");
```
rfrenchseti: And here

opus/application/static_media/js/browse.js
```diff
@@ -1178,7 +1181,17 @@ var o_browse = {
     },
 
     startRangeSelect: function(opusId) {
-        o_browse.undoRangeSelect(); // probably not necessary...
+        let tab = opus.getViewTab();
+        let galleryElement = o_browse.getGalleryElement(opusId);
+        let obsNum = galleryElement.data("obs");
+        let action = (galleryElement.hasClass("op-in-cart") ? "removerange" : "addrange");
+        let actionText = (action === "removerange" ? "Remove" : "Add");
+        $(`${tab} .op-range-select-info-box`).html(`${actionText} range from cart starting at observation #${obsNum} ${opusId}  (ESC to cancel)`).addClass("op-range-select");
```
rfrenchseti: Also needs to change "from"/"to" here

opus/application/static_media/js/browse.js
```diff
@@ -1191,13 +1204,36 @@ var o_browse = {
...
+    showPageLoaderSpinner: function() {
+        if (o_browse.pageLoaderSpinnerTimer === null) {
+            o_browse.pageLoaderSpinnerTimer = setTimeout(function() {
+                $(`.op-page-loading-status > .loader`).show(); }, opus.spinnerDelay);
+        }
+    },
+
+    hidePageLoaderSpinner: function() {
+        if (o_browse.pageLoaderSpinnerTimer !== null) {
+            // This should always be true - we're just being careful
```
rfrenchseti: Actually I don't think this should always be true. For example, if you scroll such that the next page is being prefetched (spinner comes on), you can still click on a thumbnail to add it to the cart (also turns on spinner). Then when the page load is finished it turns off the spinner, and later when the cart operation returns it also tries to turn off the spinner, but the spinner is already off by then.

The right way to fix this is probably to reference count the on/off actions, since they should always be paired, and that way the spinner won't turn off until the last operation is complete. However, this sounds like a recipe for bugs, and it isn't a common occurrence, so I suggest we just remove this comment, perhaps put a new comment describing this issue, and forget about it for now.

opus/application/static_media/js/cart.js
```diff
@@ -526,92 +523,12 @@ var o_cart = {
...
-
-            o_browse.updateCartIcon(fromOpusId, action);
-            o_cart.editCart(fromOpusId, action);
-            return action;
-        }
-    },
-
     // action = add/remove/addrange/removerange/addall
-    editCart: function(opusId, action, updateBadges=true) {
+    getEditURL: function(opusId, action, updateBadges=true) {
```
rfrenchseti: Remove updateBadges

opus/application/static_media/js/cart.js
```diff
@@ -631,17 +548,74 @@ var o_cart = {
...
+    },
+
+    toggleInCart: function(fromOpusId, toOpusId) {
+        let tab = opus.getViewTab();
+        let length = null;
+        let fromIndex = null;
+        let action = null;
+        let elementArray = $(`${tab} .op-thumbnail-container`);
+        let opusIdRange = fromOpusId;
+        let updateBadges = true;
```
rfrenchseti: Remove updateBadges

opus/application/static_media/js/cart.js
```diff
@@ -650,18 +624,30 @@ var o_cart = {
...
+                    if (action === "add") {
+                        $(`.op-thumbnail-container[data-id=${opusId}]`).addClass("op-in-cart");
+                    } else {
+                        $(`.op-thumbnail-container[data-id=${opusId}]`).removeClass("op-in-cart");
+                    }
+                    $("input[name="+opusId+"]").prop("checked", checked);
+                    o_browse.updateCartIcon(opusId, status);
+                });
+                // we only update the cart badge result count from the latest request
+                if (updateBadges) {
```
rfrenchseti: Remove updateBadges

rfrenchseti (changes requested):
The following comments have not been addressed:

- When adding a range the little window as "Add range from cart" instead of "Add range to cart". The hamburger menu says this too.
- You need to look at updateBadges. I think its whole purpose is gone now that we don't need to add items to the cart one at a time when doing an addrange. Please remove it.

Also:

- When disabling clicking during cart operations, it only disables the gallery thumbnail area. You can still click on tab names. Is this the desired behavior?
- You can still Ctrl+click on an item in the cart tab (gallery or table) and have it add/remove from the cart. This needs to be disabled as well.

rfrenchseti:
Also I don't think this fully fixes #787 because it doesn't do anything about the potential reset.json problems.

rfrenchseti (approved):

Issue #162, opened 2015-06-13
Implement "Preprogrammed Ranges"


basilleaf:
Originally reported by: **lisa ballard (Bitbucket: [basilleaf](https://bitbucket.org/basilleaf), GitHub: [basilleaf](https://github.com/basilleaf))**

---

Useful Ranges were never ported to OPUS 2 but they appear in OPUS 1 ![Screen Shot 2015-06-12 at 6.33.16 PM.png](https://bitbucket.org/repo/ex8BEA/images/1098419844-Screen%20Shot%202015-06-12%20at%206.33.16%20PM.png)

---
- Bitbucket: https://bitbucket.org/ringsnode/opus2/issue/162

rfrenchseti:
Implementation ideas:

- The details of "useful ranges" should be maintained in Python and served to the JavaScript along with a numerical input widget. The useful ranges can be encoded as a list containing a series of descriptions and value ranges.
- The useful ranges will be available as a popup box or as a selectable dropdown menu. When a range is chosen, the numerical fields will be filled in.

rfrenchseti:
Also, to support the example given in the original bug report, the JSON should allow the encoding of categories, with each category containing its own set of descriptions and value ranges.

rfrenchseti:
As an added bonus, we could have the same thing work for string fields. It wouldn't be a "range", but a single value.

rfrenchseti:
Another good thing for ranges would be wavelength - for example, have selections for each useful color along with its wavelength range.

rfrenchseti:
- Whenever there's range info, we always show the dropdown with the categories, with the appropriate ranges hidden/shown based on what the user has typed so far. There might be zero ranges, one range, or multiple ranges available. When the user hasn't typed anything, all the ranges are shown.
- When there is one or more ranges shown, disable verification of the input and show a "valid input" border
- When there are zero ranges shown, go ahead and do verification normally. This will catch the case of the user typing in a range name that doesn't exist, as well as handling the normal case of just typing in numbers like they normally do.
- When the user hits enter, if there's a single range shown, pretend the user just clicked on it and fill in both min/max boxes. If there are zero or multiple ranges shown, then just validate it as normal and let the back end decide what to do with it.

rfrenchseti:
For wavelength, we want:
- PDS4 wavelength ranges
- CRC color names
- Various photometric systems: https://en.wikipedia.org/wiki/Photometric_system

rfrenchseti:
[...]

PR #844, opened 2019-08-21
Preprogrammed ranges info in widgets (#162)

juzen2003:
- Initial implementation for #162 
- Local branches: useful_ranges_162
- Currently available in the following widgets:
  - ring radius
  - wavelength
- Some notes for behaviors of the dropdown list:
  - "Preprogrammed Ranges" button is placed below input fields. I was trying to put it in the same line with input fields, but default font sizes are different in different browsers. Without making the button smaller and widen the widget, the button will not stay in the same line with input fields in Safari and FF, and it will shrink to the next line once the widget width is getting narrower, so I make it a link button and place it below input fields (regular button in the next line looks a bit weird).
  - When dropdown list is open, and the widget column is scrolled, the dropdown list will close. This matches the behavior of string autocomplete dropdown.
  - When scrolling reaches to both ends in dropdown list and we keep scrolling (or wheeling), overscrolling will not happen (outer ps in widget column will not get scrolled in this case).
  - When dropdown list is closed, any expanded items will be collapsed. So the next time when dropdown list is open, all list items are collapsed and only categories are shown.


opus/import/table_schemas/param_info.json
```diff
@@ -98,6 +98,11 @@
         "field_key": "primary",
         "field_notnull": true
     },
+    {
```
rfrenchseti: This should go above "id" so "id" and "timestamp" are always the last two fields in every table.

opus/application/apps/ui/templates/ui/widget.html
```diff
@@ -45,6 +46,49 @@ <h6 class="card-title mt-0">
...
+                                                </div>
+                                                <div class="col-4 font-weight-bold">
+                                                    {{range_data_entry.heading1|title}}
+                                                </div>
+                                                <div class="col-4 font-weight-bold">
+                                                    {{range_data_entry.heading2|title}}
+                                                </div>
+                                            </div>
+                                            {% for range_info in range_data_entry.ranges %}
+                                                        <div class="op-useful-ranges-data-item row" data-min="{{range_info.field1|intcomma}}" data-max="{{range_info.field2|intcomma}}" data-widget="widget__{{slug}}">
```
rfrenchseti: Fix indent amounts

opus/application/apps/ui/templates/ui/widget.html
```diff
@@ -45,6 +46,49 @@ <h6 class="card-title mt-0">
 
                         {{ form|safe }}
                 </ul>
+                {% if ranges %}
+                        <div class="op-useful-ranges dropdown">
```
rfrenchseti: Fix indent amounts

opus/application/apps/ui/templates/ui/widget.html
```diff
@@ -4,6 +4,7 @@
     #op-qtype-tooltip div is appended in base.html, and popover is initialized
     in getWidget function in widgets.js
 {% endcomment %}
+{% load humanize %}
```
rfrenchseti, line 7: Didn't we decide to get rid of this?

opus/application/static_media/css/opus.css
```diff
@@ -1939,3 +1940,51 @@ ul.ui-autocomplete a.ui-state-active {
     animation: hvr-ripple-in 1s infinite;
     -webkit-animation: hvr-ripple-in 1s infinite;
 }
+
+/* Styling for ranges info dropdown */
+.op-useful-ranges .dropdown-item:active {
```
rfrenchseti: Should call the class op-preprogrammed-ranges for consistency

opus/application/static_media/js/widgets.js
```diff
@@ -409,7 +462,13 @@ var o_widgets = {
                     let helpIcon = '<li class="op-range-qtype-helper">\
                                     <a class="text-dark" tabindex="0" data-toggle="popover" data-placement="left">\
                                     <i class="fas fa-info-circle"></i></a></li>';
-                    $(`#widget__${slug} .widget-main ul`).append(helpIcon);
+
+                    // if ($(`#widget__${slug} .widget-main ul .op-useful-ranges`).length !== 0) {
```
rfrenchseti: Remove comments

opus/application/apps/ui/templates/ui/widget.html
```diff
@@ -45,6 +46,49 @@ <h6 class="card-title mt-0">
 
                         {{ form|safe }}
                 </ul>
+                {% if ranges %}
+                        <div class="op-useful-ranges dropdown">
+                            <button type="button" id="dropdownMenu" class="btn btn-link dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
```
rfrenchseti: Rename dropdownMenu to op-ranges-dropdown-menu

opus/import/do_param_info.py
```diff
@@ -54,6 +68,13 @@ def create_import_param_info_table():
                            f'"{category_name}/{field_name}" has RANGE type '
                            +'without numerical format')
 
+            # if pi_ranges exist in .json, get the corresponding ranges info
+            # from dict and convert it to str before storing to database
+            ranges = column.get('pi_ranges', None)
+            if ranges:
+                ranges = ranges_json.get(ranges)
```
rfrenchseti: Use ranges_json[ranges] because we don't want to have a default.
Check to see if key is in ranges first and display an error if not.

opus/import/do_param_info.py
```diff
@@ -54,6 +68,13 @@ def create_import_param_info_table():
                            f'"{category_name}/{field_name}" has RANGE type '
                            +'without numerical format')
 
+            # if pi_ranges exist in .json, get the corresponding ranges info
```
rfrenchseti: exists

opus/import/table_schemas/param_info_ranges.json
```diff
@@ -0,0 +1,1091 @@
+{
+    "obs_ring_geometry": [
+        {
+            "cat_name": "jupiter rings",
```
rfrenchseti: All of these names (category and fields) should be capitalized the way we want them to display. We shouldn't re-capitalize them later, taking control away from the person who wrote this JSON file.

opus/import/table_schemas/param_info_ranges.json
```diff
@@ -0,0 +1,1091 @@
...
+            "heading1": "min",
```
rfrenchseti: Change to "Min (km)" and "Max (km)"

opus/import/table_schemas/param_info_ranges.json
```diff
@@ -0,0 +1,1091 @@
...
+                    "unit": "km",
+                    "qtype": null
+                }
+            ]
+        }
+    ],
+    "obs_wavelength": [
+        {
+            "cat_name": "wavelengths",
+            "heading1": "min",
```
rfrenchseti: Change to "Min (microns)" and "Max (microns)"

opus/application/apps/ui/templates/ui/widget.html
```diff
@@ -45,6 +46,49 @@ <h6 class="card-title mt-0">
...
+                                Preprogrammed Ranges
+                            </button>
+                            <ul class="dropdown-menu scrollable-menu pl-0 w-100" aria-labelledby="dropdownMenu" role="menu">
+                                {% for range_data_entry in ranges %}
+                                    <li class="mr-0 d-block">
+                                        <a class="dropdown-item bg-light dropdown-toggle collapsed" data-toggle="collapse" href="#{{range_data_entry.cat_name|slugify}}" role="button" aria-expanded="false" aria-controls="{{range_data_entry.cat_name|slugify}}">{{range_data_entry.cat_name|title}}</a>
```
rfrenchseti: Don't title case here - do it in the JSON file. See below.

opus/application/apps/ui/templates/ui/widget.html
```diff
@@ -45,6 +46,49 @@ <h6 class="card-title mt-0">
...
+                                                    {{range_data_entry.heading1|title}}
+                                                </div>
+                                                <div class="col-4 font-weight-bold">
+                                                    {{range_data_entry.heading2|title}}
+                                                </div>
+                                            </div>
+                                            {% for range_info in range_data_entry.ranges %}
+                                                        <div class="op-useful-ranges-data-item row" data-min="{{range_info.field1|intcomma}}" data-max="{{range_info.field2|intcomma}}" data-widget="widget__{{slug}}">
+                                                            <div class="col-4">
+                                                                {{range_info.name|title}}
```
rfrenchseti: Don't title case here

rfrenchseti (changes requested):
- Make the "Preprogrammed ranges" the same color and weight as the main categories on the left of the search page. Don't change color on hover.
- If there is only one category, open it automatically.
- Put the units after the min and max table headers like "Min (km)".
- Figure out how to right-justify the numbers and the table header so that we can align things by decimal point. You will need some amount of margin on the right to make it look good.
- Change the width of the whole drop-down box to be narrower. It doesn't need to be the width of the whole widget.
- For wavelength, remove Gamma Ray, X-Ray, and Submillimeter and everything later
- For wavelength, fix range of ultraviolet
- For wavelength, add other colors found at https://www.energy.gov/sites/prod/files/styles/borealis_photo_gallery_large_respondlarge/public/2018/04/f50/figure%201.jpg?itok=K1QNG90T
Use the data from column 1 (CRC)
- For ring radius, figure out what's wrong with the Dawes Ringlet and the Anthe, Methone, and Pallene rings. Also I think those last three should be called "Ring Arcs".

juzen2003:
# Current progress
- Items done:
    - Update the label and color in the widget.
    - Make sure the dropdown stays right below the corresponding input all the time.
    - Properly collapse/expand dropdown categories based on user's input.
    - Highlight matched characters in ranges names based on user's input.
    - Prevent race conditions of collapsing/expanding animations when user types fast in input field.
    - If there is a collapsed empty category, it will not be expandable when clicking on it.
    - When input is focus and dropdown shows up, make sure dropdown will not hide when clicking inside the input once again.
    - Add proper behaviors to open the dropdown at the right time. (Now it will open any time when user clicks a non-focus input.) 
    - Properly call input validation and search (mentioned in #162).
    - If you pull up the dropdown, scroll it, leave focus, scrollbar in dropdown will be reset to top.
    - If you start typing a range name, it will scroll to the first category where there are still range names showing.
    - When there are no results in a category, the category will be hidden.
    - When there is only one result and you hit enter, it will select that result.
    - When you use TAB to change focus into an input field, the dropdown will show up.
    - Figure out how to right-justify the numbers and the table header so that we can align things by decimal point.
    - Adjust the dropdown width and responsiveness.
    - Update json data based on email (mentioned in #162)
- Items in progress:

opus/application/apps/ui/templates/ui/widget.html
juzen2003: This is not the customized one, this is for filter "intcomma" to add "," in numbers.

opus/import/do_param_info.py
```diff
@@ -23,6 +24,19 @@ def create_import_param_info_table():
...
 
+    # read json file for ranges info
+    ranges_filename = os.path.join('table_schemas', 'param_info_ranges.json')
+    with open(ranges_filename, 'r') as fp:
+        try:
+            # read contents (str) and convert it to a json object (dict)
+            contents = fp.read()
+            ranges_json = json.loads(contents)
+        except json.decoder.JSONDecodeError:
+            impglobals.LOGGER.log('debug', f'Was reading ranges json file "{ranges_filename}"')
```
rfrenchseti: This can be "logger.log" instead of "impglobals.LOGGER.log"

opus/import/do_param_info.py
```diff
@@ -54,6 +68,17 @@ def create_import_param_info_table():
...
+            # if pi_ranges exists in .json, get the corresponding ranges info
+            # from dict and convert it to str before storing to database
+            ranges = column.get('pi_ranges', None)
+            if ranges:
+                if ranges in ranges_json:
+                    ranges = ranges_json[ranges]
+                    ranges = json.dumps(ranges)
+                else:
+                    logger.log('error',
+                               f'pi_ranges: "{ranges}" is not in "{ranges_filename}"')
```
rfrenchseti, line 81: Need to "return False" after this so the import is aborted. Same is true of previous error logging for units at line 55 that I didn't catch before.

opus/application/apps/search/forms.py
```diff
@@ -114,11 +115,23 @@ def __init__(self, *args, **kwargs):
...
                 pi = get_param_info_by_slug(slug, 'search')
+                # placeholder for input hints (only apply to Min input for now)
+                hints = pi.field_hints1 if pi.field_hints1 else ''
 
                 self.fields[slug] = MultiFloatField(
-                     required=False,
-                     label = label,
-                     widget = forms.TextInput(attrs={'class':label + ' RANGE'}),
+                    required=False,
+                    label = label.title(),
```
rfrenchseti: What is the "title()" for?

rfrenchseti (changes requested):
A few comments on brief experimentation:

- If you pull up the dropdown, scroll it, leave focus, and then come back, the dropdown remembers where the scroll was before. I think it would be better to reset it to the top.
- If you start typing a range name, it should scroll to the first category where there are still range names showing so you know there is a positive result.
- When there are no results in a category, right now it's confusing because it looks like you can still expand the category by clicking on it because of the ">" arrow. It might work better to grey out the category name, and perhaps remove the ">" arrow. In fact I think I'm changing my mind about showing empty categories at all. With the long list of ring categories, it's just confusing. Perhaps we should hide the inactive categories entirely.
- When there is only one result and you hit enter, it doesn't select that result.
- When you use TAB to change focus into an input field, it doesn't pop up the dropdown.
- If the focus is in a numeric range and the dropdown is showing, and then you click on the input field a second time, the dropdown goes away.
- The wavelength colors should be in different ranges. The red/green/blue colors should be marked something like "CRC Wavelength Ranges", the others should be "PDS4 Wavelength Ranges" and the eventual Johnson filters should be marked "Johnson Filters".
- "Name" in "Radius of Feature Name" should be capitalized.

opus/application/apps/search/forms.py
juzen2003: Change to capitalize(), this is used to change "min" & "max" to "Min" & "Max".

juzen2003:
Update based on comments.

opus/application/apps/search/forms.py
rfrenchseti: OK, I see what's going on here - since the min/max is also a class name. That does seem kind of like a bad class name, though. How about prepending something like "op-range-input-" to the "min"/"max" to make a better class name?

opus/application/apps/search/forms.py
```diff
@@ -114,11 +115,27 @@ def __init__(self, *args, **kwargs):
                 label = 'max' if num == '2' else 'min'
 
                 pi = get_param_info_by_slug(slug, 'search')
+                # placeholder for input hints (only apply to Min input for now)
+                hints = pi.field_hints1 if pi.field_hints1 else ''
```
rfrenchseti: Shouldn't we go ahead and support the hints2 field even though we don't have any cases that use it yet?

opus/application/apps/search/forms.py
```diff
@@ -114,11 +115,27 @@ def __init__(self, *args, **kwargs):
...
+                # dropdown only available when ranges info is available
+                ranges = pi.get_ranges_info()
+                dropdown_class = 'op-ranges-dropdown-menu dropdown-toggle' if ranges else ''
+                data_toggle = 'dropdown' if ranges else ''
 
                 self.fields[slug] = MultiFloatField(
-                     required=False,
-                     label = label,
-                     widget = forms.TextInput(attrs={'class':label + ' RANGE'}),
+                    required=False,
```
rfrenchseti: Add space around =

opus/application/apps/ui/templates/ui/widget.html
```diff
@@ -45,6 +46,46 @@ <h6 class="card-title mt-0">
...
+                        <ul class="dropdown-menu op-scrollable-menu pl-0" aria-labelledby="op-ranges-dropdown-menu" role="menu">
+                            {% for range_data_entry in ranges %}
+                                <li class="mr-0 d-block" data-category="{{range_data_entry.cat_name|slugify}}">
+                                    <a class="dropdown-item bg-light dropdown-toggle collapsed" data-toggle="collapse" href="#{{range_data_entry.cat_name|slugify}}" role="button" aria-expanded="false" aria-controls="{{range_data_entry.cat_name|slugify}}">{{range_data_entry.cat_name}}</a>
+                                    <div class="container collapse" id="{{range_data_entry.cat_name|slugify}}" data-widget="widget__{{slug}}">
+                                        <div class="op-preprogrammed-ranges-data-header row border-bottom border-light">
+                                            <div class="col-4">
+                                            </div>
+                                            <div class="col-4 font-weight-bold text-right">
+                                                {{range_data_entry.heading1}}
```
rfrenchseti: Elsewhere we've been putting a space after {{ and before }}

opus/application/apps/ui/templates/ui/widget.html
```diff
@@ -45,6 +46,46 @@ <h6 class="card-title mt-0">
...
+                                            </div>
+                                        </div>
+                                        {% for range_info in range_data_entry.ranges %}
+                                            <div class="op-preprogrammed-ranges-data-item row" data-min="{{range_info.field1|intcomma}}" data-max="{{range_info.field2|intcomma}}" data-widget="widget__{{slug}}" data-name="{{range_info.name}}">
+                                                <div class="col-4 op-preprogrammed-ranges-data-name">
+                                                    {{range_info.name}}
+                                                </div>
+                                                <div class="col-4 op-preprogrammed-ranges-min-data">
+                                                    {% if range_info.field1 %}
+                                                        {{range_info.field1|intcomma}}
```
rfrenchseti: What happens with "intcomma" when we have non-int/float field values, like date/time strings?

opus/application/static_media/js/search.js
```diff
@@ -41,41 +41,76 @@ var o_search = {
...
+                $(this).addClass("search_input_invalid");
             }
             $(this).addClass("input_currently_focused");
             $(this).removeClass("search_input_invalid_no_focus");
+
+            // Open the dropdown properly when user tabs to focus in.
+            let slugName = $(this).data("slugname");
+            let inputToTriggerDropdown = $(`#widget__${slugName} input.min`);
+            let preprogrammedRangesDropdown = $(`#widget__${slugName} .op-scrollable-menu`);
+            if (preprogrammedRangesDropdown.length === 0 || !$(e.target).hasClass("min")) {
```
rfrenchseti: Instead of returning here, wouldn't it be better to combine the if directly below and change the condition here to length !==0 && $(e.target)....  That way it's easy to add things later to this function without worrying about it prematurely returning here.

opus/application/static_media/js/search.js
```diff
@@ -41,41 +41,76 @@ var o_search = {
     slugEndpointsReqno: {},
     slugRangeInputValidValueFromLastSearch: {},
 
+    rangesNameMatchedCounter: {},
```
rfrenchseti: It seems odd to have a variable named "Counter" be a dictionary. What is this indexed by? Can you put that in the name somehow? Like rangesNameMatchedCounterBySlug or something.

opus/application/static_media/js/search.js
```diff
@@ -129,10 +167,47 @@ var o_search = {
...
+            if (o_search.rangesNameTotalMatchedCounter === 1) {
+                let matchedCatId = "";
+                for (const eachCat in o_search.rangesNameMatchedCounter) {
+                    if (o_search.rangesNameMatchedCounter[eachCat] === 1) {
+                        matchedCatId = `#${eachCat}`;
+                        break;
+                    }
+                }
+                let allItemsInMatchedCat = $(`${matchedCatId} .op-preprogrammed-ranges-data-item`);
+                for (const singelRangeData of allItemsInMatchedCat) {
```
rfrenchseti: single not singel

opus/application/static_media/js/search.js
```diff
@@ -301,6 +376,199 @@ var o_search = {
...
 
+    addPreprogrammedRangesSearchBehaviors: function() {
+        /**
+         * Add customized event handlers for preprogrammed ranges dropdown and expandable
+         * list when user types/focus in & out of the input. This function will be called
+         * in addSearchBehaviors.
+         */
+
+        // Make sure ranges info shows up automatically when the count of matched characters
+        // is not 0. We do this in the event handler when the collpasing is done. This is to avoid
```
rfrenchseti: collapsing

opus/application/static_media/js/search.js
```diff
@@ -301,6 +376,199 @@ var o_search = {
...
+        $(`#search`).on("click", ".op-scrollable-menu", function(e) {
+            o_search.isTriggeredFromInput = false;
+        });
+
+        // Reset scrollbar to top if there is no matched ranges info when dropdown is open.
+        $("#search").on("shown.bs.dropdown", function(e) {
+            $(".op-scrollable-menu").scrollTop(0);
+        });
+
+        // Make sure dropdown is not shown if user focus into an input with numerical value.
```
rfrenchseti: Is this just for numerical value or also for any string value that doesn't match?

opus/application/static_media/js/search.js
```diff
@@ -301,6 +376,199 @@ var o_search = {
...
+
+            if (o_search.rangesNameMatchedCounter[collapsibleContainerId] === 0) {
+                $(`#${collapsibleContainerId}`).collapse("hide");
+                if (currentValue) {
+                    $(`a.dropdown-item[href*="${collapsibleContainerId}"]`).addClass("op-hide-element");
+                }
+            }
+            o_search.setRangesDropdownScrollbarPos(preprogrammedRangesDropdown);
+            // Note: the selector to toggle collapse should be the one with "collapse" class.
+            // $(`#${containerId}`).collapse("show");
```
rfrenchseti: What is this for?

opus/application/static_media/js/search.js
```diff
@@ -41,41 +41,76 @@ var o_search = {
     slugEndpointsReqno: {},
     slugRangeInputValidValueFromLastSearch: {},
 
+    rangesNameMatchedCounter: {},
+    rangesNameTotalMatchedCounter: 0,
+    // Use to determine if we should automatically expand/collapse ranges info. If it's set
+    // to true, we will automatically expand/collapse ranges info depending on the matched letters.
+    isTriggeredFromInput: false,
+    performInputValidation: true,
```
rfrenchseti, line 56: This variable could use a comment too

opus/application/static_media/js/widgets.js
```diff
@@ -18,6 +18,7 @@ var o_widgets = {
      **/
 
     lastStringSearchRequestNo: 0,
+    isKeepingRangesDropdownOpen: false,
```
rfrenchseti, line 22: I think this variable could use a comment. The name isn't clear to me.

opus/application/static_media/js/widgets.js
```diff
@@ -583,12 +692,40 @@ var o_widgets = {
...
+    },
+
+    alignRangesDataByDecimalPoint: function(widget) {
+        /**
+         * Align the data of ranges info by decimal point.
+         */
+        let preprogrammedRangesInfo = $(`#${widget} .op-scrollable-menu li`);
+        for (const category of preprogrammedRangesInfo) {
+            let collapsibleContainerId = $(category).data("category");
+            let rangesInfoInEachCategory = $(`#${collapsibleContainerId} .op-preprogrammed-ranges-data-item`);
```
rfrenchseti: Probably should be called rangesInfoInOneCategory

rfrenchseti (changes requested):
- Highlight colors are inconsistent. When you hover over a string completion hint you see purple, but when you hover over a range completion hint you see dark grey. I also note that the string completion hint moves a little side to side when you hover over it. I think grey is probably the right answer in both cases.
- I also noticed that if you have the string choice list open and resize the browser, the dropdown becomes detached from the input field.

rfrenchseti:
Also the CRC colors should be reordered to go in increasing numerical order (blue to red) to make them consistent with the other two categories.

rfrenchseti:
If you type in something that could be a partial match (like "Thebe") and then hit enter, you get the red field and ?? everywhere as expected. But then if you switch to the browse tab you get the spinner and it never goes away, even when you switch back to the search tab. In other cases of illegal values when you switch to the browse tab you just get a blank gallery.

juzen2003:
- Update based on comments.
- Update the layout of ranges dropdown.
- Customize jquery ui autocomplete menu item hover background color to grey.
- Remove spinner and left gallery view blank when dataimages api has an error. (happened in both master and this branch)
- Fix the issue of having invalid search when the user focused out a string input with empty input. (happened in both master and this branch)
- Put CRC Wavelength data in increasing numerical order.
- Make jquery ui autocomplete dropdown to stay right below string input when browser is resized.

rfrenchseti:
- The string autocomplete still wiggles back and forth as you mouse over different selections
- The ranges dropdown starts narrow with all categories closed, and then widens when you open a category. It would be much better if it stayed the same width all the time.
- The location of the Max and Min headers in each category does not line up with the right side of the numbers. It's like there's always space for a decimal portion even when it's not there.
- The wavelength dropdown gets too narrow and lines start to wrap as the browser shrinks, even though there is still plenty of room in the search widget.

opus/application/static_media/js/search.js
```diff
@@ -41,41 +41,82 @@ var o_search = {
...
+                $(this).addClass("search_input_invalid");
             }
             $(this).addClass("input_currently_focused");
             $(this).removeClass("search_input_invalid_no_focus");
+
+            // Open the dropdown properly when user tabs to focus in.
+            let slugName = $(this).data("slugname");
+            let inputToTriggerDropdown = $(`#widget__${slugName} input.op-range-input-min`);
+            let preprogrammedRangesDropdown = $(`#widget__${slugName} .op-scrollable-menu`);
+            if (preprogrammedRangesDropdown.length !== 0 && $(e.target).hasClass("op-range-input-min")) {
```
rfrenchseti: Shouldn't this all just be a single if with clauses connected by &&?

rfrenchseti (changes requested):
Here are the currently remaining issues:

- Wrapping of range names happens too soon - they shouldn't wrap if there is still a lot of whitespace left on the line.
- search.js line 77 if combine into a single if
- This still happens: "If you type in something that could be a partial match (like "Thebe") and then hit enter, you get the red field and ?? everywhere as expected. But then if you switch to the browse tab you get the spinner and it never goes away, even when you switch back to the search tab. In other cases of illegal values when you switch to the browse tab you just get a blank gallery." What's worse you actually get a yellow "Message" in the top telling you there's an illegal value.
- This is still true in forms.py: "OK, I see what's going on here - since the min/max is also a class name. That does seem kind of like a bad class name, though. How about prepending something like "op-range-input-" to the "min"/"max" to make a better class name?"
- This question was not answered: "What happens with "intcomma" when we have non-int/float field values, like date/time strings?"
- The current JSON format doesn't actually support hint fields for the "max" field because it relies on there being a "slug2" column in the database.

juzen2003:
- Customize ring radius and wavelength dropdown menus separately.
    - Names in wavelength dropdown names will never wrap because they are all short words.
    - Names in ring radius dropdown names will not wrap when there are still lots of spaces between names and min data.
- Combine 3 if blocks into a single if block using &&.
- Change getJSON to ajax for dataimages api call and add error handler to remove the spinner. Now we will get a blank page (spinner is removed) when the api call has an error.
- For ranges inputs, input.min is changed to input.op-range-input-min, and input.max is chagned to input.op-range-input-max.
- For intcomma, if we pass in a string without any number, it will do nothing. It only converts the number part.
    - For example: "1000000-Test-String" will be filtered to "1,000,000-Test-String", it will convert the number part, and do nothing to rest of the string.
- Now the design will support hints for both min/max inputs. We put values to "pi_field_hints1" and "pi_field_hints2" under "slug" (or "slug1") column in the database.

rfrenchseti (changes requested):
There are still problems with illegal search values:
- Type "thebe" into ring radius min and hit enter
- Switch to the browse tab
- Notice the hash now has "RINGGEOringradius1=thebe" in it. We should never allow illegal values in the hash.
- If you refresh the page now the illegal slug value will get passed to the backend and a yellow warning message will be shown.
- Go back, clear everything out and start over; then end "thebe" again
- Go to the browse results page and notice that you can drag the slider
- There are still cases where you go to the browse results tab with an illegal search value and it gives you a spinner

I think looking for the error from the AJAX call is the wrong way to handle this. We should already know that the input is invalid, right? Although there's probably an annoying race condition in there.

- The dropdown box is just a little too wide. If there's a vertical scrollbar in the search widget column, the dropdown will partially cover it.

rfrenchseti (approved):

PR #849, opened 2019-09-04
Initial version to add the ability to right-click on (i) info to open…

dstopp:
… the detail in a new tab.  Mini-menu still not working correctly.


opus/application/static_media/js/browse.js
```diff
@@ -230,14 +230,15 @@ var o_browse = {
...
-        $('.gallery, .op-data-table').on("click", ".op-tools a", function(e) {
+        $('.gallery, .op-data-table').on("click contextmenu", ".op-tools a", function(e) {
             //snipe the id off of the image..
             let opusId = $(this).parent().data("id");
 
             switch ($(this).data("icon")) {
                 case "info":  // detail page
                     o_browse.hideMenu();
                     o_browse.showDetail(e, opusId);
+                    return; // need to skip the default action to allow the context menu to work
```
rfrenchseti: The break line should go away because it can't be reached after the return.

opus/application/static_media/js/browse.js
```diff
@@ -437,15 +439,19 @@ var o_browse = {
                         o_cart.toggleInCart(fromOpusId, opusId);
                     }
                     break;
+
                 case "info":  // detail page
                     o_browse.showDetail(e, opusId);
+                    return;  // need to not return false to allow the contextmenu to work
```
rfrenchseti: Same for this break.

opus/application/static_media/js/browse.js
```diff
@@ -1155,17 +1162,25 @@ var o_browse = {
             .attr("data-id", opusId);
     },
 
+    getDetailURL: function(opusId) {
+        let tab = opus.getCurrentTab();
+        opus.prefs.detail = opusId;
+        let link = "/opus/#/" +  o_hash.updateHash();
```
rfrenchseti: Do we really have to call updateHash() here, or is there some way to get the current hash that was previously computed? The end of updateHash() sets window.local.hash - can we just use that instead?
Also there's an extra space after the "+"

rfrenchseti (changes requested):
- In addition to Ctrl+click opening a new tab, Shift+click opens a new window. We aren't catching that special case - the new window opens but the current window also switches to the Detail tab.
- When I do a Ctrl+click on the (i), I get TWO new tabs with the Detail view in them.
- The current tooltip for view detail says "(Use CTRL for new tab)" but elsewhere we write that as "Ctrl", so should be consistent here (I know this isn't something you changed).

opus/application/static_media/js/browse.js
dstopp: we need to call updateHash because we have to add the detail= part that comes from changing the opus.prefs.

opus/application/static_media/js/browse.js
dstopp: moved return to end of function

opus/application/static_media/js/browse.js
dstopp: moved return to end of function

opus/application/static_media/js/browse.js
dstopp: modified using hashArray so that the original hash is left untouched.

opus/application/static_media/js/browse.js
```diff
@@ -230,14 +230,16 @@ var o_browse = {
         });
 
         // thumbnail overlay tools
-        $('.gallery, .op-data-table').on("click", ".op-tools a", function(e) {
+        $('.gallery, .op-data-table').on("click contextmenu", ".op-tools a", function(e) {
+            let retValue = false;   // do not use the default handler in this case...
```
rfrenchseti: Which case is "this" case?

opus/application/static_media/js/browse.js
```diff
@@ -437,19 +441,25 @@ var o_browse = {
                         o_cart.toggleInCart(fromOpusId, opusId);
                     }
                     break;
+
                 case "info":  // detail page
                     o_browse.showDetail(e, opusId);
+                     // need to not return false to allow the contextmenu to work
+                     // unless... context menu was used with shift or ctrl key...
+                    retValue = (e.shiftKey || e.ctrlKey) ? false : undefined;
```
rfrenchseti: Does this need to include metaKey? I have no way of testing it without being on a Mac.

opus/application/static_media/js/browse.js
```diff
@@ -470,7 +480,11 @@ var o_browse = {
         });
 
         $(document).on("keydown click", function(e) {
-            o_browse.hideMenu();
+            // don't close the mini-menu on the shift/ctrl key in case the user
+            // is trying to open a new window for detail
+           if (!e.shiftKey && !e.ctrlKey) {
```
rfrenchseti: Does this need to include metaKey?

opus/application/static_media/js/browse.js
```diff
@@ -1155,15 +1170,28 @@ var o_browse = {
             .attr("data-id", opusId);
     },
 
+    getDetailURL: function(opusId) {
+        let tab = opus.getCurrentTab();
```
rfrenchseti: tab is not used below

opus/application/static_media/js/hash.js
```diff
@@ -157,10 +157,10 @@ var o_hash = {
 
     hashArrayToHashString: function(hashArray) {
         let hash = "";
-        for (const param of hashArray) {
-            hash += "&"+param+"="+hashArray[param];
-        }
-        return hash;
+        $.each(hashArray, function(param,value) {
```
rfrenchseti: Add space after param,

rfrenchseti (changes requested):
If you click on "add range to cart" or "remove range from cart" in the hamburger menu, the popup window has #undefined observation names.

rfrenchseti:
- Linux Firefox: Ctrl+click on (i) works properly. Shift+click on (i) opens the Detail in a new TAB (not window) and then also pops up a new window with a brand new OPUS (no search terms, no detail, etc).
- Linux Chrome: Ctrl+click on (i) works properly. Shift+click on (i) works properly. Ctrl+click on hamburger works properly. Shift+click on hamburger works properly.
- Mac Safari: Ctrl+click on (i) works properly. Shift+click on (i) opens the Detail in a new window and ALSO opens another window with a brand new OPUS.

Also the problem with "add range from here" is still present.

rfrenchseti (approved):

PR #850, opened 2019-09-06
EN feedback widget

rfrenchseti:


juzen2003 (commented):
Do we want to add the responsiveness to feedback form and welcome message modal?
  - The feedback form starts to get cut off when browser height is about 605px or less. 
  - Same as the welcome modal which starts to get cuff off when browser height is about 575px or less.

rfrenchseti:
True. I think the right thing to do is ask the original creators of the feedback widget to do something to make it responsive. I'll see what they say.

We should probably do something about the welcome message. I'll create a separate issue for that.

PR #860, opened 2019-09-28
-add button for twitter feed

dstopp:
-added a twitter button in the footer for the @pdsopus feed


rfrenchseti:
This doesn't look very aesthetically pleasing to me, but I can't figure out how to change it yet.

PR #863, opened 2019-10-08
Support multiple searches #582

### LINKED ISSUE #582: Support searching on multiple values
It would be nice to be able to search on multiple values. For example, for strings you could look for two different volume names. For a numeric range you could like for "A->B" OR "C->D". This could be a generic feature where you can add duplicate copies of fields in a widget that are ORed together.

juzen2003:
- Status updates on implementation of multiple searches:
    - Fixes #582 
    - Changes tested on latest Chrome, FF, Safari, Opera.
    - Changes verified with JSHINT: current version of opus.js, hash.js, search.js, widgets.js

- Detailed work status:
    - Finished items:
        - updateHash, convert between URL and opus.selections and opus.extras (V)
            - convert from URL hash to opus.selections and opus.extras
                - getSelectionsExtrasFromHash
                - initFromHash
            - convert from opus.selections and opus.extras to URL hash
                - updateHash
            - Update opus.selections and opus.extras and updateHash
                - input.RANGE change event handler function 
                - input.STRING change event handler function
                - select change event handler function
                - validateRangeInput
                - click "+/OR" button event handler function
                - click trash can button event handler function
                - fillRangesInputs
                - STRING autocomplete source callback function
        - Properly renumber RANGE & STRING inputs. (V)
        - Add/remove removeInputIcon buttons & "OR" labels. (V)
            - When getting multiple inputs from pasting an URL.
            - When getting/removing multiple inputs from clicking "+/OR" or removeInputIcon icon.
        - Make "+/OR" button always stay at the right of the last set of inputs. (V)
        - When clicking "+/OR", we clone the first set of DOM, clear the clone value, and set qtype to default value "any". (V)
        - Update opus.selections, opus.extras, and updateHash after clicking trash can or "+/OR" buttons. (V)
        - Update areObjectsEqual function so that the order of objects' key will not matter when comparing two objects. (V)
        - Tune the position of STRING input remove button. (V)
        - Display empty set of inputs in one widget when only qtype value is in URL hash. (V)
        - Update opus.selections when selecting items from preprogrammed ranges. (V)
        - Make sure preprogrammed ranges info is available for multiple inputs in one widget. (V)
        - Make first input set removable if there are multiple sets of inputs and hide remove icon if there is only one set left. (V)
        - Modify logic in updateHash to update hash in URL to have the similar format like the one from normalized url. (V)
        - Limit max inputs per widget to 10. (V)
        - Open up a range dropdown by focus in, and then click "+ (OR)" or trash icon to leave focus, the dropdown will not close. (V)
        - RACE CONDITIONS:
            Potential race conditions in existing code:
            - Fix the followings by disabling "x" button when normalizeinput is in progress: (V)
                - Type a character, then close the widget before normalizeinput (for that character) returns. 
                - Leave focus, then close the widget before normalizeinput (for all fields) returns. 
            In new code: (Need to figure out better UI)
            - Fix the followings by disabling trash icon when normalizeinput is in progress: (V)
                - Type a character, then delete the search line before normalizeinput (for that character) returns; note this will ALSO activate the next case on the focus out because of the trashcan click. 
                - Leave focus, then delete the search line before normalizeinput (for all fields) returns. 
                - Delete a search line, then delete another search line while the first search is updating. 
        - Move trash icon position to the right of (i) icon. (V)
        - “:” in date-time strings doesn’t need to be encoded and shouldn’t be replaced with %3A. (V)
        - Update opus.selections when there is a widget with range input but without q-type dropdown. (V)
        - When you click on “+/OR” for Observation Duration, you get a yellow message about there being a qtype for a field  that doesn’t allow qtypes. (V)
        - Fix issue about focusing out/in between different inputs with preprogrammed range info. (V)
        - Fix autocomplete issue on multiple string search inputs. (V)
            - Init autocomplete for each input set when getWidget is called.
            - Re-init autocomplete for each input when "+ (OR)" or trash icon is clicked due to renumber.
            - Attach autocomplete dropdown position to its corresponding input when window is resized.
        - Make "--OR--" centered based on the input fields.(V)
        - When normalize input is in progress, disable "+ (OR)" button as well. Because if slugs get renumbered and data from normalize input returns after that, then it will mess up the selector to the corresponding input based on the returned data. (V) 
        - When first input has an invalid value, make sure the added cloned input has those invalid borders removed. (V)
        - click trash icon on the input set with invalid value (red background). URL is not updated, selections is obtained and compare with opus.selections to cause a reload. (V)
        - When there are multiple inputs, say 2 sets of inputs in a widget, and the user quickly inputs 2 valid values in 2 input fields and 1 invalid value in the 3rd field: (V)
            - If opus.selections is updated with an invalid value in a different input when we are iterating through previous valid normalizedInput returned data in validateRangeInput, then opus.allInputsValid is true and updateHash will update URL with invalid valid from opus.selections.
            - If we update opus.selections after normalize input api return, when there are multiple normalize input call, the url used to call later normalize input api might not be up-to-date because opus.selections is still waiting for the return of previous normalize input api call.
            - Fixed by:
                - Create a function to determine if normalize input api call is in progress
                - For each normalize input API call, updateHash in the return of API will update URL based on the same selections used to create the URL for API call.


rfrenchseti:
A few initial comments:
- normalizeurl returns search terms as `slug1_01=x&slug2_01=x ... slug1_02=x&slug2_02=x ...` but the OPUS JS hash routines set the terms as `slug1_01=x&slug1_02=x ... slug2_01=x&slug2_02=x`. I like the way normalizeurl does it better because it keeps the search terms for a given line together. The fact that these are different is causing the browser to refresh at odd times.
- You should be able to "trash" the first line along with the other ones. You just can't trash the ONLY line when there's just one.
- The "OR" between lines should be a lot prettier and stylized somehow so it doesn't look like just plain text.
- The trashcan icon is too close to the left edge of the widget.

rfrenchseti:
I think all the const...in should be const...of

juzen2003:
- Update and complete items in the list (in the 1st comment)
- The 4 places of using `for (const...in...)` is to loop through keys of an object.

requirements-python3.txt
```diff
@@ -2,7 +2,7 @@ certifi==2019.9.11
 chardet==3.0.4
 colorclass==2.2.0
 coverage==4.5.4
-Django==2.2.6
+Django==2.1.5
```
rfrenchseti: Let's go ahead and leave the non-Mac requirements alone because the new version of Django should be used on the real server, which runs Linux.

requirements-python3-mac.txt
```diff
@@ -2,7 +2,7 @@ certifi==2019.9.11
 chardet==3.0.4
 colorclass==2.2.0
 coverage==4.5.4
-Django==2.2.6
+Django==2.1.5
```
rfrenchseti, line 8: You can put comments in a requirements file with #. Let's put a comment on this line saying why you had to back off the version.

opus/application/apps/ui/views.py
```diff
@@ -143,8 +143,13 @@ def api_get_widget(request, **kwargs):
     extras = {}
 
     if request and request.GET is not None:
-        (selections, extras) = url_to_search_params(request.GET,
-                                                    allow_errors=True)
+        if form_type in settings.RANGE_FORM_TYPES or form_type == 'STRING':
```
rfrenchseti: I don't see any reason to have this "if" here. You can just always pass `allow_empty=True` because that flag doesn't do anything if it's not a RANGE or STRING type anyway.

opus/application/apps/ui/views.py
```diff
@@ -189,7 +194,9 @@ def api_get_widget(request, **kwargs):
         length = len1 if len1 > len2 else len2
 
         if not length: # param is not constrained
-            form = str(SearchForm(form_vals, auto_id=auto_id).as_ul());
+            form = '<ul class="op-search-inputs-set">' + \
```
rfrenchseti: Using the backslash line continuation is generally considered bad form in Python. Please just put the whole expression in () instead.

opus/application/apps/ui/views.py
```diff
@@ -215,10 +222,16 @@ def api_get_widget(request, **kwargs):
                 except (IndexError, KeyError, ValueError, TypeError) as e:
                     form_vals[slug2] = None
 
-                form = form + str(SearchForm(form_vals, auto_id=auto_id).as_ul())
-
-                if length > 1:
-                    form = form + '</span><div style="clear: both;"></div></section><section><span class="widget_form">'
+                if key == 0:
+                    form = form + \
```
rfrenchseti: Same thing for backslash here. Also might as well use +=.
Since the two if/else expressions are mostly the same except for the addition of op-extra-search-inputs, it would probably be cleaner to do something like:
```
extra_str = ''
if key != 0:
    extra_str = 'op-extra-search-inputs '
```
and then just use that string inside the expression.
Finally, try to limit Python lines to 80 characters. I know the rest of this file violates that, because it hasn't be reformatted, but we might as well try with new code.

opus/application/apps/ui/views.py
```diff
@@ -227,10 +240,21 @@ def api_get_widget(request, **kwargs):
             key = 0
             for value in selections[param_qualified_name]:
                 form_vals[slug] = value
-                form = form + str(SearchForm(form_vals, auto_id=auto_id).as_ul())
+                if key == 0:
```
rfrenchseti: Same as the above here.

opus/application/apps/search/views.py
```diff
@@ -651,10 +651,20 @@ def url_to_search_params(request_get, allow_errors=False, return_slugs=False,
                 # If both values are None, then don't include this slug at all
                 if new_param_qualified_names[0] not in selections:
                     selections[new_param_qualified_names[0]] = []
-                selections[new_param_qualified_names[0]].append(new_values[0])
+
```
rfrenchseti, line 654: Python's list insert doesn't work in the way you expect when inserting off the end of a list. The lists need to be padded to be the proper length.

opus/application/static_media/js/hash.js
```diff
@@ -272,14 +527,131 @@ var o_hash = {
...
+                while (selections[slug].length < longestLength) {
+                    selections[slug].push(null);
+                }
+            }
+        }
+
+        return [selections, extras];
+    },
+
+    isDateTimeSlug: function(slug) {
```
rfrenchseti: 1) This is a bad way to do this. We shouldn't be encoding specific information about fields in the JS. It should be passed in from the Django as an attribute on the widget or something.
2) Why do you have to do this at all? Why can't we just allow : to be unencoded all the time? I don't see why date/time fields should be any different.

opus/application/static_media/js/hash.js
```diff
@@ -16,20 +16,125 @@ var o_hash = {
...
+        let visited = {};
+        $.each(selections, function(key, value) {
+            if (visited[key]) {
+                return; // continue
+            }
+
             if (value.length) {
                 let encodedSelectionValues = o_hash.encodeSlugValues(value);
-                hash.push(key + "=" + encodedSelectionValues.join(","));
+                if (o_hash.isDateTimeSlug(key)) {
```
rfrenchseti: I don't see why this is necessary. Why not just let ":" be unencoded all the time?
Also, this would be easier to write as `encodedSelectionValues =  o_hash.encodeSlugValues(value, o_hash.isDateTimeSlug(key))`

opus/application/static_media/js/hash.js
```diff
@@ -16,20 +16,125 @@ var o_hash = {
...
+                    let anotherKey = key.match(/.*1/) ? `${slugNoNum}2` : `${slugNoNum}1`;
+                    let anotherEncodedSelectionValues = o_hash.encodeSlugValues(selections[anotherKey]);
+                    if (o_hash.isDateTimeSlug(anotherKey)) {
+                        anotherEncodedSelectionValues = o_hash.encodeSlugValues(selections[anotherKey], true);
+                    }
+                    visited[key] = true;
+                    visited[anotherKey] = true;
+
+                    for(let trailingCounter = 1; trailingCounter <= numberOfInputSets; trailingCounter++) {
+                        let trailingCounterString = (`${trailingCounter}`.length === 1 ?
```
rfrenchseti: Here's a better way to do this:
http://www.codigomanso.com/en/2010/07/simple-javascript-formatting-zero-padding/

opus/application/static_media/js/hash.js
```diff
@@ -16,20 +16,125 @@ var o_hash = {
...
+            }
+
             if (value.length) {
                 let encodedSelectionValues = o_hash.encodeSlugValues(value);
-                hash.push(key + "=" + encodedSelectionValues.join(","));
+                if (o_hash.isDateTimeSlug(key)) {
+                    encodedSelectionValues = o_hash.encodeSlugValues(value, true);
+                }
+                let numberOfInputSets = encodedSelectionValues.length;
+                let slugNoNum = key.match(/.*(1|2)/) ? key.match(/(.*)[1|2]/)[1] : key;
```
rfrenchseti: All of these regular expressions do the wrong thing if there's a "1" or "2" in the middle of a slug (not that we have that right now, but we could conceivably some day). You need to add a "$" to the end of the regex to lock it to the end of the string like `/.*(1|2)$/`.

opus/application/static_media/js/hash.js
```diff
@@ -16,20 +16,125 @@ var o_hash = {
...
         });
 
         $.each(opus.extras, function(key, value) {
             if (value.length) {
                 let encodedExtraValues = o_hash.encodeSlugValues(value);
-                hash.push(key + "=" + encodedExtraValues.join(","));
+                if (o_hash.isDateTimeSlug(key)) {
+                    encodedExtraValues = o_hash.encodeSlugValues(value, true);
+                }
+                if (value.length > 1) {
```
rfrenchseti: I was trying to figure out what the point of all this was, and then I realized that you were actually emulating the behavior of a *bug* in normalizeurl! The output of normalizeurl currently puts "singular" qtypes at the end, so you'd get a slug order like:

```
wavelength1_01=X
wavelength2_01=X
qtype-wavelength_01=any
wavelength1_03=X
wavelength1_03=X
qtype-wavelength_03=any
qtype-wavelength_02=any
```

That qtype_02 at the end is out of place. It should be interleaved in numerical order with the other slugs. But it's definitely what normalizeurl does right now.

The correct thing to do is fix the bug in normalizeurl. I have fixed it and you can pull from my branch "fix_normalizeurl". With this fix, clauses are always in numerical order, so now you'll get:

```
wavelength1_01=X
wavelength2_01=X
qtype-wavelength_01=any
qtype-wavelength_02=any
wavelength1_03=X
wavelength1_03=X
qtype-wavelength_03=any
```

The search slugs are also always in alphabetical order.

You'll have to modify this functional accordingly to insert the "orphan" qtypes at the right place.

rfrenchseti (changes requested):
- When there is an illegal input, the yellow or red border makes the input field slightly wider, pushing the rest of the line to the right. This means that if the Min field is in error, that the Max fields don't line up, and if either the Min or Max fields are in error, then the trash cans don't line up.
- It is way more annoying than I thought it would be to not be able to click on "+OR" directly after typing a value. This is a normal thing I would do. Click on the Min box. Type a value. Click on the Max box. Type a value. Click on +OR. But that doesn't work. Is there any other way to fix this?
- For fields that don't have a qtype (like Observation Duration), there's no record of an empty line being present, which means that if you refresh the browser the empty lines go away. I don't think we have to change this, just pointing it out. Actually for fields that have units, this will be fixed when units get implemented, since we will have default unit slugs just like we have default qtype slugs.
- Adding a blank search line with +OR triggers a new search. There's no reason for it to. The comparison of opus.selections should ignore blank clauses. Same for deleting a blank clause. Note when determining if a clause is blank, we don't care about the qtype (or units eventually) just the selection values.
- When you open the ring radius widget and type something like "A Ring", then click on the line for the A Ring in the dropdown, the Min box turns red until the validate inputs returns. The current version doesn't do that. If there are multiple clauses, all of the other Mins turn red as well.
- Sometimes you can get in a state where whatever box you put focus in turns red, even though it has a good value. This happens when there is one clause with an illegal value, and then in a different clause you type in the name of a ring and click on it.
- I also find cases when I add a new clause and then quickly start typing a ring name in the new clause's Min box, and the dropdown has the wrong string matching in it. It appears to be using the string match from an earlier clause. But if I wait a moment and then start typing, it works OK.
- The "--- OR ---" is centered under the Min/Max when there's no qtype, but is not centered when there is a qtype.
- Likewise I think it would look nicer for the "--- OR ---" to be centered under the string input field. Right now it's centered under the input field plus the qtype.

opus/application/static_media/js/hash.js
```diff
@@ -183,21 +319,121 @@ var o_hash = {
...
+                    // so that page will not reload in these two cases.
+                    if (inputInAWidget.hasClass("RANGE")) {
+                        if (!(`${widgetSlug}1` in selections) && !(`${widgetSlug}2` in selections) &&
+                            !(`${widgetSlug}1` in opus.selections) && !(`${widgetSlug}2` in opus.selections)) {
+                            selections[`${widgetSlug}1`] = [null];
+                            selections[`${widgetSlug}2`] = [null];
+                            opus.selections[`${widgetSlug}1`] = [null];
+                            opus.selections[`${widgetSlug}2`] = [null];
+                        } else {
+                            selections[`${widgetSlug}1`] = (selections[`${widgetSlug}1`] ?
```
rfrenchseti: Check out the JavaScript || and && operators. They do what's called "shortcutting". You should be able to say: `x = x || y`. If x evaluates to true, then x is assigned to x. If it evaluates to false, then x=y. So you can say:
```
selections[`${widgetSlug}2`] = selections[`${widgetSlug}2`] || [];
```
I think. This will make a lot of this code simpler. All see various functions later in this file for the same paradigm.

rfrenchseti (changes requested):
One more comment on hash.js while you're working on it.

opus/application/static_media/js/opus.js
```diff
@@ -131,6 +150,11 @@ var opus = {
 
         let [selections, extras] = o_hash.getSelectionsExtrasFromHash();
 
+        // Align data in opus.selections and opus.extras to make sure empty
+        // inputs will also have null in opus.selections
+        [opus.selections, opus.extras] = (o_hash.alignDataInSelectionsAndExtras(opus.selections,
```
rfrenchseti: The outside set of () is not needed

opus/application/static_media/js/opus.js
```diff
@@ -238,10 +278,11 @@ var opus = {
          * Given the result of the search parameter normalization, execute the search
          * that will eventually return the result count.
          */
-
         // If there are more normalized data requests in the queue, don't trigger
         // spurious result counts that we won't use anyway
         if (normalizedData.reqno < opus.lastAllNormalizeRequestNo) {
+            opus.normalizeInputForAllFieldsInProgress[opus.allSlug] = false;
+            o_widgets.disableButtonsInAWidget(false);
```
rfrenchseti: Should this be called disableButtonsInWidgets since it applies to all widgets, not a single widget?

opus/application/static_media/js/opus.js
```diff
@@ -1080,7 +1125,44 @@ var opus = {
             $.cookie("visited", opus.splashVersion, {expires: 1000000});
             opus.displaySplashDialog();
         }
-    }
+    },
+
+    // The following two functions are used in multiple files: search.js,
```
rfrenchseti: Maybe these two functions below belong in utils?

opus/application/static_media/js/opus.js
```diff
@@ -44,7 +44,23 @@ var opus = {
     // avoiding race conditions in ajax calls
     lastAllNormalizeRequestNo: 0,
     lastResultCountRequestNo: 0,
-    waitingForAllNormalizedAPI: false,
+
+    // An object stores the input (slug) & boolean value to indicate whether the normalize input
+    // API (for all fields) is in progress:
+    // Key: input slug that triggers the normalize input API call. (Use opus.allSlug for API from getResultCount)
+    // Value: true & false to indicate whether the API is in progress.
+    normalizeInputForAllFieldsInProgress: {},
```
rfrenchseti, line 52: Rather than allowing the field to be true or false, which gives complicated code like `isAnyNormalizeInputInProgress` below, would it be better to have a slug value be present or not present? Then to check if any normalize is in progress, you can just see if this object has the slug as a property.

Are there any properties? `Object.keys(normalizeInputForAllFieldsInProgress).length > 0`
Is there a particular property? `if (normalizeInputForAllFieldsInProgress[slug])`
Set a particular property? `normalizeInputForAllFieldsInProgress[slug] = true;`
Remove a particular property? `delete normalizeInputForAllFieldsInProgress[slug];`

I'm not sure if this is better or cleaner - I'll let you decide.

opus/application/static_media/js/widgets.js
```diff
@@ -85,17 +111,249 @@ var o_widgets = {
...
+        /**
+         * Add event handlers for click events on "+ (OR)" and trash icons.
+         * These are behaviors of adding/removing inputs in a single widget.
+         */
+
+        // Create a new set of inputs when clicking the "+ (OR)" button in a widget.
+        $("#search").on("click", ".op-add-inputs-btn", function(e) {
+            e.preventDefault();
+            // When normalize input api is in progress, we have to disable "+ (OR)" so
+            // that no renumber will not happen. This will make sure the corresponding
```
rfrenchseti: "no renumber will not happen" is redundant.
Also isn't it OK for +OR to happen? There's no renumber in this case and it would solve the problem I mentioned earlier.

opus/application/static_media/js/widgets.js
```diff
@@ -85,17 +111,249 @@ var o_widgets = {
...
+                          '<hr class="op-or-label-divider"></ul>';
+
+            let removeInputIcon = '<li class="op-remove-inputs">' +
+                                  '<button type="button" title="Delete this set of search inputs" \
+                                  class="p-0 btn btn-small btn-link op-remove-inputs-btn"' +
+                                  `data-widget="widget__${slug}" data-slug="${slug}">` +
+                                  `<i class="${trashIcon}"></i></button></li>`;
+            // If first input set doesn't have remove icon, we add remove icon to the first
+            // input set, and add or label & remove icon to cloned input set, else we only
+            // or label to cloned input set.
```
rfrenchseti: "only or label"?

opus/application/static_media/js/widgets.js
```diff
@@ -85,17 +111,249 @@ var o_widgets = {
...
+
+            let numberOfInputSets = $(`#widget__${slug} .op-search-inputs-set`).length;
+            o_widgets.attachAddInputIcon(slug, addInputIcon);
+
+            // Update opus.selections & opus.extras
+            let newlyAddedInput = $(`#widget__${slug} .op-search-inputs-set input`).last();
+            let newlyAddedQtype = $(`#widget__${slug} .op-search-inputs-set select`).last();
+
+            if (newlyAddedInput.hasClass("RANGE")) {
+                opus.selections[`${slug}1`] = opus.selections[`${slug}1`] ? opus.selections[`${slug}1`] : [];
```
rfrenchseti: See earlier comment on use of ||

opus/application/static_media/js/widgets.js
```diff
@@ -85,17 +111,249 @@ var o_widgets = {
...
+                opus.selections[`${slug}2`] = opus.selections[`${slug}2`] ? opus.selections[`${slug}2`] : [];
+                while (opus.selections[`${slug}1`] .length < numberOfInputSets) {
```
rfrenchseti, line 248: Remove extra space before .length and .push

opus/application/static_media/js/widgets.js
```diff
@@ -85,17 +111,249 @@ var o_widgets = {
...
+                while (opus.selections[`${slug}2`] .length < numberOfInputSets) {
+                    opus.selections[`${slug}2`] .push(null);
+                }
+
+                if (newlyAddedQtype.length > 0) {
+                    opus.extras[`qtype-${slug}`].push(defaultQtypeVal);
+                }
+            } else if (newlyAddedInput.hasClass("STRING")) {
+                opus.selections[slug] = opus.selections[slug] ? opus.selections[slug] : [];
+                while (opus.selections[slug] .length < numberOfInputSets) {
```
rfrenchseti: Remove extra space before .length and .push

opus/application/static_media/js/widgets.js
```diff
@@ -495,24 +787,38 @@ var o_widgets = {
             // This will also put qtype in the url when a widget with qtype is open.
             // Need to wait until api return to determine if the widget has qtype selections
             let hash = o_hash.getHashArray();
+
+            // NOTE: inputs & qtypes are not renumnered yet at this stage.
```
rfrenchseti: renumbered

opus/application/static_media/js/widgets.js
```diff
@@ -696,13 +866,215 @@ var o_widgets = {
...
+                // Renumber max inputs.
+                o_widgets.renumberAttributesOfSearchElements(maxRangeInputs);
+
+                // Renumber preprogrammed ranges dropdown.
+                trailingCounter = 0;
+                if (preprogrammedRangesInfo.length > 1) {
+                    for (const eachRangeDropdown of preprogrammedRangesInfo) {
+                        let rangesDropdownCategories = $(eachRangeDropdown).find("li");
+                        trailingCounter++;
+                        trailingCounterString = (`${trailingCounter}`.length === 1 ?
```
rfrenchseti: See earlier comment on how to zero-pad a number.
Actually given that this happens in multiple places it should really be a function in utils.

opus/application/static_media/js/widgets.js
```diff
@@ -803,14 +1175,194 @@ var o_widgets = {
...
+                    opus.selections[slug] = [currentValue];
+                }
+
+                let newHash = o_hash.updateHash(false);
+                /*
+                We are relying on URL order now to parse and get slugs before "&view" in the URL
+                Opus will rewrite the URL when a URL is pasted, all the search related slugs would be moved ahead of "&view"
+                Refer to hash.js getSelectionsFromHash and updateHash functions
+                */
+                let regexForHashWithSearchParams = /(.*)&view/;
```
rfrenchseti: Actually I don't think this is true anymore, is it? view= appears near the end of the URL. I think cols= comes first now. This appears in multiple places throughout OPUS.

opus/application/static_media/js/search.js
```diff
@@ -664,24 +724,32 @@ var o_search = {
...
 
     parseFinalNormalizedInputDataAndUpdateHash: function(slug, url) {
+        /**
+         * Parse the return data from a normalize input API call. validateRangeInput
+         * is called here.
+         */
         $.getJSON(url, function(normalizedInputData) {
             // Make sure it's the final call before parsing normalizedInputData
             if (normalizedInputData.reqno < o_search.slugNormalizeReqno[slug]) {
+            // if (normalizedInputData.reqno < o_search.lastSlugNormalizeRequestNo) {
```
rfrenchseti: Remove comment?

rfrenchseti (changes requested):

rfrenchseti (changes requested):
One more bug. Activate the ring radius field. Add a second line of inputs. Click in the min box. Type "a". Scroll down using the mouse wheel. If you scroll down past the end of the input box, the whole OPUS window will scroll up, causing the range dropdown to close. This doesn't happen on the first input line.

rfrenchseti:
Also a reminder here that we will need to update the Getting Started Guide with info on using multiple searches.

juzen2003:
- Pending items:
    - I also find cases when I add a new clause and then quickly start typing a ring name in the new clause's Min box, and the dropdown has the wrong string matching in it. It appears to be using the string match from an earlier clause. But if I wait a moment and then start typing, it works OK. 
        - Need to make performInputValidation an object and use slug with id as the key to distinguish different inputs.
    - Center "-- OR --"
- Updates after comments:
    - Fix the string concatenation in python (api_get_widget and url_to_search_params).
    - Clean up logic in api_get_widget and url_to_search_params.
    - Update requirement files.
    - Make sure all ":" is unencoded before creating an Hash for URL
    - Use slice() to create slug trailing counter string 
    - Fix regex to be more specific
    - Use shortcutting to make codes cleaner
    - Modify normalizeInputForAllFieldsInProgress & normalizeInputForCharInProgress to determine if normalize input is in progress by checking if slug (key) is present.
    - Do not disabled "+ (OR)" buttons when normalize input api is in progress
    - Change the regex pattern to get the search params in URL (now cols comes first in URL)
        - re-write updateHash for this will be cleaner?
    - Make input validation border and original input border the same, this will make sure the trash icons (or each input set) are aligned in a widget. 
    - "+ (OR)" button is clickable now when normalize input is in progress. (V)
    - For fields that don't have a qtype (like Observation Duration), there's no record of an empty line being present, which means that if you refresh the browser the empty lines go away. I don't think we have to change this, just pointing it out. Actually for fields that have units, this will be fixed when units get implemented, since we will have default unit slugs just like we have default qtype slugs. (V leave the behavior as is for now)
    - Make the input selector more specific for triggering the change event. 
    - Refactor updateHash into multiple functions
        - getHashStrFromSelections
        - getFullHashStr (call getHashStrFromSelections & getHashStrFromOPUSPrefs)
    -  Fixe the issue that when typing something in the input with dropdown, and select an item while there are multiple matched items, the min box turns red until the validate inputs returns.
    - Do not perform search when adding a new set of inputs or deleting an empty set of inputs.
    - Create a unique id for each input field.
    - Prevent overscrolling for newly added dropdown.
    - Make sure a widget with all input sets empty can have all input sets displayed when page reloads.
    - Prevent red background of input fields with good values when user focuses in and other clause has invalid values.

opus/application/apps/search/views.py
```diff
@@ -651,10 +651,28 @@ def url_to_search_params(request_get, allow_errors=False, return_slugs=False,
...
+                if allow_empty and clause_num_str:
+                    range_min_selection = selections[new_param_qualified_names[0]]
+                    len_min = len(range_min_selection)
+                    if len_min < clause_num:
+                        range_min_selection += [None] * (clause_num - len_min)
```
rfrenchseti: In the general case, this is a dangerous thing to do because the clause_num can be any positive integer. We could be adding 10,000 Nones here. *However*, this never happens because allow_empty is only True when this is called from api_get_widget, and in that case the cause numbers have already been normalized. I think it would be good to put a comment somewhere around here explaining that.

opus/application/static_media/js/utils.js
```diff
@@ -56,6 +71,50 @@ var o_utils = {
...
+         * Takes in a number and left zero padding to a string with the specify
+         * number of digits, by default we want a 2-digits string:
+         * 1 -> "01"
+         * 2 -> "02"
+         * ...
+         * 9 -> "09"
+         * 10 -> "10"
+         */
+        numOfDigits *= -1;
+        return ("0" + num).slice(numOfDigits);
```
rfrenchseti: Easier to just say `slice(-numOfDigits)`. Also " should be '.

opus/application/static_media/js/utils.js
```diff
@@ -56,6 +71,50 @@ var o_utils = {
...
+         * Takes in a slug from normalize input api call and if there is
+         * unique id, return the unique id, else return an empty string.
+         */
+        return (slug.match(/_([0-9]+)_.*$/) ?
+                slug.match(/_([0-9]+)_.*$/)[1] : "");
+    },
+
+    convertToTrailingCounterStr: function(num, numOfDigits=2) {
+        /**
+         * Takes in a number and left zero padding to a string with the specify
```
rfrenchseti: This comment doesn't parse. Maybe you want to say:
Left zero pad the given number to the specified number of digits. By default we want a 2-digit string.

opus/application/apps/search/views.py
```diff
@@ -651,10 +651,28 @@ def url_to_search_params(request_get, allow_errors=False, return_slugs=False,
                 # If both values are None, then don't include this slug at all
                 if new_param_qualified_names[0] not in selections:
                     selections[new_param_qualified_names[0]] = []
-                selections[new_param_qualified_names[0]].append(new_values[0])
+
+                if allow_empty and clause_num_str:
+                    range_min_selection = selections[new_param_qualified_names[0]]
```
rfrenchseti: Please keep lines 80 characters or less

opus/application/apps/search/views.py
```diff
@@ -651,10 +651,28 @@ def url_to_search_params(request_get, allow_errors=False, return_slugs=False,
...
+                if allow_empty and clause_num_str:
+                    range_max_selection = selections[new_param_qualified_names[1]]
+                    len_max = len(range_max_selection)
+                    if len_max < clause_num:
+                        range_max_selection += [None] * (clause_num - len_max)
+                    range_max_selection[clause_num - 1] = new_values[1]
+                else:
+                    selections[new_param_qualified_names[1]].append(new_values[1])
+
                 # There was at least one value added - include the qtype
```
rfrenchseti: Don't we have to do the same thing with the qtypes? Otherwise I think we can have the min/max selections and the qtypes out of sync (and different lengths).

opus/application/apps/search/views.py
```diff
@@ -678,7 +696,16 @@ def url_to_search_params(request_get, allow_errors=False, return_slugs=False,
...
+            if allow_empty and clause_num_str:
+                str_selection = selections[param_qualified_name_no_num]
+                len_s = len(str_selection)
+                if len_s < clause_num:
+                    str_selection += [None] * (clause_num - len_s)
+                str_selection[clause_num - 1] = new_value
+            else:
+                selections[param_qualified_name_no_num].append(new_value)
+
             if param_qualified_name_no_num not in qtypes:
```
rfrenchseti, line 727: Same thing as above for qtypes here.

opus/application/apps/search/views.py
```diff
@@ -651,10 +651,28 @@ def url_to_search_params(request_get, allow_errors=False, return_slugs=False,
                 # If both values are None, then don't include this slug at all
                 if new_param_qualified_names[0] not in selections:
                     selections[new_param_qualified_names[0]] = []
-                selections[new_param_qualified_names[0]].append(new_values[0])
+
+                if allow_empty and clause_num_str:
```
rfrenchseti, line 655: There are currently no test cases that check allow_empty=True along with clause numbers. Please add new tests that check the behavior of this code...with empty clauses that require the insertion of Nones for the min, the max, and the qtype, for both numeric and string fields. That will probably require at least 5 new test functions. They should go in apps/search/test_search.py at some reasonable place (perhaps separated into the range and string sections).

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
      *  changing, reading, and initiating a session from the browser hash
      *
      **/
+    updateHash: function(updateURL=true, searchOnly=false, selections=opus.selections) {
```
rfrenchseti: I thought we were going to get rid of this function entirely? The point was to replace each place that called updateHash with an appropriate call to one of the new functions you wrote below.

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
...
+        /**
+         * updates the hash & URL based on the selections passed in.
+         */
+        // Make sure selections and extras are aligned before updating hash.
+        // This will avoid issue that qtype is not properly updated in the URL hash when a
+        // widget just open and opus.selections is not updated.
+        [selections, opus.extras] = o_hash.alignDataInSelectionsAndExtras(selections, opus.extras);
```
rfrenchseti: This seems problematic to me - a function that might only return a string could actually modify opus.extras. Also, we might be passing in our own selections, but we aren't passing in our own extras. Doesn't that mean the passed-in selections and the global opus.extras could be out of sync? I think this whole function needs to go away anyway, but if this functionality is put elsewhere this needs to be addressed.

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
...
+            hashStr = o_hash.getHashStrFromSelections(selections);
+        }
+        return hashStr;
+    },
 
-    // updates the hash according to user selections
-    updateHash: function(updateURL=true, searchOnly=false) {
+    getHashStrFromSelections: function(selections=opus.selections, forAPICall=false) {
+        /**
+         * Get the hash string from selections, ignore any slug in opus.prefs.
```
rfrenchseti: I'm confused by this comment. Nothing in opus.selections should ever be in opus.prefs, right? Also opus.prefs is never referenced in this function so it doesn't do what it says anyway.

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
...
+            let value = selections[slug];
             if (value.length) {
                 let encodedSelectionValues = o_hash.encodeSlugValues(value);
-                hash.push(key + "=" + encodedSelectionValues.join(","));
+                let numberOfInputSets = encodedSelectionValues.length;
+                let slugNoNum = slug.match(/.*(1|2)$/) ? slug.match(/(.*)[1|2]$/)[1] : slug;
+                let qtypeSlug = `qtype-${slugNoNum}`;
+
+                // If the slug has an array of more than 1 value, and it's either a STRING or RANGE input slug,
+                // we attach the trailing counter string to the slug and assign the corresponding before pushing
```
rfrenchseti: Assign the corresponding...what?

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
...
+                if (slug.match(/.*(1|2)$/)) { // RANGE inputs
+                    let oppositeSuffixSlug = slug.match(/.*1$/) ? `${slugNoNum}2` : `${slugNoNum}1`;
+                    let oppositeSuffixEncodedSelectionValues = o_hash.encodeSlugValues(selections[oppositeSuffixSlug]);
+
+                    let slug1 = slug;
+                    let slug2 = oppositeSuffixSlug;
+                    let slug1EncodedSelections = encodedSelectionValues;
+                    let slug2EncodedSelections = oppositeSuffixEncodedSelectionValues;
+                    if (slug.match(/.*2$/)) {
+                        slug1 = oppositeSuffixSlug;
```
rfrenchseti: Cleaner and easier to read just to say `[slug1, slug2] = [slug2, slug1]` and `[slug1EncodedSelections, slug2EncodedSelections] = [slug2EncodedSelections, slug1EncodedSelections]`

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
...
+
+                    visited[slug1] = true;
+                    visited[slug2] = true;
+
+                    for (let trailingCounter = 1; trailingCounter <= numberOfInputSets; trailingCounter++) {
+                        let trailingCounterString = o_utils.convertToTrailingCounterStr(trailingCounter);
+                        let slug1WithCounter = (numberOfInputSets === 1) ? slug1 : `${slug1}_${trailingCounterString}`;
+                        let slug2WithCounter = (numberOfInputSets === 1) ? slug2 : `${slug2}_${trailingCounterString}`;
+
+                        if (forAPICall) {
```
rfrenchseti: `forAPICall` doesn't really capture what we're doing here - because there are lots of API calls where we don't want to do this substitution. How about something like `useFieldUniqueIDs`? Also put something in the comment for the function explaining what that parameter does.

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
...
+
+    updateURL: function() {
+        /**
+         * Update URL with full hash string
+         */
+        let fullHashStr = o_hash.getFullHashStr();
+        window.location.hash = '/' + fullHashStr;
+    },
+
+    updateURLHashFromExtras: function(hash, qtypeInExtras, numberOfInputSets, counter) {
```
rfrenchseti: Why is this called `updateURLHashFromExtras` when it doesn't actually update the URL?

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
...
+    getFullHashStr: function(selections=opus.selections) {
+        /**
+         * Get the full hash string from search params and opus.prefs
+         */
+        let hashStrFromSelections = o_hash.getHashStrFromSelections(selections);
+        let hashStrFromOPUSPrefs = "&" + o_hash.getHashStrFromOPUSPrefs();
+        return hashStrFromSelections + hashStrFromOPUSPrefs;
+    },
+
+    updateURL: function() {
```
rfrenchseti: How about `updateURLFromCurrentHash` or something like that?

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
...
+                }
             }
-        });
+        }
 
-        $.each(opus.extras, function(key, value) {
+        // For slugs only exist in extras, make sure they are updated in URL.
+        // This will make sure multiple empty input sets can show up after page reloads.
+        for (const qtypeSlug in opus.extras) {
+            if ((qtypeSlug.match(/qtype-(.*)$/)[1] in selections ||
```
rfrenchseti: Put `qtypeSlug.match(/qtype-(.*)$/)[1]` in a variable so it doesn't execute three times. Better yet, if you put the qtypeSlug in `visited` in the two places above you could just check to see if qtypeSlug was in visited here.

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
...
+                    }
+                } else { // Multi/single choice inputs
+                    hash.push(slug + "=" + encodedSelectionValues.join(","));
+                }
             }
-        });
+        }
 
-        $.each(opus.extras, function(key, value) {
+        // For slugs only exist in extras, make sure they are updated in URL.
```
rfrenchseti: "updated in URL" is confusing since we aren't updating a URL here.

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
...
+        // This will make sure multiple empty input sets can show up after page reloads.
+        for (const qtypeSlug in opus.extras) {
```
rfrenchseti: Does this whole thing actually work? The whole point of the sorting of the slugs earlier is to put them in order in the hash. But no we're looking at "lonely" qtypes and putting them at the end, aren't we? I would think these would have to be sorted in along with the other slugs or else they will be out of alphabetical order. This is how I handle it in normalizeurl. I expect you have to do something similar in JS.

```
    # Sort to make tests deterministic and to group qtype with search slugs
    def _sort_func(key):
        clause_num_str = ''
        if '_' in key:
            clause_num_str = key[key.index('_'):]
            key = key[:key.index('_')]
        if key.startswith('qtype-'):
            key = key[6:]+'0' # Add 0 to sort before 1/2 suffixes
        key += clause_num_str
        return key
```

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
...
-            $.each(opus.prefs, function(key, value) {
-                hash.push(key + "=" + value);
-            });
+                if (value.length > 1) {
+                    let numberOfQtypeInputs = encodedExtraValues.length;
 
-            if (updateURL && opus.allInputsValid) {
-                window.location.hash = '/' + hash.join('&');
+                    for(let trailingCounter = 1; trailingCounter <= numberOfQtypeInputs; trailingCounter++) {
+                        let trailingCounterString = (`${trailingCounter}`.length === 1 ?
```
rfrenchseti: This is still using the old zero-padding method.

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
...
+                                                    `0${trailingCounter}` : `${trailingCounter}`);
+                        let newKey = `${qtypeSlug}_${trailingCounterString}`;
+
+                        if (value[trailingCounter-1] !== null) {
+                            // if (!hash.includes(newKey + "=" + encodedExtraValues[trailingCounter-1])) {
```
rfrenchseti: Remove comments here and below.

opus/application/static_media/js/hash.js
```diff
@@ -15,37 +15,211 @@ var o_hash = {
...
+            }
+        } else {
+            hashStr = o_hash.getHashStrFromSelections(selections);
+        }
+        return hashStr;
+    },
 
-    // updates the hash according to user selections
-    updateHash: function(updateURL=true, searchOnly=false) {
+    getHashStrFromSelections: function(selections=opus.selections, forAPICall=false) {
```
rfrenchseti: Don't you need an optional extras argument too?

opus/application/static_media/js/hash.js
```diff
@@ -183,21 +361,117 @@ var o_hash = {
             let value = pair.slice(idxOfFirstEqualSign + 1);
 
             if (!(slug in opus.prefs) && value) {
+                let slugNoCounter = o_utils.getSlugOrDataWithoutCounter(slug);
+                let slugCounter = o_utils.getSlugOrDataTrailingCounterStr(slug);
+                slug = slugNoCounter;
+
+                if (slugCounter > opus.maxAllowedInputSets) {
+                    return; //continue
```
rfrenchseti: Why return here?

opus/application/static_media/js/hash.js
```diff
@@ -183,21 +361,117 @@ var o_hash = {
...
+                    }
+
+                    // Sync up selections and opus.selections when a set of inputs is removed or added
+                    // so that page will not reload in these two cases.
+                    if (inputInAWidget.hasClass("RANGE")) {
+                        if (!(`${widgetSlug}1` in selections) && !(`${widgetSlug}2` in selections) &&
+                            !(`${widgetSlug}1` in opus.selections) && !(`${widgetSlug}2` in opus.selections)) {
+                            selections[`${widgetSlug}1`] = [null];
+                            selections[`${widgetSlug}2`] = [null];
+                            opus.selections[`${widgetSlug}1`] = [null];
```
rfrenchseti: This is another one of those cases where a function with a name like "get..." actually mutates global variables behind our back. There's nothing in the function name that indicates this would happen and it's potentially dangerous. Is there a way to put this process somewhere else and only call it when it's actually needed, making that function name more explicit? Or at least put in a parameter that says whether to do this so it's obvious the caller has to want it?

opus/application/static_media/js/hash.js
```diff
@@ -229,15 +503,34 @@ var o_hash = {
...
                 if (slug.match(/qtype-.*/)) {
-                    // range drop down, add the qtype to the global extras array
-                    let id = slug.match(/qtype-(.*)/)[1];
-                    opus.extras['qtype-' + id] = value.split(',');
+                    let slugNoCounter = o_utils.getSlugOrDataWithoutCounter(slug);
+                    let slugCounter = o_utils.getSlugOrDataTrailingCounterStr(slug);
+                    slug = slugNoCounter;
+
+                    if (slugCounter > opus.maxAllowedInputSets) {
+                        return; //continue;
```
rfrenchseti: Why return?

opus/application/static_media/js/hash.js
```diff
@@ -229,15 +503,34 @@ var o_hash = {
         // just updating prefs here..
```
rfrenchseti: I think the comment about custom widget sizes can be deleted - we don't have those anymore.

opus/application/static_media/js/hash.js
```diff
@@ -272,14 +565,117 @@ var o_hash = {
                     // if ($(`input[name="${slug}"]`).hasClass("STRING")) {
                     //     opus.selections[slug] = [value];
                     // } else {
-                    opus.selections[slug] = value.split(',');
+                    let slugNoCounter = o_utils.getSlugOrDataWithoutCounter(slug);
+                    let slugCounter = o_utils.getSlugOrDataTrailingCounterStr(slug);
+                    slug = slugNoCounter;
+
+                    if (slugCounter > opus.maxAllowedInputSets) {
+                        return; //continue;
```
rfrenchseti: Another return

opus/application/static_media/js/utils.js
```diff
@@ -56,6 +71,50 @@ var o_utils = {
             urlPrefix += `:${window.location.port}`;
         }
         return urlPrefix;
+    },
+
+    // The following two functions are used in multiple files: search.js,
```
rfrenchseti: This comment isn't useful and can get out of date easily

opus/application/static_media/js/utils.js
```diff
@@ -56,6 +71,50 @@ var o_utils = {
...
+    // widgets.js and hash.js.
+    getSlugOrDataWithoutCounter: function(slugOrData) {
+        /**
+         * Takes in a slugOrData from input's name attribute and if there are
+         * multiple inputs, return the version without trailing counter.
+         */
+        return (slugOrData.match(/^([^_]*)_.*/) ?
```
rfrenchseti: Could store results of match in a variable so it doesn't have to be called twice.

opus/application/static_media/js/utils.js
```diff
@@ -56,6 +71,50 @@ var o_utils = {
...
+                slugOrData.match(/^([^_]*)_.*/)[1] : slugOrData);
+    },
+
+    getSlugOrDataTrailingCounterStr: function(slugOrData) {
+        /**
+         * Takes in a slugOrData from input's name attribute and if there are
+         * multiple inputs, return the trailing counter, else return an
+         * empty string.
+         */
+        return (slugOrData.match(/_([0-9]+)$/) ?
```
rfrenchseti: Same thing here

opus/application/static_media/js/utils.js
```diff
@@ -56,6 +71,50 @@ var o_utils = {
...
+                slugOrData.match(/_([0-9]+)$/)[1] : "");
+    },
+
+    getSlugUniqueIdStr: function(slug) {
```
rfrenchseti: See my global comment - this function isn't needed.

rfrenchseti (changes requested):
- js/widgets.js line 1290 still has a regex looking for &cols= instead of using the new routines that don't care about slug ordering.
- Test `test__api_normalizeurl_real_6` is failing.
- I don't understand what you're doing with the unique IDs. The idea was to replace the clause number in normalizeinput with the unique ID, but it appears you are adding them together like "_unique_clause". This doesn't make any sense to the backend, which is only expecting a single "_". It might be working, but only by accident. There's no reason to include the clause number.
- Also, I was thinking there would be one unique ID per input line, not one per input field. It looks to me like you're making a unique ID for each field, which shouldn't be necessary. If you have one per line, then you can map the unique ID to the clause number easily. Maybe this is why you felt like you had to include the clause number too?

```
[16/Nov/2019 14:23:12] DEBUG [tools.app_utils:247] API 96 /opus/__api/normalizeinput.json {
    "observationduration1_103_01": "12.0000",
    "observationduration1_117_04": "5.0000",
    "observationduration2_104_01": "13.0000",
    "observationduration2_118_04": "6.0000",
    "qtype-time_01": "any",
    "qtype-time_02": "any",
    "reqno": "31"
}
```

- Please replace the lines in apps/help/templates/help/gettingstarted.html starting at 91 with:
```
<p>For the latter two types of fields, as you type your search value you will
    receive feedback indicating if the value is legal. A green box means it
    is fine as-is. A yellow box means the value is illegal and you should change
    something before leaving the search field (by clicking elsewhere or hitting
    the Enter key). If you leave the field with an invalid value, it will turn
    red and no search will be performed until you fix the value.</p>

<p>For string, numeric, and date/time fields, you have the option of searching
    on multiple sets of values. You do this by clicking on the "+(OR)" button,
    which will add an additional set of input fields. The multiple input ranges
    are "OR"ed together. For example, if you search on emission angles between
    10 and 80 and also between 100 and 170, your results will contain
    observations that match either set of criteria. You can have up to 10 such
    ranges for each search field.</p>
    
<p>As you enter search values, the number of results that are selected will
    constantly update in the <b>Browse Results</b> tab. In addition,
    you will be provided with the minimum/maximum values or date/times,
    or the number of results for each multiple-choice item, based on your
    current search criteria.</p>

<p>Once you have entered some search constraints, it's time to browse the
    results.</p>
```

rfrenchseti (changes requested):
More bugs:

- If there are no search terms, the "opus.prefs" URL still starts with "&". This means you get a URL like: `/opus/#/&cols=opusid`
- I haven't quite found the minimum set of things to reproduce this problem, but try this:
1) Turn on API delays
2) Start with a fresh OPUS
3) Open Observation Duration
4) Add 4 new empty lines; wait for all the API calls to finish
5) In the last line, min box, type "2". As soon as you go. click on the trashcan for the first line 3 times (removing the first 3 lines). When the validate input call returns, OPUS will refresh the browser.

opus/application/static_media/js/opus.js
```diff
@@ -1080,7 +1128,17 @@ var opus = {
...
             opus.displaySplashDialog();
         }
-    }
+    },
+
+    isAnyNormalizeInputInProgress: function() {
+        /**
+         * Check if any normalize input API call is in progress
+         */
+        // return Object.keys(opus.normalizeInputForAllFieldsInProgress).length > 0;
```
rfrenchseti: Remove comment

rfrenchseti (changes requested):
Another bug:
- Open a string field
- Type something in the field
- Hit +OR
- Type something in the new field
- Exit focus
- OPUS will refresh the browser

rfrenchseti (changes requested):
Another bug:

- Open the ring radius widget
- Type "thebe" and select the "Thebe Ring" with the mouse
- Click on +OR
- Type "a ring". Note in the dropdown box that the A ring in Saturn is not available (and the Saturn category isn't opened). Only the "a ring" in Jupiter is shown.

juzen2003:
- Update based on comments.
- One pending item: 
    - center "-- OR --", fix the responsiveness of "-- OR --"

opus/application/apps/search/views.py
```diff
@@ -651,14 +651,53 @@ def url_to_search_params(request_get, allow_errors=False, return_slugs=False,
...
```
rfrenchseti, line 655: I did a very slight reformat to agree with my coding style in the rest of the Django. Easier just to cut-and-paste the following rather than explain each change:

```
                if allow_empty and clause_num_str:
                    range_min_selection = selections[
                                                new_param_qualified_names[0]]
                    len_min = len(range_min_selection)
                    # Note: clause_num here will not be a very large
                    # number because allow_empty is only True when it's
                    # called from api_get_widget, and in that case,
                    # clause numbers have already been normalized.
                    if len_min < clause_num:
                        range_min_selection += [None] * (clause_num-len_min)
                    range_min_selection[clause_num-1] = new_values[0]
                else:
                    selections[new_param_qualified_names[0]].append(
                                                                new_values[0])

                if new_param_qualified_names[1] not in selections:
                    selections[new_param_qualified_names[1]] = []

                if allow_empty and clause_num_str:
                    range_max_selection = selections[
                                                new_param_qualified_names[1]]
                    len_max = len(range_max_selection)
                    if len_max < clause_num:
                        range_max_selection += [None] * (clause_num-len_max)
                    range_max_selection[clause_num-1] = new_values[1]
                else:
                    selections[new_param_qualified_names[1]].append(
                                                                new_values[1])

                # There was at least one value added or allow_empty is set -
                # include the qtype
                if param_qualified_name_no_num not in qtypes:
                    qtypes[param_qualified_name_no_num] = []

                if allow_empty and clause_num_str:
                    range_qtype = qtypes[param_qualified_name_no_num]
                    len_qtype = len(range_qtype)
                    if len_qtype < clause_num:
                        range_qtype += [None] * (clause_num-len_qtype)
                    range_qtype[clause_num-1] = qtype_val
                else:
                    qtypes[param_qualified_name_no_num].append(qtype_val)
```

opus/application/apps/search/views.py
```diff
@@ -651,14 +651,53 @@ def url_to_search_params(request_get, allow_errors=False, return_slugs=False,
...
+                else:
+                    (selections[new_param_qualified_names[1]]
+                     .append(new_values[1]))
+
                 # There was at least one value added - include the qtype
                 if param_qualified_name_no_num not in qtypes:
                     qtypes[param_qualified_name_no_num] = []
-                qtypes[param_qualified_name_no_num].append(qtype_val)
+
+                if allow_empty and clause_num_str:
```
rfrenchseti, line 689: Same thing here:

```
            if allow_empty and clause_num_str:
                str_selection = selections[param_qualified_name_no_num]
                len_s = len(str_selection)
                if len_s < clause_num:
                    str_selection += [None] * (clause_num-len_s)
                str_selection[clause_num-1] = new_value
            else:
                selections[param_qualified_name_no_num].append(new_value)

            if param_qualified_name_no_num not in qtypes:
                qtypes[param_qualified_name_no_num] = []

            if allow_empty and clause_num_str:
                str_qtype = qtypes[param_qualified_name_no_num]
                len_qtype = len(str_qtype)
                if len_qtype < clause_num:
                    str_qtype += [None] * (clause_num-len_qtype)
                str_qtype[clause_num-1] = qtype_val
            else:
                qtypes[param_qualified_name_no_num].append(qtype_val)
```

opus/application/static_media/js/widgets.js
```diff
@@ -85,17 +104,300 @@ var o_widgets = {
...
+                $(`#${widgetId} input.op-range-input-min[name="${minInputSlug}"]`).dropdown("toggle");
+                $(`#${widgetId} input.RANGE[name="${minInputSlug}"]`).trigger("change");
+
+                let minInput = $(`#${widgetId} input.RANGE[name="${minInputSlug}"]`);
+                let slug = o_utils.getSlugOrDataWithoutCounter(minInputSlug);
+                let slugName = minInput.attr("data-slugname");
+                let uniqueid = minInput.attr("data-uniqueid");
+                let oppositeSuffixSlug = (slug.match(/(.*)1$/) ? `${slugName}2` : `${slugName}1`);
+                $(`#${widgetId} input.RANGE[name*="${oppositeSuffixSlug}"][data-uniqueid="${uniqueid}"]`)
+                .trigger("change");
```
rfrenchseti: I think doing multiple line expressions like this in JS is dangerous because JS will randomly decide the first line should be ended with a virtual ";". Better to just leave on one line.

opus/application/static_media/js/widgets.js
```diff
@@ -85,17 +104,300 @@ var o_widgets = {
...
+            if (defaultQtypeVal) {
+                cloneInputs.find("select").val(defaultQtypeVal);
+            }
+
+            // Assign unique id to new inputs. Inputs in the same set will have the same id.
+            o_widgets.uniqueIdForInputs += 1;
+            cloneInputs.find("input").attr("data-uniqueid", o_widgets.uniqueIdForInputs);
+            cloneInputs.find("input").val("");
+
+            let orLabel = '<ul class="op-or-labels text-secondary">' +
```
rfrenchseti: Likewise probably safest to put () around this expression and the one below

opus/application/static_media/js/widgets.js
```diff
@@ -696,13 +933,219 @@ var o_widgets = {
...
+                // Close dropdown list when ps scrolling is happening in widget container
+                if ($(`#${widget} .op-scrollable-menu`).hasClass("show")) {
+                    // Note: the selector to toggle dropdown should be the one with data-toggle="dropdown"
+                    // or "dropdown-toggle" class, and in this case it's the li (.op-ranges-dropdown-menu).
+                    $(`#${widget} input.op-range-input-min`).dropdown("toggle");
+                }
+            });
+
+            if (widgetInputs.hasClass("RANGE") || widgetInputs.hasClass("STRING")) {
+                let addInputIcon = '<li class="op-add-inputs">' +
```
rfrenchseti: More places for ()

opus/application/static_media/js/hash.js
```diff
@@ -15,49 +15,231 @@ var o_hash = {
...
+                    let oppositeSuffixEncodedSelectionValues = (o_hash.encodeSlugValues(
+                                                                opus.selections[oppositeSuffixSlug]));
+
+                    let slug1 = slug;
+                    let slug2 = oppositeSuffixSlug;
+                    let slug1EncodedSelections = encodedSelectionValues;
+                    let slug2EncodedSelections = oppositeSuffixEncodedSelectionValues;
+                    if (slug.match(/.*2$/)) {
+                        [slug1, slug2] = [slug2, slug1];
+                        [slug1EncodedSelections, slug2EncodedSelections] = ([slug2EncodedSelections,
```
rfrenchseti: () aren't needed here

rfrenchseti (changes requested):
- When calling stringseachchoices, we get all the qtypes for the other clauses for the current string slug, but not their search values:

```
http://127.0.0.1:8000/#/volumeid_01=COISS_2002&qtype-volumeid_01=contains&volumeid_02=C&qtype-volumeid_02=excludes&volumeid_03=COISS_2111&qtype-volumeid_03=contains&volumeid_04=ISS&qtype-volumeid_04=contains

[20/Nov/2019 11:10:40] DEBUG [tools.app_utils:247] API 435 /opus/__api/stringsearchchoices/volumeid.json {
    "qtype-volumeid_01": "contains",
    "qtype-volumeid_02": "excludes",
    "qtype-volumeid_03": "contains",
    "qtype-volumeid_04": "contains",
    "reqno": "13",
    "volumeid_02": "C"
}

I think the correct behavior is to not include the qtypes either. The current behavior is probably harmless, just messy. We definitely don't want to include the search values, because that would then constrain the string choices in ways we don't want it constrained.
```

- Bring up the ring radius search. Enter an invalid value in the "Max" field. Then search for a feature in the "Min" field and click on it. Both min and max will get filled in, but the max field will stay red.

juzen2003:
Center "-- OR --" label by adding dummy items to the right. Dummy items have the same width as qtype select and qtype helper if they exist.

rfrenchseti (approved):

Issue #46, opened 2014-06-25
Implement unit indicators/selectors for numerical input widgets


basilleaf:
Originally reported by: **lisa ballard (Bitbucket: [basilleaf](https://bitbucket.org/basilleaf), GitHub: [basilleaf](https://github.com/basilleaf))**

---

---
- Bitbucket: https://bitbucket.org/ringsnode/opus2/issue/45

rfrenchseti:
Implementation ideas:

- There should be classes of units (e.g. distance, time) and each class should have a set of units that are available (e.g. nm, angstrom, micron, mm, meter, km, AU). The list of units is maintained in Django and served to the JavaScript when a numerical widget is created.
- There should be a default unit for each numerical field, taken from param_info, also served to the JavaScript.
- When displaying a numerical field (or pair of fields), there should be a single dropdown box that has all the possible units in it, and it should be preset to the default unit.
- There should be a new slug type, like q-type, that says which unit to use to interpret the input field. The default unit is used if no slug is specified.
- All unit conversion should be handled by the Python, with no intelligence in the JavaScript.
- **Looking at the Apache logs, some users want to be able to type the unit name after the numerical value like "625 nm" and we should support that, automatically changing the unit selected in the dropbox menu.**

rfrenchseti:
We also need to show the units when displaying metadata, like "observation duration (secs)" or perhaps "observation duration: 45.0 secs".

rfrenchseti:
The initial implementation of this is just going to consist of:

1) Choosing which units we support in the database
2) Putting the proper units in the table_schemas JSON files for all RANGE entries in all obs_* tables
3) Creating a dictionary that converts between the internal database unit abbreviations and the "pretty" versions we want users to see (such as "s" -> "seconds")
3) Modifying api_get_widget() to put the unit type in the widget title
4) Modifying get_metadata() to put the unit type in the metadata name

Later we will allow the user to select the specific units they want to use.

rfrenchseti:
In the table_schemas JSON files, the "data_source" field tells you where the data comes from, and thus where to look to figure out the proper units.
- If it says "FUNCTION", then the second line is the name of a Python function in one of the "populate_xxx" files (sometimes with an instrument or mission name substituted in where obvious, like "obs_pds_<INST>_primary_file_spec" maps to "populate_obs_pds_COISS_primary_file_spec" for COISS volumes). You should look at the function to see if you can figure out what units it's producing.
[...]

PR #891, opened 2019-10-31
Support for units in Django

### LINKED ISSUE #886: Remove trailing zeroes from many places in UI
Currently normalizeinput, range/endpoints, and all the data APIs format numbers with trailing zeros after the decimal point. We should only do this when displaying metadata, and remove it from the input fields and range endpoints.

rfrenchseti:
- Fixes #886 (remove trailing zeros)
- Server-side support for #46 (add units)

Query strings now support units in the same way they support qtypes, including multiple search clauses:

```
http://127.0.0.1:8000/api/meta/result_count.json?observationduration1=10&observationduration2=20
{"data": [{"result_count": 1921}]}

http://127.0.0.1:8000/api/meta/result_count.json?observationduration1=10&observationduration2=20&unit-observationduration=milliseconds
{"data": [{"result_count": 591}]}

http://127.0.0.1:8000/api/meta/result_count.json?observationduration1=10000&observationduration2=20000&unit-observationduration=milliseconds
{"data": [{"result_count": 1921}]}
```

The API supports different units for each clause, but the UI isn't going to. Thus `normalizeurl` enforces that all units are the same.

/api/meta/range/endpoints now takes an optional "?units=" argument and gives the min/max in those units.

/api/fields.json includes the list of available units, and the Help/API Guide displays them. The first unit displayed is the default.


lib/opus_support.py
```diff
@@ -433,16 +435,125 @@ def _range_time_decode(iso):
...
+                'jupiterradii': ('Rs (71492)', 71492.),
+                'neptuneradii': ('Rn (25225)', 25225.),
+                'uranusradii':  ('Ru (25559)', 25559.),
+            }
+        },
+    'km/pixel':
+        {
+            'display_name': 'km/pixel',
+            'conversions': {
+                'm/pixel':      ('m/pixel', 1e3),
```
juzen2003, line 476: 1e-3 for this one.

opus/application/apps/ui/views.py
```diff
@@ -745,6 +765,74 @@ def _sort_func(key):
...
+            if pi.old_slug:
+                handled_slugs.append('unit-'
+                                     +strip_numeric_suffix(pi.old_slug)
+                                     +clause_num_str)
+            handled_slugs.append('unit-'
+                                 +strip_numeric_suffix(pi.slug)
+                                 +clause_num_str)
+            if (valid_units and
+                original_slugs[old_unit_slug+clause_num_str]
+                    not in valid_units):
```
juzen2003, line 797: An extra indentation here? (same at line 740), was comparing to similar styling at line 1149 in results/views.py.

juzen2003 (changes requested):

PR #894, opened 2019-11-14
Fix slider jump. Tighten up margins around numbers in slider.

### LINKED ISSUE #838: Slider jumps on first move action
If you grab the slider and start to move it, it doesn't move from the place where you grabbed it. It first jumps to the right a bit and then starts moving.

rfrenchseti:
- Fix slider jumping when you click on it and move the mouse slightly.
- Tighten up the margins around the numbers in the slider.
- Fix problem with slider running off the right edge of the browser in some circumstances.
- Fixes #838.
- Tested on Chrome/Linux, FF/Linux, Chrome/Windows, FF/Windows, Safari/Mac, Safari/iOS.


dstopp (approved):
looks good

PR #896, opened 2019-11-15
Remove links to old feedback form

### LINKED ISSUE #857: Some feedback links still point to the old PDS feedback form
There are links in the help system that say "your comments are welcome" that still link to the old pds-rings feedback form. They need to pop up the new feedback widget instead. Or perhaps just not be there at all.

rfrenchseti:
-  Fixes #857

### DIFF
--- a/opus/application/apps/help/templates/help/about.html
+++ b/opus/application/apps/help/templates/help/about.html
@@ -185,8 +185,7 @@
 	and <a href="https://twitter.com/pdsopus" target="_blank">Twitter</a>.
 </p>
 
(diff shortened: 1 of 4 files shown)


dstopp (commented):
do you need to remove from apps\dictionary\templates\dictionary\footer.html as well?

rfrenchseti:
Right now the dictionary doesn’t have access to the new feedback tab so I thought I’d leave this until we redo the dictionary interface so there’s at least some way to give feedback. Also we aren’t running the dictionary anywhere.

dstopp (approved):
looks good.

PR #898, opened 2019-11-15
Add current columns to search URLs on detail page

### LINKED ISSUE #834: Detail tab search icon should preserve selected metadata
When clicking on the magnifying glass search icon next to a metadata value on the Detail tab, you get a new OPUS with that single search but default selected metadata. It would be better for the current selected metadata to be carried over to the new OPUS.

rfrenchseti:
- Fixes #834


dstopp (approved):
checked, looks good

PR #907, opened 2019-11-23
Minor bugfix to remove trashcan from selectMetadata

### LINKED ISSUE #724: Grey out or remove trashcan in Select Metadata when only one column left
B018: It is no longer possible to remove all columns, but when only one is visible the trashcan should be greyed out or disappear.

dstopp:
Fixes #724 (Remove trashcan when there is only one metadata column remaining)

Includes minor re-organization of file structure to isolate the files and data associated with the selectMetadata dialog out of browse.js and into selectMetadata.js.


rfrenchseti (changes requested):
- Look into replacing "info" and "unselect" with op- names
- If you click on the top row trash can repeatedly as fast as you can, you easily get into a state where there is more than one line but all the trashcans are gone

rfrenchseti:
Also, can you only remove the trash can on the first item in the list? Right now you can see the icon on the second element disappear before the animation starts to remove the whole line and it looks awkward.

dstopp:
Issues address; also fixed a bug in widgets that caused menu items in select metadata to be highlighted when they were only present in widgets but not in select metadata

rfrenchseti (approved):

PR #909, opened 2019-11-24
Fix download to exclude items in recycle bin.

### LINKED ISSUE #884: Internal error when attempting to download cart with no product types
When you clear all the product types and attempt to download an archive file, you get an internal error. You should go ahead and get the normal archive with checksum, manifest, urls, etc. Basically it should be the same as the URL archive in this case.

rfrenchseti:
- Fixes #884 as well


dstopp (approved):
looks good

Issue #845, opened 2019-08-29
Add recycle bin functionality


rfrenchseti:
Due to issue #739, we have currently disabled the ability to remove items from the cart while on the cart tab. We want to replace this with a new "recycle bin" functionality:

Back end changes:

- DONE: Modify the "cart" table to include a boolean "recycled" column. Thus an observation can be not in the cart, in the cart but not in the recycle bin, or in the cart and in the recycle bin.
- DONE: Modify the "dataimages.json" API to return the tri-state status of the observation by renaming "in_cart" to "cart_state" and making the return values False, "cart", and "recycle"
- DONE: Modify the cart remove/removerange API calls to take a new "recyclebin=0/1" argument. If 0, the behavior is as it is now. If 1, remove takes an obs in the cart and sets the recycled field to true; add always takes an obs where the recycled field is true and sets it to false.
- DONE: Modify the cart status API call to return both the number of items in the cart and the number of items in the recycle bin.
- DONE: Modify the cart reset API call to take the recyclebin=0/1 argument to say whether to empty the entire cart (including the recycle bin) or just the recycle bin.
- DONE: Modify addall.json to move items from the recycle bin back to the main cart.

Front end changes:

- Nothing has to change on the browse gallery/table except items in "recycle" state should not be highlighted.
- On the cart tab, items in "recycle" state should have a large "recycle bin" icon plastered across them, just like we do with the binoculars.
- On the cart tab, the "add to cart" icon should be "move to recycle bin" and have a different icon; likewise for "remove from cart" being "restore from recycle bin". Same for the hamburger menu items and the range options.
- On the cart tab, ctrl+click should toggle recycle bin state.
- On the cart tab, there need to be new buttons "Empty Recycle Bin" (which causes a hard refresh) and "Restore Recycle Bin". Somewhere related to these buttons there should be an indication of how many items are in the recycle bin.
- The cart status badge needs to display only the number of items in the cart, not the number of items in the recycle bin.

For responsiveness, the buttons that say "Empty Recycle Bin" etc might be combined together into some kind of drop-down menu. Also as the screen gets narrower the words might go away and be replaced by only icons.

rfrenchseti:
Log of backend changes:

- The "cart" table has a new "recycled" column. Before using these changes, you must regenerate the "cart" table by doing `python main_opus_import.py --create-cart`
[...]

PR #910, opened 2019-11-24
845 recycle bin functionality

dstopp:
Minor fixes to previous pull request #845 
 -disable buttons when the cart is empty;
 -fix the issue having to select two times to return obs to recycle bin after restore was selected
 -close popover when the screen gets too short and keep it closed

### DIFF
--- a/opus/application/apps/cart/templates/cart/cart.html
+++ b/opus/application/apps/cart/templates/cart/cart.html
@@ -14,7 +14,7 @@
                 </p>
             </div>
             <div class="op-download-options
-                    {% if count < 1 and recycled_count < 1 %}
+                    {% if count < 1 %}
                         op-button-disabled
                     {% endif %}
                 ">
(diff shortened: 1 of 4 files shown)


opus/application/apps/cart/templates/cart/cart.html
```diff
@@ -13,7 +13,11 @@
                     <span id="op-total-download-count">{{ total_download_count|intcomma }}</span>
                 </p>
             </div>
-            <div class="op-download-options">
+            <div class="op-download-options
+                    {% if count < 1 and recycled_count < 1 %}
```
rfrenchseti: Shouldn't this just be based on count, not recycled_count? Even if there are things in the recycle bin, we don't want to allow people to download, since there's nothing in the main cart.

rfrenchseti (changes requested):
- If count==0 and recycle_count>0 during browser refresh, the "Download CSV"/"Archive"/"URL" buttons are not disabled. They should be.
- If the user manipulates the cart such that everything is in the recycle bin and nothing is in the main cart, the "Download" buttons do not become disabled. They should.

rfrenchseti (approved):

PR #916, opened 2019-11-27
Add current search fields to metadata selector

### LINKED ISSUE #402: Give preference to search terms in select metadata
From Matt:

It’s a bit annoying that keywords on which I have searched are not default columns, and they’re not even called out as things I might want to show.  I need to click on “choose columns” and then go all the way through the directory tree exactly as I just did a minute ago to configure the search.  If I make a mistake, I might even show a column that isn’t exactly the same keyword on which I searched, possibly leading to confusion.  
One thing that might help might be to simply add a panel to the “choose columns” module that shows the keywords that were used in the current search but are not currently shown in the table.

rfrenchseti:
- Add "Current Search Fields" category to Select Metadata if there are any open widgets.
- Open the Search Fields category by default and not the General category if there are open widgets.
- Fixes #402 .
- Also removed old cookie code that stored the current widgets as a cookie.


dstopp (commented):
code looks ok.
I think it's a bit odd to have the duplicated items in the list twice - the Current Search Fields category just looks like another category that has the same things as the ones below, which is confusing.  
Couple of suggestions:
 - remove the current search fields items from their 'home' category so they are not duplicated below;
 - consider different formatting for current search fields to differentiate it from the other categories.

PR #918, opened 2019-11-27
Fix sort order of product types and add NH alternate

### LINKED ISSUE #833: Some New Horizons product types missing tooltips
Some New Horizons product types don't have tooltip info, such as "Duplicate Image (Lossy, Unbinned)". These come from PdsFile with their opus_type constructed instead of present in the opus type table.

rfrenchseti:
- Fixes #833 
- Should be coordinated with changes to pds-webtools (rules/NHxxxx_xxxx.py) made on 11/27/2019.
- Only affects newly imported databases


rfrenchseti:
The new order has been approved by Mark.

Issue #868, opened 2019-10-11
Add buttons to download metadata from Select Metadata dialog


rfrenchseti:
Request from Mark:

Add two buttons to the Select Metadata dialog: Download CSV of All Metadata and Download CSV of Metadata from Cart (shorten to make readable/pretty as necessary). These would do the exact same thing as the existing Download links on the browse and cart tabs.

Possible future change:

Remove "Download CSV" from the nav bar entirely, and rename Select Metadata to Select/Download Metadata. This is still open for debate.

PR #925, opened 2019-12-04
868 add buttons for download csv

dstopp:
#868 Add button to select metadata dialog to download CSV

-Added a button below the currently selected metadata columns to allow the user to download the selected metadata from the dialog.
If the user is on the browse tab, the button will download all observations in current search for selected metadata columns.
If the user is on the cart tab, the button will download the selected metadata columns for the observations in the cart.

Button text and title updated depending on which tab the user is currently on.


opus/application/static_media/js/browse.js
```diff
@@ -2239,4 +2224,23 @@ var o_browse = {
         $("#browse .gallery").empty();
         $("#browse .op-data-table tbody").empty();
     },
+
+    downloadCSV: function(obj) {
+        let colStr = opus.prefs.cols.join(',');
+        let selectionsHash = [];
+        for (let param in opus.selections) {
```
rfrenchseti: This will utterly fail with the new multiple-input selections feature. In fact if you have multiple input rows and then select download CSV you'll get a download error. Sadly, this appears to be in the production version too! So it needs to be fixed.

Look at the new function o_hash.GetHashStrFromSelections and that should be able to replace all this code.

rfrenchseti (changes requested):

rfrenchseti (approved):

PR #929, opened 2019-12-06
Issue 46 units selection

### LINKED ISSUE #46: Implement unit indicators/selectors for numerical input widgets

juzen2003:
- Fixes #46 and #886 
- Merge pull request https://github.com/SETI/pds-opus/pull/891 as the base, and update from there.
- Run JSHint on: search.js, hash.js, and widgets.js
- Test on latest Chrome, FF, Opera, and Safari.
- Done Items:
    - Update unit- to opus.extras in the following cases: 
        - Open a widget and select a different value in qtype:
            - properly update opus.extras, unit should stay in unit- not qtype-
            - properly update opus.selections/selections, unit should stay in extras with unit- not selections
        - Properly add unit- in URL when open a widget
        - Properly remove unit- in URL when close a widget
        - Update unit- in URL when an input is added/remove 
        - Update unit- in URL when refreshing the page with multiple input in a widget 
        - Update unit- in URL when calling API (getHashStrFromSelections) 
        - Update unit- when page reloads (getHashStrFromSelections, initFromHash) 
    - Update hints when unit is changed.
    - No search will be performed if unit is changed when all input sets are empty, only hints will be updated.
    - No search will be performed if qtype is changed when corresponding input set is empty.


log_analyzer/opus/slug.py
```diff
@@ -70,7 +70,7 @@ class ToInfoMap:
     # Slugs that should be ignored when see them as either a column name or as a search term.
     SLUGS_NOT_IN_DB = {'browse', 'cart_browse', 'cart_startobs', 'col_chooser', 'colls_browse', 'cols', 'detail',
                        'gallery_data_viewer', 'limit', 'loc_type', 'order', 'page',
-                       'range', 'recyclebin', 'reqno', 'request', 'startobs', 'types', 'url_cols', 'view', 'widgets', 'widgets2',
```
rfrenchseti, line 73: There appears to be a merge error with this. The recyclebin and url_cols slugs need to stay in the list.

opus/application/apps/ui/templates/ui/widget.html
```diff
@@ -11,7 +11,14 @@
         <h6 class="card-title mt-0">
             <div>
                 <i class="fas fa-info-circle" title="{{ tooltip|safe }}"></i>
-                {{ label }} {{ units }}
+                {{ label }}&nbsp;
+                {% if valid_units %}
+                    <select class="unit-{{ slug }}" name="unit-{{ slug }}" tabindex="0">
```
rfrenchseti: Shouldn't these names of "op-" in front of them? And do we really need identical classes and IDs?

opus/application/static_media/js/hash.js
```diff
@@ -126,38 +131,59 @@ var o_hash = {
...
-                let qtypeValue = opus.extras[qtypeSlug];
-                if (qtypeValue.length) {
-                    let encodedExtraValues = o_hash.encodeSlugValues(qtypeValue);
+                let slugInExtrasVal = opus.extras[slugInExtras];
+                if (slugInExtrasVal.length) {
+                    let encodedExtraValues = o_hash.encodeSlugValues(slugInExtrasVal);
 
-                    if (qtypeValue.length > 1) {
+                    if (slugInExtrasVal.length > 1) {
                         let numberOfQtypeInputs = encodedExtraValues.length;
```
rfrenchseti: Shouldn't this be just `numberOfInputs` or something since this can either be qtype or unit slug?

opus/application/static_media/js/hash.js
```diff
@@ -126,38 +131,59 @@ var o_hash = {
...
                         for(let trailingCounter = 1; trailingCounter <= numberOfQtypeInputs; trailingCounter++) {
                             let trailingCounterString = o_utils.convertToTrailingCounterStr(trailingCounter);
-                            let newKey = `${qtypeSlug}_${trailingCounterString}`;
+                            let newKey = `${slugInExtras}_${trailingCounterString}`;
 
-                            if (qtypeValue[trailingCounter-1] !== null) {
+                            if (slugInExtrasVal[trailingCounter-1] !== null) {
                                 hash.push(newKey + "=" + encodedExtraValues[trailingCounter-1]);
                             }
+                            if (unitSlug in opus.extras) {
```
rfrenchseti, line 192: This is kind of tricky. I finally figured out that "unitSlug" is empty if the main slug is already a unit slug, which is counter-intuitive. I think this needs some kind of comment here and line 177, or perhaps just above where unitSlug is first created.

opus/application/static_media/js/hash.js
```diff
@@ -126,38 +131,59 @@ var o_hash = {
...
+                    if (slugInExtrasVal.length > 1) {
                         let numberOfQtypeInputs = encodedExtraValues.length;
 
                         for(let trailingCounter = 1; trailingCounter <= numberOfQtypeInputs; trailingCounter++) {
                             let trailingCounterString = o_utils.convertToTrailingCounterStr(trailingCounter);
-                            let newKey = `${qtypeSlug}_${trailingCounterString}`;
+                            let newKey = `${slugInExtras}_${trailingCounterString}`;
 
-                            if (qtypeValue[trailingCounter-1] !== null) {
+                            if (slugInExtrasVal[trailingCounter-1] !== null) {
```
rfrenchseti, line 187: Why is the "main" slug (whether qtype or unit) inserted directly using hash.push, but the associated unit slug, if any, is inserted by calling `updateHashFromExtras`? Shouldn't they be consistent?

opus/application/static_media/js/hash.js
```diff
@@ -354,7 +383,7 @@ var o_hash = {
             return [undefined, undefined];
         }
 
-        hash = (hash.search('&') > -1 ? hash.split('&') : [hash]);
+        hash = (hash.search("&") > -1 ? hash.split("&") : [hash]);
```
rfrenchseti, line 408: Comment before this function should say "qtype or unit"

opus/application/static_media/js/hash.js
```diff
@@ -419,7 +448,7 @@ var o_hash = {
         return [selections, extras];
     },
 
-    SyncUpWithOPUSSelectionsForWidgetsWithNoQtype: function(selections, extras, widgetSlug) {
+    syncUpWithOPUSSelectionsForWidgetsWithNoQtype: function(selections, extras, widgetSlug) {
```
rfrenchseti: I probably understood this function once, but I don't anymore. Can you elaborate in the comments why this function here and what it does? Under what circumstances is there a string/range widget but no qtype?

rfrenchseti: Also, below we check for qtype in extras. Do we also need to check for unit-? Is the name of the function correct?

opus/application/static_media/js/hash.js
```diff
@@ -522,7 +554,7 @@ var o_hash = {
             let value = pair.slice(idxOfFirstEqualSign + 1);
 
             if (value) {
-                if (slug.match(/qtype-.*/)) {
+                if (slug.match(/qtype-.*/) || slug.match(/unit-.*/)) {
```
rfrenchseti: Better with startsWith?

opus/application/static_media/js/search.js
```diff
@@ -360,38 +360,99 @@ var o_search = {
...
-
-            switch ($(this).attr("class")) {  // form type
-                case "RANGE":
+            let isInputSetEmpty = true;
+            if ($(this).attr("name").startsWith("qtype-")) {
+                let qtypes = [];
+                let counterStr = o_utils.getSlugOrDataTrailingCounterStr($(this).attr("name"));
+                let idx = counterStr ? counterStr - 1 : 0;
+                switch ($(this).attr("class")) {  // form type
+                    case "RANGE":
```
rfrenchseti, line 391: I'm not sure why this wasn't done before, but the code after each "case" should be indented one level.

opus/application/static_media/js/search.js
```diff
@@ -360,38 +360,99 @@ var o_search = {
...
+                    if (opus.selections[`${slugNoNum}`][idx]) {
+                        isInputSetEmpty = false;
+                    }
                     break;
+                }
+            } else if ($(this).attr("name").startsWith("unit-")) {
+                let units = [];
+                let slugNoNum = $(this).attr("name").match(/unit-(.*)$/)[1];
+                let numberOfInputSets = $(`#widget__${slugNoNum} .op-search-inputs-set`).length;
+                while(units.length < numberOfInputSets) {
```
rfrenchseti: Space after while

opus/application/static_media/js/widgets.js
```diff
@@ -591,6 +601,7 @@ var o_widgets = {
 
         delete opus.extras[`qtype-${slugNoNum}`];
         delete opus.extras[`z-${slugNoNum}`];
+        delete opus.extras[`unit-${slugNoNum}`];
```
rfrenchseti, line 605: Let's get rid of that line that says "z-". That's left over from some ancient feature we no longer support. Perhaps it was the old version of units? Before my time.

opus/application/static_media/js/widgets.js
```diff
@@ -909,15 +920,29 @@ var o_widgets = {
             // This will also put qtype in the url when a widget with qtype is open.
             // Need to wait until api return to determine if the widget has qtype selections
             let hash = o_hash.getHashArray();
-
             // NOTE: inputs & qtypes are not renumbered yet at this stage.
-            let qtype = "qtype-" + slug;
-            let qtypeInputs = $(`#widget__${slug} select[name="${qtype}"]`);
+            let qtype = `qtype-${slug}`;
+            let qtypeInputs = $(`#widget__${slug} .widget-main select[name="${qtype}"]`);
```
rfrenchseti: Should .widget-main be renamed to .op-widget-main?

rfrenchseti (changes requested):
- Nicely done. I only found one real issue and a handful of small code comments.
- Preprogrammed ranges have units associated with them. Right now they are ignored. But when the user selects a range, the UI needs to check if the ranges' unit is equal to the widget's currently selected unit. If not, and there are other input lines with values in them, there needs to be some kind of error dialog. If there are no other lines with values in them, then the widget's unit needs to be set appropriately.
- Unrelated to the current pull, but I notice whenever you create a new widget, there is a spinner in the header to the right of the title that comes on and eventually goes away, but I have no idea what it's spinning for. Nothing is being waiting on. The result count is good, the hints are good, etc. Maybe that spinner should go away?

rfrenchseti (changes requested):
One more bug:
- The dropdown boxes are using the "internal" name for each unit (like "saturnradii") not the "display" name (like "Rs (71492)"). See opus_support.py for how the display names are encoded. The idea is the display name should show up to the user, but when the user selects it the internal name should be put in the unit-XXX slug.

juzen2003:
- Update preprogrammed ranges info when unit is changed.
- Still tuning styling.

log_analyzer/opus/slug.py
rfrenchseti: This has not been fixed.

rfrenchseti (changes requested):
- When you change the units, the URL is updated with the current input values even if they are illegal. For example, type something bad in the ring radius "min" field. Exit the field and change the widget's unit to something else. Notice that the bad value is now in the URL.
- Look at the ring radius "thebe extension" when using "m" units and narrow the browser. The numbers are wide enough that they lay on top of the word "extension".
- How are you figuring out the number of decimal places to put when you change the units in the range dropdown? For example, in the wavelength widget, select "cm" and then type in a color name. It will just show "0.0" because there aren't enough digits available. Check out the function `adjust_format_string_for_units` in opus_support.py. It will automatically adjust a format string to the proper number of decimal places given the new units.

opus/application/static_media/js/search.js
```diff
@@ -603,6 +605,11 @@ var o_search = {
                     o_search.highlightMatchedRangesName(singleRangeData, currentInputValue);
                     o_search.rangesNameMatchedCounterByCategory[collapsibleContainerId] += 1;
                     if (!$(`#${collapsibleContainerId}`).hasClass("show")) {
+                        // Normally inputsRangesNameMatchedInfo gets updated later in this function,
+                        // but since we are opening the collpasible item here, to make sure all behaviors
```
rfrenchseti: collapsible spelled wrong here and below

lib/opus_support.py
```diff
@@ -463,8 +463,8 @@ def _range_time_decode(iso):
             'display_name': 'km',
             'conversions': {
                 'm':            ('m', 1e-3),
-                'saturnradii':  ('Rj (60330)', 60330.),
-                'jupiterradii': ('Rs (71492)', 71492.),
+                'saturnradii':  ('Rs (60330)', 60330.),
```
rfrenchseti: Whoops. Good catch. But Jupiter should come before Saturn in the list. They are in order inner to outer planet.

rfrenchseti (changes requested):
- normalizeinput does not have the units passed, so the returned "pretty" value can have the wrong number of digits, resulting in the number being truncated incorrectly.

rfrenchseti:
Test comment

opus/application/apps/search/views.py
```diff
@@ -653,6 +687,13 @@ def url_to_search_params(request_get, allow_errors=False, return_slugs=False,
                                       +'function "%s"', form_type_func)
                             return None, None
                     if value_to_use:
+                        if is_sourceunit or sourceunit_val is not None:
```
rfrenchseti: Note the code starting at line 697. It catches exceptions during the execution of func() and prevents the back end from crashing. The new code calls func() when there’s a sourceunit without catching these. Also if there’s a sourceunit, func() gets called twice, once here and once inside the try. Not fatal if func is float or int, but bad if it’s something converting with a more complicated format. 

As much as I hate putting lots of code in a try block, I think the right thing to do is move the unit conversion a few lines down to just before the if pretty_results and use new_value instead of func(value_to_use).

opus/application/static_media/js/search.js
```diff
@@ -376,6 +376,8 @@ var o_search = {
         // range behaviors and string behaviors for search widgets - qtype select dropdown
         $('#search').on("change", "select", function() {
             let isInputSetEmpty = true;
+            // Use this flag to determine if
```
rfrenchseti: Comment is truncated

opus/application/static_media/js/search.js
```diff
@@ -376,6 +376,8 @@ var o_search = {
...
+            let performNormalizeInput = false;
             if ($(this).attr("name").startsWith("qtype-")) {
```
rfrenchseti, line 386: The comment before this function should mention unit drop down too

rfrenchseti (changes requested):
- There's a race condition with the new unit conversion. If you change units from A to B, and then change units again B to C before the first normalizeinput returns, the second call to normalizeinput gets called with unit=C, sourceunit=B but the numbers that get passed in are still in units of A. I'm not sure what the right way to fix this is. It seems like you should keep the units along with the numbers in the inputs, not just in the dropdown box. Then update the units in the inputs when normalizeinput successfully returns and you're updating the input contents.

rfrenchseti (changes requested):
- Type in a value. Change the unit without otherwise leaving focus first. The value you just typed will not be converted.
- Something similar happens with preprogrammed ranges. Type in a range name. Click on the range. Change the units before normalizeinput returns. The value will be in the wrong units.

Issue #138, opened 2015-04-12
Add 'Add all to Cart' button on main gallery page


basilleaf:
Originally reported by: **lisa ballard (Bitbucket: [basilleaf](https://bitbucket.org/basilleaf), GitHub: [basilleaf](https://github.com/basilleaf))**

---

Mark and user Michael both requested this, quick and dirty benchmark says it's not as slow as I'd thought it would be! 

---
- Bitbucket: https://bitbucket.org/ringsnode/opus2/issue/137

basilleaf:
this is actually ready [on branch "builk_add"](https://github.com/basilleaf/opus/tree/bulk_add) but holding it until dev server is ready for testing it out on full production-sized database.

basilleaf:
This is now merged into master, but is hidden in interface b/c needs more testing/limits and such. 

see: [0787e6da0fbe81e86](https://github.com/basilleaf/opus/commit/ff094a96ebaaf411c824f460787e6da0fbe81e86)

it's the edit_collection_addall() in user_collections

basilleaf:
here is where the [feature was added](https://github.com/basilleaf/opus/commit/5219f5990a519534822452f9d57bf7371efd43f9)

rfrenchseti:
This will need changes to the Django to permit efficient adding/removing all, efficient check of an image in the cart, and checks on download and CSV size.

rfrenchseti:
addall is implemented in the API, but still needs to be invoked by a button on the browse tab.

rfrenchseti:
Meeting today: Put a checkbox at the top of the cart column in the table view. Put an entry in the hamburger menu. Both should only be available on the browse tab.

rfrenchseti:
Also change styling on hamburger menu to have much less vertical padding between items

PR #943, opened 2019-12-25
Issue 730 & 138: implement add all to cart

### LINKED ISSUE #730: Add select all option to top of table view
B071: Add shopping cart icon to the top of the "checkbox" column. Clicking on the icon will do an "add all" operation.

note: separate out the cart and hamburger into two columns; add the cart icon over the cart column; add a 'tools' icon over the hamburger column.

juzen2003:
- Initial implementation of add all to the cart 
- Fixes #730 and #138 
- Pending Items:
  - Finalize the UI for the 1st column of table header in browse tab.
    - Current icons are a shopping cart (add all button & tooltips) icon and a toolbox icon (only tooltips). Maybe just one shopping cart icon will look better?
  - Fix the dropdown menu location when hamburger is clicked.
    - Need to fix the calculation of top & left in showMenu in browse.js


rfrenchseti (changes requested):
- Add all to cart should only be in the hamburger menu while on the browse tab. It should disappear on the cart tab.
- "Add all results to cart" should be in lower case like the other menu options, not in title case.
- Move "add all" after "add range" in hamburger menu.
- I think we can squeeze more vertical space out of the hamburger menu.
- The dialog asking for confirm of add all should have the number of observations (result count) in it. "Are you sure you want to add all XXX,XXX observations to the cart?" Be sure to include commas.
- We should not allow the user to access the add all option if we know it will fail because there are too many observations. Pass in the settings.MAX_SELECTIONS_ALLOWED parameter as a constant through header.html and then check it in the JS. You can't get it precisely right because you don't know what's already in the cart, but you can at least see if the current result count is > MAX_SELECTIONS_ALLOWED and pop up a dialog that says "There are too many results to add all to the cart. Please reduce the number of results to {MAX_SELECTIONS_ALLOWED} or fewer and try again."
- I think browse.js:addAllToCart should actually be in cart.js. And most of that function duplicates code already in cart.js:toggleInCart. Those functions should somehow be combined or refactored so the code isn't duplicated.

opus/application/apps/ui/templates/ui/base.html
```diff
@@ -362,6 +363,27 @@ <h3 class='op-header-text'></h3>
...
+            target="op-addall-to-cart"
+            yes_text="Yes"
+            no_text="No"
+            allow_close=True
+            draggable=True
+            close_by_keyboard="esc"
+        %}
+        {% include "ui/confirm_modal.html" with
+            id_name="op-addall-warning-msg"
+            confirm_title="Message About Add All Action"
```
rfrenchseti: Let's change the title to "Failed to Add All Observations"

opus/application/apps/ui/templates/ui/header.html
```diff
@@ -59,6 +59,7 @@
         const DEFAULT_SORT_ORDER = '{{ default_sort_order }}';
         const STATIC_URL = '{{ STATIC_URL }}';
         const MAX_SELECTIONS_FOR_DATA_DOWNLOAD = '{{ MAX_SELECTIONS_FOR_DATA_DOWNLOAD }}';
+        const MAX_SELECTIONS_ALLOWED = '{{ max_selections_allowed }}';
```
rfrenchseti, line 61: The line about "MAX_SELECTIONS_FOR_DATA_DOWNLOAD" seems to be old, dead code. The variable isn't even being passed in to the template context. Let's go ahead and remove it.

opus/application/static_media/js/browse.js
```diff
@@ -523,6 +526,26 @@ var o_browse = {
...
+         * Display a modal before add all action. If the result count is more
+         * than MAX_SELECTIONS_ALLOWED, display a warning modal to ask user
+         * to reduce the number of obs before add all aciton. If the result
+         * count is less than MAX_SELECTIONS_ALLOWED, display a confirmation
+         * modal.
+         */
+        if (o_browse.totalObsCount <= MAX_SELECTIONS_ALLOWED) {
+            $("#op-addall-to-cart").modal("show");
+        } else {
+            let warningMsg = "There are too many results to add all to the cart. " +
```
rfrenchseti: Change to "There are too many results to add them all to the cart."

opus/application/static_media/js/cart.js
```diff
@@ -696,25 +699,52 @@ var o_cart = {
...
+        action = (action === "addrange" || action === "add" || action === "addall" ? "add" : action);
         let status = (action === "add" ?  "add" : "remove");
         let checked = (action === "add");
 
         $.when(o_cart.sendEditRequest(url)).done(function(statusData) {
             o_utils.enableUserInteraction();
             if (statusData.error) {
                 // if previous error modal is currently open, we store the error message for later displaying
-                if ($("#op-cart-status-error-msg").hasClass("show")) {
+                if ($("#op-cart-status-error-msg").hassClass("show")) {
```
rfrenchseti: Spelling error hasClass?

rfrenchseti (changes requested):
- I think the toolbox icon in the table view ends up just being confusing, since you can click on the shopping cart but not on the toolbox. Probably best to just remove it.
- When you're in the middle of adding/removing a range, and do an add all, it leaves you in the middle of doing a range with the green square continuing to pulse. I think it should abort the add range when the add all is confirmed. It looks like you might have explicitly removed this feature around cart.js line 743?

opus/application/static_media/js/cart.js
```diff
@@ -539,6 +539,9 @@ var o_cart = {
                     }
                 }
             });
+        } else {
+            // Make sure "Add all results to cart" is hidden in cart tab.
+            $("#op-obs-menu .dropdown-item[data-action='addall']").addClass("op-hide-element");
         }
```
dstopp, line 546: Curiously, aren't you already doing this in loadData()?

dstopp (changes requested):
I did not get a fail message when I started w/9989 items in my cart and tried an "AddAll" w/the search set for 207.  I would expect it to complain that I was exceeding my limit of 10k but i never saw the error.  I know it did error because i see this in my log: 
{"error": "Your request to add all 207 observations to the cart failed. The resulting cart and recycle bin would have more than the maximum (10,000) allowed. None of the observations were added.", "count": 9989, "recycled_count": 0, "reqno"

rfrenchseti:
Actually the problem with not showing an error is not just with add all. If you add 10,000 observations and then try to add a single more using the normal "add cart" feature, the backend gives an error return but the UI hangs.

opus/application/static_media/js/cart.js
juzen2003: Yes, but because every time we go to browse tab (call activateBrowseTab, and loadData is called as well in the function), the icon will be shown. And if we switch back to cart tab (from browse tab) again without reloading obs in cart tab, the icon will not be hidden (loadData is not called in activateCartTab). Therefore, we hide the icon here for the case that obs in cart tab is not reloaded.

juzen2003:
Note: before performing add all, if the result count in browse tab is larger than MAX_SELECTIONS_ALLOWED, we will pop out the warning modal, else we will pop out a confirm modal (with result count info in browse tab) to confirm the add all action.

rfrenchseti (approved):

Issue #870, opened 2019-10-11
Option to show thumbnails in table view


rfrenchseti:
Request from Mark:

Have an option (like a checkbox) that displays very small thumbnails to the left of each table row.

PR #954, opened 2020-01-09
870 mini thumbnails in table

dstopp:
#870 
-Added mini-thumbnails to the table view for both browse and cart tabs
-column not resizeable
-roll over image temporarily zooms the image
Note: the mini images have been scaled to not change the size of the row from the original height.

### DIFF
--- a/opus/application/static_media/css/opus.css
+++ b/opus/application/static_media/css/opus.css
@@ -616,7 +616,6 @@ th.sticky-header {
 }
 
 .op-mini-thumbnail div {
-    /*text-align: center;*/
     display: flex;
     justify-content: center;
     align-items: center;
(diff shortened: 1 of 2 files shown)


opus/application/static_media/css/opus.css
```diff
@@ -610,6 +610,36 @@ th.sticky-header {
     bottom: 1px;
 }
 
+.op-mini-thumbnail {
+    width: 3em;
+    min-width: auto !important;
+}
+
+.op-mini-thumbnail div {
+    /*text-align: center;*/
```
rfrenchseti: Remove comment?

opus/application/static_media/css/opus.css
```diff
@@ -610,6 +610,36 @@ th.sticky-header {
...
+}
+
+.op-mini-thumbnail img {
+    transition: transform .1s;
+    width: 68%;
+}
+
+/* zoom the mini-thumbnail on hover */
+.op-mini-thumbnail img:hover {
+  -ms-transform: scale(4.5); /* IE 9 */
```
rfrenchseti: Should be 4 character indent

rfrenchseti (changes requested):
- The topmost thumbnail in the table, when expanded, appears to go *under* the table header, so you can't see all of it.
- When a thumbnail is expanded so that it covers multiple small thumbnails and you move the mouse up or down to the next underlying small thumbnail, it doesn't activate. You have to move the mouse all the way off the expanded thumbnail. This means that as you move your mouse down the thumbnails, you only see every 2 or 3 instead of triggering an expansion of all of them one at a time. I think it would be better to be able to see all of them.

rfrenchseti (approved):

PR #956, opened 2020-01-14
Issue 872 UI change for surface geometry searching

### LINKED ISSUE #663: Expand target/surface target on permalink
When coming in from a permalink, such as:

http://127.0.0.1:8000/#/SURFACEGEOplutocenterresolution2=5.00000&SURFACEGEOplutophase1=160.000&surfacegeometrytargetname=Pluto&qtype-SURFACEGEOplutophase=any&cols=opusid,instrument,time1,SURFACEGEOplutoplanetographiclatitude1,SURFACEGEOplutoplanetographiclatitude2,SURFACEGEOplutoIAUwestlongitude1,SURFACEGEOplutoIAUwestlongitude2,SURFACEGEOplutocenterdistance,SURFACEGEOplutocenterresolution,SURFACEGEOplutophase1,SURFACEGEOplutophase2&widgets=SURFACEGEOplutocenterresolution,SURFACEGEOplutophase,surfacegeometrytargetname&order=time1,opusid&view=search&browse=gallery&cart_browse=gallery&startobs=1&cart_startobs=1&detail=

that has an Intended Target or a Surface Geometry Target, any planets that have a selection inside them should be expanded so the user can actually see the selection by default. For example, in this case the Surface Geometry Target category for "Pluto" should be expanded so the user can see that Pluto is selected.

### LINKED ISSUE #872: UI changes for surface geometry searching
Ideas from Mark:

When changing the Surface Geometry Target, we currently destroy all search widgets for the previous target. This means the user's search parameters are lost. Instead, we should change those widgets (and their respective slugs) over to the new target. This way a user could search on a certain phase angle for Enceladus, then search on the same phase angle for Tethys, without having to open new widgets and type in the search parameters again.

A second part of this might be to silently add new metadata fields as well. If the user already has a field such as "body center phase angle [Enceladus]" and switches the surface target from Enceladus to Tethys, we could add a new metadata field "body center phase angle [Tethys]" for convenience.

Mark suggests we go ahead and add part 2 as well and see how we like it.

juzen2003:
- Fixes #872
- Fixes #663 
- Run JSHint on: search.js, widgets.js, menu.js, hash.js
- Test on latest: Chrome, Firefox, Safari, and Opera.
- One issue: getting mults hints when switching to a different surfacegeo target:
  -  We not only change surfacegeo target name, but also change all related surfacegeo slugs to new target ones. So the mults hints displayed in surface target name is the correct return from api call but it will not be the result count for the future next click. For example, switching in surfacegeo target name from surfacegeometrytargetname=Mimas&SURFACEGEOmimasplanetographiclatitude1=7 (Pallene has 1 result) to surfacegeometrytargetname=Pallene&SURFACEGEOpalleneplanetographiclatitude1=7 (Pallene actually has 0 result, not 1).

### DIFF
--- a/opus/application/apps/results/views.py
+++ b/opus/application/apps/results/views.py
@@ -1577,9 +1577,9 @@ def get_triggered_tables(selections, extras, api_code=None):
             if (trigger_tab == 'obs_surface_geometry' and
                 trigger_val.upper() ==
                 selections['obs_surface_geometry.target_name'][0].upper()):
-                if (trigger_val.upper() in
-                    [r['target_name'].upper() for r in results]):
-                    triggered_tables.append(partable)
+                # If the selected surfacegeo target has no result, we
+                # still want to have related menu item displayed.
+                triggered_tables.append(partable)
 
     # Now hack in the proper ordering of tables
     final_table_list = []
--- a/opus/application/apps/results/views.py
+++ b/opus/application/apps/results/views.py
@@ -1578,7 +1578,7 @@ def get_triggered_tables(selections, extras, api_code=None):
                 trigger_val.upper() ==
                 selections['obs_surface_geometry.target_name'][0].upper()):
                 # If the selected surfacegeo target has no result, we
-                # still want to have related menu item displayed.
+                # still want to have the related menu item displayed.
                 triggered_tables.append(partable)
 
     # Now hack in the proper ordering of tables
--- a/opus/application/apps/ui/templates/ui/base.html
+++ b/opus/application/apps/ui/templates/ui/base.html
@@ -485,6 +485,17 @@
             draggable=False
             close_by_keyboard="esc"
         %}
+        {% include "ui/confirm_modal.html" with
+            id_name="op-close-surfacegeo-widgets"
+            confirm_title="Confirm Close All Surfacegeometry Widgets"
+            confirm_msg="Are you sure you want to close all surfacegeometry widgets?"
+            target="op-close-surfacegeo-widgets"
+            yes_text="Yes"
+            no_text="No"
+            allow_close=True
+            draggable=True
+            close_by_keyboard="esc"
+        %}
 
         {% endblock %}
 
--- a/opus/application/apps/ui/templates/ui/base.html
+++ b/opus/application/apps/ui/templates/ui/base.html
@@ -487,8 +487,8 @@
         %}
         {% include "ui/confirm_modal.html" with
             id_name="op-close-surfacegeo-widgets"
-            confirm_title="Confirm Close All Surfacegeometry Widgets"
-            confirm_msg="Are you sure you want to close all surfacegeometry widgets?"
+            confirm_title="Confirm Close All Surface Geometry Search Fields"
+            confirm_msg="Closing the Surface Geometry Target Name search field will also close all target-specific search fields. Are you sure you wish to proceed?"
             target="op-close-surfacegeo-widgets"
             yes_text="Yes"
             no_text="No"
--- a/opus/application/static_media/js/hash.js
+++ b/opus/application/static_media/js/hash.js
@@ -404,7 +404,6 @@ var o_hash = {
         if (!hash) {
             return [undefined, undefined];
         }
-
         hash = (hash.search("&") > -1 ? hash.split("&") : [hash]);
         hash = o_hash.decodeHashArray(hash);
         let selections = {};  // the new set of pairs that will not include the result_table specific session vars
@@ -641,9 +640,9 @@ var o_hash = {
                     let slugCounter = o_utils.getSlugOrDataTrailingCounterStr(slug);
                     slug = slugNoCounter;
 
-                    // For surfacegeometrytargetname, we have to assign the init value
-                    // to opus.oldSurfacegeoTarget here if the page reloads with URL
-                    // that has surfacegeometrytargetname.
+                    // If there is a value specified for surfacegeometrytargetname in the URL,
+                    // we have to initialize opus.oldSurfacegeoTarget with it so that we can
+                    // properly handle changes to surfacegeometrytargetname in the future.
                     if (slug === "surfacegeometrytargetname") {
                         opus.oldSurfacegeoTarget = o_utils.getSurfacegeoTargetSlug(value);
                     }
--- a/opus/application/static_media/js/hash.js
+++ b/opus/application/static_media/js/hash.js
@@ -641,6 +641,13 @@ var o_hash = {
                     let slugCounter = o_utils.getSlugOrDataTrailingCounterStr(slug);
                     slug = slugNoCounter;
 
+                    // For surfacegeometrytargetname, we have to assign the init value
+                    // to opus.oldSurfacegeoTarget here if the page reloads with URL
+                    // that has surfacegeometrytargetname.
+                    if (slug === "surfacegeometrytargetname") {
+                        opus.oldSurfacegeoTarget = o_utils.getSurfacegeoTargetSlug(value);
+                    }
+
                     if (slugCounter > opus.maxAllowedInputSets) {
                         return; // continue to next iteration
                     }
--- a/opus/application/static_media/js/utils.js
+++ b/opus/application/static_media/js/utils.js
@@ -109,7 +109,18 @@ var o_utils = {
     // or inner objects. In our case, we only have arrays as values.
     deepCloneObj: function(obj) {
         return JSON.parse(JSON.stringify(obj));
-    }
+    },
+
+    getSurfacegeoTargetSlug: function(target) {
+        /**
+         * Take in a surface geo target pretty name and return a slug name
+         * for the target.
+         */
+        let slugName = target.toLowerCase();
+        // remove all "_", "/", and " "
+        slugName = slugName.replace(/_/g, "").replace(/\//g, "").replace(/ /g, "");
+        return slugName;
+    },
 };
 
 /**
--- a/opus/application/static_media/js/utils.js
+++ b/opus/application/static_media/js/utils.js
@@ -111,6 +111,7 @@ var o_utils = {
         return JSON.parse(JSON.stringify(obj));
     },
 
+    // This function is patterned after slug_name_for_sfc_target in import_util.py.
     getSurfacegeoTargetSlug: function(target) {
         /**
          * Take in a surface geo target pretty name and return a slug name
--- a/opus/application/static_media/js/widgets.js
+++ b/opus/application/static_media/js/widgets.js
@@ -164,15 +164,17 @@ var o_widgets = {
 
                 // Update metadata selections
                 let currentMetadataSelections = opus.prefs.cols.slice();
+                let newSurfacegeoMetadataSelections = [];
                 for (const col of currentMetadataSelections) {
                     if (col.match(oldTargetStr)) {
                         let newCol = col.replace(oldTargetStr, newTargetStr);
                         if (!opus.prefs.cols.includes(newCol)) {
-                            let idxToInsert = opus.prefs.cols.indexOf(col);
-                            opus.prefs.cols.splice(idxToInsert, 0, newCol);
+                            newSurfacegeoMetadataSelections.push(newCol);
                         }
                     }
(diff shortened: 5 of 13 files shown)


opus/application/static_media/js/widgets.js
```diff
@@ -93,18 +93,71 @@ var o_widgets = {
             }
         });
 
-        // close opened surfacegeo widget if user select another surfacegeo target
+        // Update surfacegeo widgets in place if user selects another surfacegeo target.
+        // 1. Update surfacegeo attributes/text in DOMs of all related surfacegeo widgets.
+        // 2. Update selections (opus.selection, opus.extras) & widgets.
         $('#search').on('change', 'input.singlechoice', function() {
```
rfrenchseti, line 105: Don't we need some kind of test here that it's the surface geo selection that's changing? What if we had another radio button widget in the future?

opus/import/import_util.py
```diff
@@ -262,6 +262,8 @@ def table_name_for_sfc_target(target_name):
         target_name = TARGET_NAME_MAPPING[target_name.upper()]
     return encode_target_name(target_name)
 
+# NOTE: whenever we change this function, we will have to change
+# getSurfacegeoTargetSlug in JS codes as well.
```
rfrenchseti: "code". Also mention what file it's in.

opus/application/apps/results/views.py
```diff
@@ -1577,9 +1577,9 @@ def get_triggered_tables(selections, extras, api_code=None):
             if (trigger_tab == 'obs_surface_geometry' and
                 trigger_val.upper() ==
                 selections['obs_surface_geometry.target_name'][0].upper()):
-                if (trigger_val.upper() in
-                    [r['target_name'].upper() for r in results]):
-                    triggered_tables.append(partable)
+                # If the selected surfacegeo target has no result, we
+                # still want to have related menu item displayed.
```
rfrenchseti: have THE related menu item

opus/application/static_media/js/hash.js
```diff
@@ -641,6 +641,13 @@ var o_hash = {
                     let slugCounter = o_utils.getSlugOrDataTrailingCounterStr(slug);
                     slug = slugNoCounter;
 
+                    // For surfacegeometrytargetname, we have to assign the init value
```
rfrenchseti: I found this comment somewhat confusing, because I didn't remember that this function was only called on page reload.

How about "If there is a value specified for surfacegeometrytargetname in the URL, we have to initialize opus.oldSurfacegeoTarget with it so that we can properly handle changes to surfacegeometrytargetname in the future."

opus/application/static_media/js/utils.js
```diff
@@ -109,7 +109,18 @@ var o_utils = {
     // or inner objects. In our case, we only have arrays as values.
     deepCloneObj: function(obj) {
         return JSON.parse(JSON.stringify(obj));
-    }
+    },
+
+    getSurfacegeoTargetSlug: function(target) {
```
rfrenchseti, line 115: Maybe put in a comment here referencing the file in the import pipeline. Something like "This function is patterned after slug_name_for_sfc_target in import_util.py".

opus/application/static_media/js/widgets.js
```diff
@@ -93,18 +93,71 @@ var o_widgets = {
             }
         });
 
-        // close opened surfacegeo widget if user select another surfacegeo target
+        // Update surfacegeo widgets in place if user selects another surfacegeo target.
+        // 1. Update surfacegeo attributes/text in DOMs of all related surfacegeo widgets.
+        // 2. Update selections (opus.selection, opus.extras) & widgets.
```
rfrenchseti: Should be opus.selections

opus/application/static_media/js/widgets.js
```diff
@@ -93,18 +93,71 @@ var o_widgets = {
...
+                    o_widgets.updateSURFACEGEOattrInPlace(addBtns, newTargetSlug, "data-widget");
+                    o_widgets.updateSURFACEGEOattrInPlace(addBtns, newTargetSlug, "data-slug");
+                }
+            }
+
+            // Update selections & extras
+            o_widgets.updateSURFACEGEODataObj(opus.selections, oldTargetStr, newTargetStr);
+            o_widgets.updateSURFACEGEODataObj(opus.extras, oldTargetStr, newTargetStr);
+
+            // Update widgets column in opus.prefs
```
rfrenchseti: Is widgets really a "column"? Maybe "Update widgets array"?

opus/application/static_media/js/widgets.js
```diff
@@ -135,6 +188,34 @@ var o_widgets = {
...
+        let oldTargetStr = `SURFACEGEO${opus.oldSurfacegeoTarget}`;
+        let newTargetStr = `SURFACEGEO${newSurfacegeoTargetSlug}`;
+        let newTargetAttr = targetElement.attr(attribute).replace(oldTargetStr, newTargetStr);
+        targetElement.attr(attribute, newTargetAttr);
+    },
+
+    updateSURFACEGEODataObj: function(dataObj, oldTargetStr, newTargetStr) {
+        /**
+         * Update data object (opus.selections, opus.extras, and opus.currentUnitBySlug)
+         * in place with newly select target.
```
rfrenchseti: selected

opus/application/static_media/js/widgets.js
```diff
@@ -738,6 +821,14 @@ var o_widgets = {
                        $(mult_id).next().slideDown("fast");
                    }
                });
+
```
rfrenchseti: Just for future reference, these few lines above are also where we should fix issue #663 that's assigned to you.

rfrenchseti (changes requested):
- Didn't we say that if the "surface geometry target name" widget got closed that it would cause all of the other surface geometry widgets to close as well? Popping up a dialog first if any of them are populated with actual search vales?
- I don't actually understand the "issue" you mention about hint results. If I do your example of selecting Mimas, setting the latitude, and then switching to Pallene, I get a Pallene count of 0, which is correct.

juzen2003:
- Update based on comments
- Fixed #663

opus/application/static_media/js/widgets.js
rfrenchseti: Not yet fixed.

opus/application/static_media/js/widgets.js
```diff
@@ -135,6 +201,51 @@ var o_widgets = {
...
+         * target name widget is closed.
+         */
+        for (const closeWidgetIcon of $(".close_card[data-slug^='SURFACEGEO']")) {
+            let slug = $(closeWidgetIcon).attr("data-slug");
+            o_widgets.closeWidget(slug);
+            let id = "#widget__"+slug;
+            try {
+                $(id).remove();
+            } catch (e) {
+                console.log("error on close widget, id="+id);
```
rfrenchseti: We have a opus.logError function if you want to leave this in here

opus/application/static_media/js/widgets.js
```diff
@@ -41,6 +41,12 @@ var o_widgets = {
     // want to perform a search when a widget has just opened.
     isGetWidgetDone: false,
 
+    // This flag is used to determined if we are closing any widget with empty input.
```
rfrenchseti: to determine.

opus/application/static_media/js/widgets.js
```diff
@@ -41,6 +41,12 @@ var o_widgets = {
...
+    // In a series of closing widget actions, if there is one input in a widget with a
+    // value, we will make sure a search is performed after updating URL. If none of
+    // inputs has value, we don't need to perform a search.
```
rfrenchseti: none of the inputs have values

opus/application/static_media/js/widgets.js
```diff
@@ -135,6 +201,51 @@ var o_widgets = {
         o_widgets.addAttachOrRemoveInputsBehaviors();
     },
 
+    updateSURFACEGEOattrInPlace: function(targetElement, newSurfacegeoTargetSlug, attribute) {
+        /**
+         * Update surfacegeo attributes in place with newly select target
```
rfrenchseti: selected

rfrenchseti (changes requested):
- Existing functionality has broken. Before, if you selected something in the "Planet" field, the appropriate Intended Target and Surface Geo Target categories would expand. This no longer happens. This should also happen when a fresh URL is loaded with a Planet selected.
- What happened to part two of the issue, where there would be new selected metadata columns as the surface geo target changes?
- When the user clicks on the close icon for surface geo target, we had talked about warning the user that this would also close all of the associated widgets and ask the user for confirmation. If the user said no, then don't close the surface geo target widget at all. This confirmation should be asked if any there are any associated search widgets, even if they don't have values in them.

juzen2003:
- Restore the behavior to expand the related input groups when a planet is selected.
- Add new metadata selections when surfacegeo target is changed. Keep the selected metadata items related to the old target, and add new metadata items related to the new target.
- Pop out a confirmation modal when closing surfacegeometrytargetname widget if there are other related SURFACEGEO widgets open. If there is no SURFACEGEO widget open, there will be no confirmation modal popped out, and user can directly close surfacegeometrytargetname widget.

opus/application/apps/ui/templates/ui/base.html
```diff
@@ -485,6 +485,17 @@ <h3 class='op-header-text'></h3>
             draggable=False
             close_by_keyboard="esc"
         %}
+        {% include "ui/confirm_modal.html" with
+            id_name="op-close-surfacegeo-widgets"
+            confirm_title="Confirm Close All Surfacegeometry Widgets"
```
rfrenchseti: "Surfacegeometry" and "surfacegeometry" are not very friendly terms for the end user. "Widget" is not a term the user understands. We call them "search fields" in the documentation. Also it won't be clear to them why the widgets are being closed.

Title: "Confirm Close All Surface Geometry Search Fields"

Msg: "Closing the Surface Geometry Target Name search field will also close all target-specific search fields. Are you sure you wish to proceed?"

opus/application/static_media/js/widgets.js
rfrenchseti: Not fixed

rfrenchseti (changes requested):
- When adding the metadata fields, the new fields should appear at the end of the list in the same order as they were originally. So if we have Latitute [Pallene], Longitude [Pallene] and the user selects Epimetheus, we should end up with Latitute [Pallene], Longitude [Pallene], Latitute [Epimetheus], Longitude [Epimetheus]. Feel free to ask about this kind of thing before coding.
- We should also update the getting started guide. Be sure to pull the changes I made to the master branch a few minutes ago first. Here is text to add at line 120:

```
<p>The <b>Surface Geometry Target Name</b> search field is special. Only one
    surface geometry target can be selected at a time. When you select a surface
    geometry target, a new <b>[Target] Surface Geometry Constraints</b>
    category will become available providing search fields for that single
    target. If you later change the surface geometry target, all of
    the existing target-specific fields you selected will automatically switch
    to the new target but keep their existing search values. This is useful
    for cases such as searching for similar lighting geometry for different
    targets. In addition, any metadata fields you selected for the previous
    target will be duplicated for the new target. If you close the
    <b>Surface Geometry Target Name</b>, all target-specific search fields will
    also be closed.</p>
```

juzen2003:
From today's meeting:
- If there is any SURFACEGEO constraint (any value in related SURFACEGO widgets), in surfacegeometrytargetname widget, we leave green hints next to the current selected target name, but green "-" next to the non-selected ones (because they are not the result counts for the user's next click).
- If there is no SURFACEGO constraint, in surfacegeometrytargetname widget, we leave green hints next to all target names (because they are the actual result counts for the user's next click.)

rfrenchseti (changes requested):
- If you click on radio buttons rapidly, you can get the selected metadata list to screw up. This is especially true with "Pan" and "Pandora".
- The "Surface Geometry Target Name" field should never show up in the "Current Search Terms" part of select metadata.
- It's possible to-deselect all radio buttons by clicking just to the left of the selected button.

PR #965, opened 2020-01-22
872 ui change for surface geometry searching 

### LINKED ISSUE #872: UI changes for surface geometry searching

juzen2003:
- Fixes #872, Related to this pull: https://github.com/SETI/pds-opus/pull/956
  - Database used: opus3_test_200107 (obs_surface_geometry_target.json is modified)
  - All Django tests pass: Y (some tests are modified and added)

Description of changes:
- Fixed the issue of blank rows in selected metadata list when switching between "Pan" and "Pandora" in surfacegeo target name.
- Fixed the issue of all radio buttons being de-selected when surfacegeo target widget is dragged.
- Remove "Surface Geometry Target Name" from the "Current Search Terms" part of select metadata.
- Make sure new input can be added/removed after selecting a new surfacegeo target.

Known problems: (Fixed)
- From https://github.com/SETI/pds-opus/pull/956
    - If you click on radio buttons rapidly, you can get the selected metadata list to screw up. This is especially true with "Pan" and "Pandora".
    - The "Surface Geometry Target Name" field should never show up in the "Current Search Terms" part of select metadata.
    - It's possible to-deselect all radio buttons by clicking just to the left of the selected button. 
- If the user switches to a new surfacegeo target, the add input button will not work in related SURFACEGEO widgets after DOM attributes are updated in place.


opus/application/apps/search/views.py
```diff
@@ -452,16 +452,19 @@ def url_to_search_params(request_get, allow_errors=False, return_slugs=False,
...
+                if clause_num > 0:
+                    slug = slug[:slug.rindex('_')]
+                else:
                     raise ValueError
             except ValueError:
-                log.error('url_to_search_params: Slug has illegal clause '+
-                          'number "%s"', orig_slug)
-                return None, None
+                # If clause_num is not a positive integer, leave the slug as is.
+                # If the slug is unknown, it will be catched later as the
```
rfrenchseti: catched => caught and "as the" => "as an"

opus/application/apps/ui/views.py
```diff
@@ -591,17 +592,18 @@ def _sort_func(key):
...
+                    slug = slug[:slug.rindex('_')]
+                else:
                     raise ValueError
             except ValueError:
-                msg = ('Search term "' + escape(orig_slug)
-                       + '" has a bad clause number; it has been ignored.')
-                msg_list.append(msg)
-                continue
+                # If clause_num is not a positive integer, leave the slug as is.
+                # If the slug is unknown, it will be catched later as the
```
rfrenchseti: catched => caught and "as the" => "as an"

opus/application/static_media/js/browse.js
```diff
@@ -1891,6 +1893,9 @@ var o_browse = {
         // then after the load is complete, instead of hiding the galleryView slide, update the metadata.
         let updateMetadataBox = $("#op-select-metadata").hasClass("show") && $("#galleryView").hasClass("show") ;
 
+        // Note: Increment the reqno here instead of getDataURL because infiniteScroll path also use getDataURL,
```
rfrenchseti: also use => also uses

rfrenchseti: reqno for infiniteScroll path => reqno for the infiniteScroll URL

opus/application/static_media/js/opus.js
```diff
@@ -372,7 +372,11 @@ var opus = {
 
         // Finally, update all the hints
         $.each(opus.prefs.widgets, function(index, slug) {
-            o_search.getHinting(slug);
+            if (slug === "surfacegeometrytargetname" && !o_search.isAllSURFACEGEOSelectionsEmpty()) {
```
rfrenchseti: I think the grammar would be better as "areAll..."

opus/application/static_media/js/search.js
```diff
@@ -556,6 +559,20 @@ var o_search = {
         return isInputSetEmpty;
     },
 
+    isAllSURFACEGEOSelectionsEmpty: function() {
+        let isInputSetEmpty = true;
+        for (const slug in opus.selections) {
+            if (slug.match(/^SURFACEGEO/)) {
```
rfrenchseti: slug.startsWith?

opus/application/static_media/js/search.js
```diff
@@ -556,6 +559,20 @@ var o_search = {
...
+                isInputSetEmpty = o_search.isSlugSelectionsEmpty(opus.selections[slug], isInputSetEmpty);
```
rfrenchseti: is => are

opus/application/static_media/js/search.js
```diff
@@ -1112,7 +1129,7 @@ var o_search = {
         }); // end mults ajax
     },
 
-    getValidMults: function(slug) {
+    getValidMults: function(slug, hideHintsForSurfacegeoTarget=false) {
```
rfrenchseti: This doesn't actually test for surfacegeotarget and works for any widget with radio buttons, right? So maybe it should be called "hideHintsForNonSelectedRadioButtons" or something like that.

opus/application/static_media/js/widgets.js
```diff
@@ -84,27 +99,98 @@ var o_widgets = {
...
                 }
-            });
+
+                // Update selections & extras
+                o_widgets.updateSURFACEGEODataObj(opus.selections, oldTargetStr, newTargetStr);
+                o_widgets.updateSURFACEGEODataObj(opus.extras, oldTargetStr, newTargetStr);
+
+                // Update widgets array in opus.prefs
+                $.each(opus.prefs.widgets, function(widgetIdx, widget) {
+                    if (widget.match(oldTargetStr)) {
```
rfrenchseti: startsWith?

opus/application/static_media/js/widgets.js
```diff
@@ -84,27 +99,98 @@ var o_widgets = {
...
+                        let newWidget = widget.replace(oldTargetStr, newTargetStr);
+                        opus.prefs.widgets[widgetIdx] = newWidget;
+                    }
+                });
+
+                // Update metadata selections
+                let currentMetadataSelections = opus.prefs.cols.slice();
+                let newSurfacegeoMetadataSelections = [];
+                for (const col of currentMetadataSelections) {
+                    if (col.match(oldTargetStr)) {
```
rfrenchseti: startsWith?

opus/application/static_media/js/widgets.js
```diff
@@ -135,6 +221,72 @@ var o_widgets = {
...
+    },
+
+    updateSURFACEGEODataObj: function(dataObj, oldTargetStr, newTargetStr) {
+        /**
+         * Update data object (opus.selections, opus.extras, and opus.currentUnitBySlug)
+         * in place with newly selected target.
+         * NOTE: the passed in dataObj will be modified.
+         */
+        for (const oldTargetSlug in dataObj) {
+            if (oldTargetSlug.match(oldTargetStr)) {
```
rfrenchseti, line 259: startsWith?

opus/application/test_api/test_ui_api.py
```diff
@@ -2371,8 +2451,8 @@ def test__api_normalizeurl_real_5(self):
...
+        url = '/opus/__normalizeurl.json?SURFACEGEOpluto_centerresolution2=5&SURFACEGEOpluto_phase1=160&surfacegeometrytargetname=Pluto&qtype-SURFACEGEOpluto_phase=any&cols=opusid,instrument,time1,SURFACEGEOpluto_planetographiclatitude1,SURFACEGEOpluto_planetographiclatitude2,SURFACEGEOpluto_IAUwestlongitude1,SURFACEGEOpluto_IAUwestlongitude2,SURFACEGEOpluto_centerdistance,SURFACEGEOpluto_centerresolution,SURFACEGEOpluto_phase1,SURFACEGEOpluto_phase2&widgets=SURFACEGEOpluto_centerresolution,SURFACEGEOpluto_phase,surfacegeometrytargetname&order=time1,opusid&view=search&browse=gallery&cart_browse=gallery&startobs=1&cart_startobs=1&detail='
```
rfrenchseti: I think you want to leave the old SURFACEGEO formats here (without the "_") in the URL so that we test normalizeurl's ability to switch from the old slug format to the new one.

rfrenchseti (changes requested):

opus/application/static_media/js/widgets.js
juzen2003: For this one, we will stay with .match because it's used to update the dataObj, including opus.extras, and there are "qtype-SURFACEGEO.*" and "unit-SURFACEGEO.*" that have "SURFACEGEO" in the middle of the string. So we will use .match, instead of .startsWith.

opus/application/test_api/test_ui_api.py
rfrenchseti: This wasn't entirely fixed so I went ahead and removed the remaining two "_" that shouldn't be there.

rfrenchseti (approved):

rfrenchseti:
- Database used: opus3_test_200107 (modified for new surface geo slugs)
- All Django tests pass: Y

Issue #855, opened 2019-09-20
Help dialogs: Closing (x) not vertically aligned with text


rfrenchseti:
In any of the "Help" dialogs, the (x) to close the dialog is not vertically aligned with the dialog header text.

PR #970, opened 2020-01-29
Navbar padding, help panel (x), copy mults (issue 855,962)

### LINKED ISSUE #962: Nav bar loses horizontal padding when browser is narrow
When you shrink the browser so that the nav bar loses the "Recent Announcements" section, the horizontal padding between the Search/Browse/Cart/Detail tabs also goes away to the point where it's pretty ugly. They should maintain a reasonable spacing. I expect using the original spacing will work fine even when the browser is as narrow as we allow.

juzen2003:
- Fixes #962, #855 
  - All Django tests pass: Y

Description of changes:
- Fixed the issue to keep the spacing between Search/Browse Results/Cart/Detail the same even when the browser gets narrower.
- Fixed the issue that the position of "OPUS" (on the top left corner) will jump a bit when browser gets narrow.
- Make input texts and hints copiable in widgets with checkboxes and radio buttons.  
- Make sure "x" in the help panel is vertically aligned with the header text.

Known problems:
- Hold & copy texts and hints for checkboxes and radio buttons will not work in iOS . (It should work by default, haven't found the solution for this.)

### DIFF
--- a/opus/application/static_media/js/widgets.js
+++ b/opus/application/static_media/js/widgets.js
@@ -51,6 +51,7 @@ var o_widgets = {
         $("#op-search-widgets").sortable({
             items: "> li",
             cursor: "move",
+            cancel: ".multichoice span, .multichoice label, .singlechoice span, .singlechoice label",
             // we need the clone so that widgets in url gets changed only when sorting is stopped
             // Note: this will make radio buttons deselected when a widget with radio buttons is dragged.
             // We have to restore radio button checked status in stop event handler.
--- a/opus/application/static_media/js/widgets.js
+++ b/opus/application/static_media/js/widgets.js
@@ -51,6 +51,8 @@ var o_widgets = {
         $("#op-search-widgets").sortable({
             items: "> li",
             cursor: "move",
+            // Specify the elements that will not trigger sorting when being grabbed. This is to make
+            // input texts and hints copiable in widgets with checkboxes and radio buttons.
             cancel: ".multichoice span, .multichoice label, .singlechoice span, .singlechoice label",
             // we need the clone so that widgets in url gets changed only when sorting is stopped
             // Note: this will make radio buttons deselected when a widget with radio buttons is dragged.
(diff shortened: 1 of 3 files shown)


rfrenchseti (changes requested):
- The OPUS3 icon moves up and down slightly as the browser changes width.

rfrenchseti (changes requested):
The size of the Search/Browse/Cart/Detail buttons has changed dramatically.

opus/application/static_media/js/widgets.js
```diff
@@ -51,6 +51,9 @@ var o_widgets = {
         $("#op-search-widgets").sortable({
             items: "> li",
             cursor: "move",
+            // Specify the elements that will not trigger sorting when being grabbed. This is to make
```
rfrenchseti: Debby has a pull that prevents drag on anything except the widget title. I think that eliminates the need for this change.

rfrenchseti (changes requested):

rfrenchseti (changes requested):
Also, I know it's not one of these bugs, but while you're at it can you vertically center the (X) in the Select Metadata dialog?

rfrenchseti:
The black navbar is slightly shorter than it used to be. Is that intentional? Question asked on opus-dev-chat to see which people prefer.

rfrenchseti:
I verified the mult text copy functionality with Debby's recently merged pull and it does the correct thing.

rfrenchseti:
The Select Metadata dialog header is now slightly taller, but it is consistent with the other modal dialogs (previously it was shorter than the others).

I made the nav bar the same height it used to be because I think it looks better a little taller. To do this I added:

```
/* Make the main nav bar just a little taller than normal (style says 8px) */
#op-main-nav {
    padding-top: 10px;
    padding-bottom: 10px;
}
```

rfrenchseti (approved):

rfrenchseti:
- Database used: N/A
- All Django tests pass: NA
- JSHINT run on all affected files: N/A

PR #971, opened 2020-01-31
Allow clicking on whitespace to right of categories and menu options (issue 729)

### LINKED ISSUE #729: Allow clicking on whitespace to right of categories and menu options
B016: In Select Metadata, you can click on the white space to the right of a metadata field name to select it, but not to the right of a category name to expand it. Should be able to click there as well.
S039: Since we can click on whitespace to the right of entries in Select Metadata, we should allow it on the Search page too (both widget names and categories)

juzen2003:
- Fixes #729 
  - All Django tests pass: Y
  - JSHINT run on all affected files: Y (menu.js)

Description of changes:
- Make the whitespace to right of categories and menu options clickable in search sidebar menu and select metadata menu.

### DIFF
--- a/opus/application/static_media/css/opus.css
+++ b/opus/application/static_media/css/opus.css
@@ -703,14 +703,13 @@ button.op-mini-thumbnail-zoom {
     right: 1.25em;
 }
 
-/* Make sure the a tags for categories take up full width in select metadata menu */
-.op-search-menu-category > a {
-    display: flex;
+/* Make sure the a tags for categories take up full width in select metadata
+and search sidebar menu */
+a[href^="#selector-submenu-search"], a[href^="#selector-submenu-obs"], 
+a[href^="#search-submenu-obs"] {
+    display: block;
     margin-top: 0 !important;
     margin-bottom: 0 !important;
-    flex: 0 0 100%;
-    /* Vertically align arrow icon with the category text */
-    align-items: center;
 }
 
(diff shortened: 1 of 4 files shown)


opus/application/static_media/css/opus.css
```diff
@@ -703,6 +703,15 @@ button.op-mini-thumbnail-zoom {
     right: 1.25em;
 }
 
+/* Make sure the a tags for categories take up full width in select metadata
+and search sidebar menu */
+a[href^="#selector-submenu-search"], a[href^="#selector-submenu-obs"],
```
rfrenchseti: Would it be better to add a class to these instead of using the ID?

juzen2003: Added a class "op-submenu-category" to these anchor tags.

rfrenchseti (changes requested):
There's too much whitespace above the first category in the Select Metadata dialog. Flip your tab from your new version of OPUS to the one on tools to see the difference. Also it seems that the fields are a little closer together vertically than they used to be.

rfrenchseti (changes requested):
I have no idea why, but now in the Select Metadata dialog, when you expand a category on the left, the scrollbar height doesn't update. When you nudge the scrollbar just a tiny bit, then it updates.

rfrenchseti (approved):

PR #973, opened 2020-01-31
Prevent mini-thumbnails from disappearing when the width of the screen shrinks

### LINKED ISSUE #960: Thumbnails in table view disappear when browser is narrow
The mini-thumbnails in the table view are not a fixed width. As the number of columns in the table increases, or the browser width decreases, the thumbnails reach a point where they start to shrink, and soon disappear entirely. This is not a problem with responsiveness, per se, because it doesn't depend on an absolute browser width. Instead it depends on how wide the table is.

dstopp:
- Fixes #960 
  - All Django tests pass: NA
  - Tested for race conditions (with delayed API calls if applicable):NA

Description of changes:
 - Set the height to a specific value instead of a percentage.  This takes care of both the disappearing image and the height being diff sizes when an image is taller.

Known problems:
 - none

### DIFF
--- a/opus/application/static_media/css/opus.css
+++ b/opus/application/static_media/css/opus.css
@@ -623,7 +623,7 @@ th.sticky-header {
 
 .op-mini-thumbnail img {
     transition: transform .1s;
-    width: 68%;
+    height: 21px;
 }
 
 /* zoom the mini-thumbnail on hover */


rfrenchseti (approved):

rfrenchseti:
- Database used: N/A
- All Django tests pass: NA
- Tested on Chrome, Firefox, Safari, and iOS: Chrome, Firefox, iOS

PR #978, opened 2020-02-04
Remember nav bar clicks during normalizeinput so we can process them later

### LINKED ISSUE #853: Clicking on browse results after bad input changes to good doesn't work
The current input validation works like this:

- Validate as the user types
- On focus out, disable the browse results tab. This is because we can't allow the user to click on it and change to the browse page when we aren't certain the search is going to be valid yet. We don't know because the final normalize inputs call hasn't returned.
- When the final normalize inputs call returns and everything is OK, enable the browse results tab again.

This works fine in the normal cases, like when a person leaves the focus of an input by clicking in another widget, elsewhere on the page, hitting enter, or hitting tab. However, if the user leaves focus by actually clicking on the browse results tab, that click will be lost because the tab will be disabled by the focus out because the click is registered for the tab (which is as it should be). However, the click won't be remembered when the normalize inputs call comes back OK. The end result is the user will have to click on the tab twice to make it work.

I don't think the average user will ever encounter this.

rfrenchseti:
- Fixes #853 
  - Database used: N/A
  - All Django tests pass: NA

Description of changes:

Previously we simply turned off clicks for the browse tab when calling normalizeinput, since if the inputs came back invalid we didn't want to switch tabs. Now we remember which tab the user clicked on, and when normalizeinput is done we jump to that tab, if any. Note we now do this for all tabs, not just browse, so that when the user has an illegal input we don't let them leave the search tab.

Known problems:

None.

### DIFF
--- a/opus/application/static_media/css/opus.css
+++ b/opus/application/static_media/css/opus.css
@@ -2493,12 +2493,6 @@ wiggling when mouse over autocomplete menu items */
     margin: 0;
 }
 
-/* Use to disable nav link (for the case that disables browse tab when
-we have an invalid search) */
-.op-disabled-nav-link {
-    pointer-events: none;
-}
-
 /* Styling for the button to delete a set of inputs in a widget */
 .op-remove-inputs {
     margin-left: 10px;
--- a/opus/application/static_media/js/opus.js
+++ b/opus/application/static_media/js/opus.js
@@ -94,6 +94,11 @@ var opus = {
     // An object that stores normalize input validation result for each range input.
     rangeInputFieldsValidation: {},
 
+    // Remember nav clicks during normalize input so we can actually process them
+    // later if the inputs are valid.
+    navLinksNormalizeInProgress: false,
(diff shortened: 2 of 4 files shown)


juzen2003 (approved):

juzen2003:
One issue: (Only happened when there is a delay that allows user to perform multiple actions between the return of normalize input and switching to browse tab)
Here are steps to reproduced the issue:
1. Add an API (~3000) delay.
2. Open a range input widget and add another input set to have multiple input sets in the widget.
3. Type something good in a range input and focus out to click Browse tab, but then move to another input quickly.
4. In another input set, quickly type something bad. 
5. OPUS will go to browse tab with the search constraints in step 3 (URL with search terms from step 2) but without search constraints from step 4. There will be no image displayed and results will be “?” in browse tab.
6. And then we will stuck in browse tab, can’t go to any other tabs.

rfrenchseti:
OK I think I fixed that bug. Please look at the code and try to break it again.

PR #979, opened 2020-02-05
Tooltip support for mobile devices. Fixes #796.

### LINKED ISSUE #796: Need tooltips on click not just on hover for mobile devices
Right now (i) tooltips only show up on mouse-over hover, making them impossible to read on touch-only screens. The easy fix is to make them show up on both hover and click.

Here's one possible solution:

https://gist.github.com/csasbach/867744

It might be better to use a fancier tooltip library in general:

http://qtip2.com/

rfrenchseti:
- Fixes #796
  - Database used: N/A
  - All Django tests pass: N/A
  - Tested on Chrome, Firefox, Safari, and iOS: Y (and Android)
  - Tested for race conditions (with delayed API calls if applicable): N/A

Description of changes:

When a touch event happens, a handler is added to each (i) element that has a title. When that (i) is finger-clicked, the tooltip text shows up top-center in the browser window. Any subsequent touch of any kind makes it go away, unless it's another (i) in which case the text is replaced (well first it goes away, then it comes back again). I didn't make the tooltip show up near the (i) that was touched because 1) it was really painful to make that work and 2) this solves all problems about tooltips running off the edge of the screen and having to be intelligently oriented. I made a fairly obvious border so people would see it there.

Known problems:

None.

### DIFF
--- a/opus/application/static_media/css/opus.css
+++ b/opus/application/static_media/css/opus.css
@@ -2611,6 +2611,26 @@ at the right of -- OR -- label. */
     padding-right: 3px;
 }
 
+/* Tooltips */
+
+[title] {
+    cursor: default;
+}
+
+/* This is for touch screens */
+.op-tooltip-text {
+    position: fixed;
+    top: 80px;
+    left: 10%;
+    right: 10%;
+    border: 2px solid rgba(0, 0, 0, 0.2);
+    background-color: white;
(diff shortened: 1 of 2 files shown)


juzen2003 (approved):
- All Django tests pass: Y
- JSHINT run on all affected files: Y (opus.js)

PR #985, opened 2020-02-06
Catch server errors in UI and generate fake errors in backend (Fixes #590, Fixes #608)

### LINKED ISSUE #590: Add error handling to all AJAX calls
The Django API returns 404 or 500 errors for many possible problems, but the OPUS JS code rarely tests for those. It should always check the return status and report an error to the user.

### LINKED ISSUE #608: get_search_results_chunk needs to return better error message
Currently get_search_results_chunk logs something detailed to the log file, but just returns None. This results in the caller often returning a useless or incorrect error to the end user.

rfrenchseti:
- Fixes #590 
- Fixes #608 
  - All Django tests pass: Y
  - Tested for race conditions (with delayed API calls if applicable): N/A

Description of changes:

- Cleaned up creation of HTTP 404 and 500 errors in the back end and updated all tests.
- Added ability to randomly simulate server errors. New fields required in opus_secrets.py (see template):

```
# Allow random throwing of Http404 or Http500 errors
OPUS_FAKE_SERVER_ERROR404_PROBABILITY = .2
OPUS_FAKE_SERVER_ERROR500_PROBABILITY = 0.
```

- Catch all server errors in UI and display a modal dialog with the error message and allow user to restart OPUS.

Known problems:

Does not always catch server errors during infinite scroll actions.


opus/application/apps/cart/test_cart.py
```diff
@@ -119,13 +132,24 @@ def test__api_reset_session_no_get(self):
...
+            r'Internal error \(No request was provided\) for /__cart/download.json'):
             api_create_download(request)
+
+    def test__api_create_download_opusid_no_get(self):
+        "[test_cart.py] api_create_download: no GET"
+        c = Client()
+        request = self.factory.get('/__cart/download.json')
+        request.GET = None
+        with self.assertRaisesRegex(Http404,
+            r'Internal error \(No request was provided\) for /api/download/testopusid.json'):
```
juzen2003: /api/download/testopusid.**zip**? I think we only support .zip format.

opus/application/apps/cart/views.py
```diff
@@ -463,7 +477,10 @@ def api_create_download(request, opus_id=None):
     api_code = enter_api_call('api_create_download', request)
 
     if not request or request.GET is None:
-        ret = Http404(settings.HTTP404_NO_REQUEST)
+        if opus_id:
+            ret = Http404(HTTP404_NO_REQUEST(f'/api/download/{opus_id}.json'))
```
juzen2003: I think f'/api/download/{opus_id}.json' doesn't match any API URL pattern. Only opus_id.zip format has a matched pattern.

opus/application/apps/cart/views.py
```diff
@@ -395,26 +408,27 @@ def api_reset_session(request):
...
     try:
         recycle_bin = int(recycle_bin)
     except:
-        log.error('api_reset_session: Bad value for recyclebin %s: %s', recycle_bin,
-                  request.GET)
-        ret = HttpResponseServerError(settings.HTTP500_INTERNAL_ERROR)
+        log.error('api_reset_session: Bad value for recyclebin %s: %s',
+                  recycle_bin, request.GET)
+        ret = HttpResponseServerError(
+                                HTTP404_BAD_RECYCLEBIN(recycle_bin, request))
```
juzen2003: Should we use ```Http404(HTTP404_BAD_RECYCLEBIN(recycle_bin, request))```? Might need to do the similar changes like line 314, 315, and 321.

opus/application/apps/cart/views.py
```diff
@@ -1318,16 +1347,14 @@ def _csv_helper(request, opus_id, api_code=None):
...
 def _create_csv_file(request, csv_file_name, opus_id, api_code=None):
     "Create a CSV file containing the cart data."
-    column_labels, page = _csv_helper(request, opus_id, api_code)
-    if column_labels is None:
-        ret = Http404(settings.HTTP404_UNKNOWN_SLUG)
-        exit_api_call(api_code, ret)
-        raise ret
+    column_labels, page, error = _csv_helper(request, opus_id, api_code)
+    if error is not None:
+        return get_search_results_chunk_error_handler(error, api_code)
```
juzen2003, line 1359: Do we want add the similar check here like what's been done at line 210?
```
if column_labels is None or throw_random_http404_error():
    ret = Http404(HTTP404_UNKNOWN_SLUG(None, request))
    exit_api_call(api_code, ret)
    raise ret
```

opus/application/apps/ui/views.py
```diff
@@ -107,7 +107,7 @@ def api_get_menu(request):
     reqno = get_reqno(request)
     if reqno is None:
```
juzen2003: Do we want to add ```or throw_random_http404_error()``` here to simulate possible 404 error like what's been done in other places when checking reqno? (same at line 155)

opus/application/apps/ui/views.py
```diff
@@ -72,7 +72,7 @@ def api_last_blog_update(request):
     api_code = enter_api_call('api_last_blog_update', request)
 
     if not request or request.GET is None:
-        ret = Http404(settings.HTTP404_NO_REQUEST)
+        ret = Http404(HTTP404_NO_REQUEST('/__lastblogupdate.json'))
```
juzen2003, line 76: Nothing wrong with this line, but just realize we didn't import Http404 in this file. We don't have to do so because we did import everything from tools/app_utils (and this is done in app_utils ```from django.http import Http404, HttpResponse```). But just to make it consistent with other view.py files, maybe we want to add the import here too.

opus/application/test_api/test_results_api.py
```diff
@@ -174,453 +188,479 @@ def test__api_data_bad_cols4_json(self):
...
+        self._run_status_equal(url, 404,
+                        HTTP404_UNKNOWN_RING_OBS_ID(None,
+                                '/api/metadata/S_IMG_VG2_ISS_4360846_N.json'))
 
     def test__api_metadata2_bad_opusid_json(self):
         "[test_results_api.py] /api/metadata_v2: bad opusid json"
-        url = '/opus/api/metadata_v2/vg-iss-2-s-,c4360845.json'
-        self._run_status_equal(url, 404, settings.HTTP404_UNKNOWN_OPUS_ID)
+        url = '/api/metadata_v2/vg-iss-2-s-,c4360845.json'
+        self._run_status_equal(url, 404)
```
juzen2003, line 566: Pass in ```HTTP404_UNKNOWN_OPUS_ID('vg-iss-2-s-,c4360845.json', '/api/metadata_v2/vg-iss-2-s-,c4360845.json')``` as the third argument? (similar at line 585)

juzen2003 (changes requested):
- All Django tests pass: Y

opus/application/test_api/test_results_api.py
rfrenchseti: Actually in this case, no. The URL with the comma in it doesn't match the Django regex because we only allow letters, numbers, and dashes in opus ids. So Django itself throws a plain 404 error with no special message.

rfrenchseti (commented):
All comments addressed

PR #990, opened 2020-02-08
Issue of selecting text nodes in mults widgets

juzen2003:
  - All Django tests pass: Y

Description of changes:
Make sure we can click and drag to copy input and hints text nodes in mults widgets. Default mousedown behavior will not be disabled when clicked targets are input and hints text nodes.

Known problems:
None


rfrenchseti (changes requested):
Can you also make it so the min/max/null hints are selectable?

rfrenchseti (approved):

PR #1000, opened 2020-02-20
Remove interval timer

### LINKED ISSUE #980: Get rid of main OPUS timer
OPUS currently uses a timer to look for changes in the selections to trigger a search. It really should use events instead. When a search term is changed, set a timer to trigger a search in N milliseconds. If another search term is changed before that time expires, delete the timer and set a new one. That way it will never trigger a search until the user has finished changing things.

rfrenchseti:
- Fixes #980 
  - Database used: N/A
  - All Django tests pass: N/A

Description of changes:

Removed 1 second interval timer and replaced with a single-shot timer that is started every time the user changes something related to search.

Known problems:

None

### DIFF
--- a/opus/application/static_media/js/search.js
+++ b/opus/application/static_media/js/search.js
@@ -239,7 +239,6 @@ var o_search = {
             Update URL (and search) if all inputs are valid
         */
         $("#search").on("change", "input.RANGE", function(e) {
-            console.log("search 242", opus.selections, opus.extras);
             let inputName = $(this).attr("name");
             let slugName = $(this).data("slugname");
             let slug = o_utils.getSlugOrDataWithoutCounter(inputName);
@@ -574,7 +573,6 @@ var o_search = {
             $(target).removeClass("input_currently_focused");
         }
 
-        console.log("search 576", opus.selections, opus.extras);
         o_search.parseFinalNormalizedInputDataAndUpdateURL(slugWithId, url);
     },
 
@@ -880,7 +878,6 @@ var o_search = {
          * Validate the return data from a normalize input API call, and update hash & URL
          * based on the selections for the same normalize input API.
          */
-        console.log("search 882", normalizedInputData, opus.selections, opus.extras);
         o_search.slugInputValidValueFromLastSearch = {};
         $.each(normalizedInputData, function(eachSlug, value) {
             if (eachSlug === "reqno") {
@@ -943,7 +940,6 @@ var o_search = {
             }
         });
 
-        console.log("search 945", opus.selections, opus.extras);
         // Update newly selected unit to currentUnitBySlug
         if (slug.startsWith("unit-") && unit) {
             let slugNoNum = slug.match(/unit-(.*)$/)[1];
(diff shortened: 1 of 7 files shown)


juzen2003:
- All Django tests pass: Y

Note: The change looks good, haven't found any bad behaviors yet. Just two minor things:
1. Add some delay, like 3 seconds.
2. Open up a RANGE or STRING widget, like the volume id widget (wait until everything is settled).
3. In volume id, type something like "C" and leave focus.
4. Before 1 second timer expires, quickly focus in the same input, remove "C" and focus out.
5. A search will be performed (both result & widget spinner will be running, but result will not change because search selections are the same). Result spinner will only run if we try 3-4 the first time when a widget is open. Do we want to prevent result spinner from running in this case because search selections are the same? -> this is the 1st one.
6. If we repeat 3-4 again, only the widget spinner will be running, but result spinner won't run (this is expected). Do we want to hide the widget spinner if no search is run? -> this is the 2nd one.

rfrenchseti:
I fixed a few bugs I found in the existing code, include the two mentioned above:

- Make string widgets spinner hidden by default since there are no hints to wait for
- When turning on widget spinners after a search change, only turn on those for range and mult widgets
- Fixed bug with validateinput return overwriting input field. This happened when an input field was changed, validateinput was called, and then the field was changed back to its original value, which did *not* trigger another called to validateinput. Then when the first call eventually returned, it overwrote the input value since there was no higher reqno telling it not to.
- Don't turn on widget spinners if search didn't change unless we need to replace ? fields. Previously we always turned on the spinners and got new hinting data even if we didn't need to.

opus/application/static_media/js/search.js
```diff
@@ -185,19 +185,6 @@ var o_search = {
...
-            2) Input value didn't change from the last successful search
-            3) Input value didn't match any ranges names
-            */
-            if (currentValue === "" || currentValue === o_search.slugInputValidValueFromLastSearch[slugWithId] ||
-                !o_search.performInputValidation[slugWithId]) {
-                $(e.target).removeClass("search_input_valid search_input_invalid");
-                $(e.target).removeClass("search_input_invalid_no_focus");
-                $(e.target).addClass("search_input_original");
-                return;
-            }
```
juzen2003, line 188: Since the above piece is removed, only one last minor bug exists:
1. Add a delay, like 3 seconds.
2. Open a range widget, for example, observation duration.
3. In min input, type something invalid, like "9x" and wait until yellow border and grey background show up.
4. Quickly remove "9x", and focus out min input. Sometimes red background will not go away.

The reason is because at step 4, no search will be performed but a normalize input api will be called in "input" event and update the border (line 205-217) after the return of api call. At the same time, "focusout" event will also update the border based on whether the input has a class "search_input_invalid" or "search_input_valid" (line 129-136); however this could happen before the return of normalize input api in "input" event and the input background will have the wrong color in this case. The fix for this is to add the following between line 206 & 207:
`$(e.target).removeClass("search_input_invalid_no_focus");`

juzen2003 (changes requested):

Issue #852, opened 2019-09-18
Welcome splash dialog does not work with short browsers


rfrenchseti:
The welcome splash dialog isn't responsive and starts to exceed the browser height at around 575px.

rfrenchseti:
We should probably just not display the dialog at all on browsers that short.

rfrenchseti:
Fixed to be responsive.

PR #1001, opened 2020-02-21
Responsive issues (Fixes #800)

### LINKED ISSUE #800: Remaining responsiveness issues (#2)
- When there is a <NEW> icon in the main nav bar next to Recent Announcements, and you shrink the browser horizontally so that it goes into the hamburger, the vertical size of the nav bar changes.
- On the search tab, when you shrink the browser so that the search field names start to wrap, the wrapped lines don't maintain the same level of indent as the first line - they are further to the left. They should remain lined up with the text (and to the right of the (i)).
- In widget titles, as the browser shrinks and the title wraps, the text should remain lined up with itself and to the right of (i).

juzen2003:
- Fixes #800 
  - All Django tests pass: Y

Description of changes:
- In search sidebar and metadata selector menu, when menu items start to wrap, make sure they are aligned with the text, not the (i) icon.
- In search tab, group qtype, (i) icon, and trash icon. They will move to a different line together when browser shrinks.
- In search widgets, group Min:/Max: and their corresponding input tags. Min:/Max: will stay together with their corresponding input tags when browser shrinks and they move to a different line.
- When search widget title starts to wrap, make sure it's aligned with the text instead of the (i) icon.
- In detail tab, group the value and the magnifying glass. They will move together to a different line when browser shrinks.
- Fixed the height of main nav bar when browser width changes.
  - Fixed in https://github.com/SETI/pds-opus/pull/970
- Help menu fit in the screen when browser height is small.
  - Fixed in #852 

Known problems:
- Note: In select metadata menu (different looks in UI):
  - To make the text aligned, we have to set .op-menu-item display to flex. Once this is done, the height of the span tag (.op-menu-item) will be a bit larger, and we have to add margin-top & bottom (1px for now) to create some spacing between each menu item. This will make the height of menu items larger than ones in production sites. 
  - In select metadata menu, when window is narrow, and if a menu item text is wrapped into the 2nd line, the highlighted background will cover the entire row for that menu item. (Match the behavior in search sidebar now)

### DIFF
--- a/opus/application/apps/results/templates/results/detail_metadata_internal.html
+++ b/opus/application/apps/results/templates/results/detail_metadata_internal.html
@@ -16,13 +16,13 @@
                                 {{ param_info.label_results }}:&nbsp;
                             {% endif %}
                             <span class="op-detail-entry-values">{{ val }}
-                            {% if param_info.is_string_or_mult %}
-                                <a href="/opus/#/{{ param_info.slug }}={{ val|encode_value }}{% if param_info.is_string %}&qtype-{{ param_info.slug }}=matches{% endif %}{% if url_cols %}&cols={{ url_cols }}{% endif %}"
-                                        target="_blank">
-                                    <i class="fas fa-search"
-                                        title='Open a new browser tab containing OPUS that searches for "{{ param_info.label_results }}" equal to "{{ val }}"'></i>
-                                </a>
-                            {% endif %}
+                                {% if param_info.is_string_or_mult %}
+                                    <a href="/opus/#/{{ param_info.slug }}={{ val|encode_value }}{% if param_info.is_string %}&qtype-{{ param_info.slug }}=matches{% endif %}{% if url_cols %}&cols={{ url_cols }}{% endif %}"
+                                            target="_blank">
+                                        <i class="fas fa-search"
+                                            title='Open a new browser tab containing OPUS that searches for "{{ param_info.label_results }}" equal to "{{ val }}"'></i>
+                                    </a>
+                                {% endif %}
                             </span>
                         </div>
                     {% endif %}
--- a/opus/application/apps/results/templates/results/detail_metadata_internal.html
+++ b/opus/application/apps/results/templates/results/detail_metadata_internal.html
@@ -11,10 +11,11 @@
                         {% endif %}
                         <div>
                             {% if param_info.get_units %}
-                                {{ param_info.label_results }} {{ param_info.get_units }}: {{ val }}
+                                {{ param_info.label_results }} {{ param_info.get_units }}:&nbsp;
                             {% else %}
-                                {{ param_info.label_results }}: {{ val }}
+                                {{ param_info.label_results }}:&nbsp;
                             {% endif %}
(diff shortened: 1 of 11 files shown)


opus/application/apps/results/templates/results/detail_metadata_internal.html
```diff
@@ -11,17 +11,19 @@
...
                         <div>
                             {% if param_info.get_units %}
-                                {{ param_info.label_results }} {{ param_info.get_units }}: {{ val }}
+                                {{ param_info.label_results }} {{ param_info.get_units }}:&nbsp;
                             {% else %}
-                                {{ param_info.label_results }}: {{ val }}
+                                {{ param_info.label_results }}:&nbsp;
                             {% endif %}
+                            <span class="op-detail-entry-values">{{ val }}
                             {% if param_info.is_string_or_mult %}
```
rfrenchseti: I think this if block should be indented

opus/application/apps/results/templates/results/detail_metadata_slugs_internal.html
```diff
@@ -8,14 +8,16 @@
                         title="{{ values.1.get_tooltip_results|striptags }}"></i>&nbsp;
                 {% endif %}
                 <div>
-                    {{ param }}: {{ values.0 }}
+                    {{ param }}:&nbsp;
+                    <span class="op-detail-entry-values">{{ values.0 }}
                     {% if values.1.is_string_or_mult %}
```
rfrenchseti: I think this if block should be indented

opus/application/static_media/css/opus.css
```diff
@@ -249,10 +260,31 @@ Not sure if this is the desired behavior
...
+        flex: 1 0 auto;
+    }
+
+    .op-range-input-min-list, .op-range-input-max-list {
+        white-space: nowrap;
+        display: inline-flex !important;
         width: 100%;
     }
+
+    /* Algin the Min/Max input tag when they are in different lines */
```
rfrenchseti: Align

opus/application/static_media/css/opus.css
```diff
@@ -2671,9 +2710,22 @@ at the right of -- OR -- label. */
     }
 }
 
-/* Make sure label names & hints in wdigets have a caret cursor when users mouseover. */
+/* Make sure label names & hints in wdigets have a caret cursor when users mouseover */
```
rfrenchseti: widgets

rfrenchseti (changes requested):
- I don't see the qtype, (i), and trashcan grouped together. The trashcan still moves to the next line by itself. The qtype and (i) are grouped together.
- Also maybe the "+(OR)" should be grouped with the qtype, (i), and trashcan.
- I found one more issue that should be fixed: The rotating chevron (> or v) next to menu category names should stay with the last word so that it never wraps by itself.
- I think this is the correct behavior: "In select metadata menu, when window is narrow, and if a menu item text is wrapped into the 2nd line, the highlighted background will cover the entire row for that menu item. (Match the behavior in search sidebar now)"

rfrenchseti:
- All Django tests pass: Y

PR #1003, opened 2020-02-28
Discard changes can corrupt select metadata

### LINKED ISSUE #914: Download pane loses product types on cart refresh
If you deselect some product types and then do an action that requires a cart refresh, like emptying the recycle bin or restoring the recycle bin, the product types are reloaded with everything selected again. This would be easy to fix by passing a list of DEselected product types into view.html so that they could show up deselected in the first place. That way any NEW product types would show up selected by default.

### LINKED ISSUE #967: Select Metadata "Discard Changes" can corrupt metadata columns
The current code for "Discard Changes" in the "Select Metadata" dialog erases the current columns then adds them back one at a time. The problem is that as they are being added back, the "qualified labels" to add to the right-hand column in the UI are taken from the DOM of the current search menu. But those labels aren't necessarily in the DOM anymore if the categories have changed since they were added. 

For example:
- Select Mission = Cassini
- Go to Select Metadata
- Select a field from Cassini Constraints
- Go back and remove Mission = Cassini so that Cassini Constraints goes away
- Go to Select Metadata
- Select Discard Changes
- Open Select Metadata again
- You will see "undefined" where the Cassini field was

dstopp:
- Fixes #967
  - All Django tests pass: NA

Description of changes:
-Only update opus.prefs.cols on user option for save;
-Save the selected list as array of 'li' html 
-On discard, replace the list of 'selected' li w/the saved/original list of li

Known problems:

### DIFF
--- a/opus/application/static_media/js/selectMetadata.js
+++ b/opus/application/static_media/js/selectMetadata.js
@@ -51,7 +51,7 @@ var o_selectMetadata = {
             // update the data table w/the new columns
             let currentCols = [];
             $("#op-select-metadata .op-selected-metadata-column > ul").find("li").each(function(index, obj) {
-                currentCols.push(obj.id.slice(9));
+                currentCols.push(obj.id.replace("cchoose__", ""));
             });
             if (!o_utils.areObjectsEqual(opus.prefs.cols, currentCols)) {
                 // only pop up the confirm modal if the user clicked the 'X' in the corner
@@ -233,9 +233,7 @@ var o_selectMetadata = {
 
     removeColumn: function(slug) {
         let menuSelector = `#op-select-metadata .op-all-metadata-column a[data-slug=${slug}]`;
-        if ($(`#op-select-metadata .op-all-metadata-column a[data-slug=${slug}]`).length > 0) {
-            o_menu.markMenuItem(menuSelector, "unselected");
-        }
+        o_menu.markMenuItem(menuSelector, "unselected");
 
         $(`#cchoose__${slug}`).fadeOut(200, function() {
             $(this).remove();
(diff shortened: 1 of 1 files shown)


opus/application/static_media/js/selectMetadata.js
```diff
@@ -17,8 +17,11 @@ var o_selectMetadata = {
     // will be used in mutation observer to determine if scrollbar location should
     // be set.
     isSortingHappening: false,
-    currentSelectedMetadata: opus.prefs.cols.slice(),
 
+    // save the original copy in case we need to discard
+    opusPrefsCols: opus.prefs.cols,
```
rfrenchseti: I'm confused by this line. When this is executed, has opus.prefs.cols even been initialized properly yet? The statement will run at script load time. If opus.js was loaded first, then this will have the default columns. But the proper columns won't be set until the hash is parsed later during the initialization, so it seems like this would get the default columns by accident. The variable is set again during render(), so perhaps it would be less confusing just to set it to [] or null here?

rfrenchseti: Also, wouldn't it be better to call it something like "originalOpusPrefsCols" since it isn't always equal to opus.prefs.cols?

opus/application/static_media/js/selectMetadata.js
```diff
@@ -259,30 +261,42 @@ var o_selectMetadata = {
...
+        // add them back in...
+        $(opus.prefs.cols).each(function(index, slug) {
+            let menuSelector = `#op-select-metadata .op-all-metadata-column a[data-slug=${slug}]`;
+            o_menu.markMenuItem(menuSelector);
+        });
+        $(o_selectMetadata.lastSavedSelected).each(function(index, selected) {
+            $("#op-select-metadata .op-selected-metadata-column > ul").append(selected);
+        });
+        $("#op-select-metadata .op-selected-metadata-column").find("li").show();
+        $("#op-select-metadata").modal('hide');
```
rfrenchseti: "" instead of single quotes

rfrenchseti (changes requested):
- Open Select Metadata
- Make a change
- Hit the (X) in the corner and answer "yes" to keeping changes
- Open Select Metadata
- Hit the (X) in the corner and you will be asked again if you want to keep your changes, even though you didn't change anything

rfrenchseti (approved):

PR #1011, opened 2020-03-13
Some small changes requested by Rob.

fyellin:
error-analysis only grabs one date in cronjob mode
Make default timezone into a constant, so that when the log files switch out of UTC, it'll be an easy fix.


fyellin:
Updated pull request to use US/Pacific instead of UTC for determining current time.

PR #1017, opened 2020-03-24
--html,  DEFAULT_FIELDS_PREFIX

fyellin:
Add --html flag to error analyzer.  It doesn't actually generate anything useful yet, but the output is sent to jinja.  Once Rob/Debby decide what the error output looks like, we only need to change that file.

DEFAULT_FIELDS_PREFIX changed to opus.


log_analyzer/templates/error_analysis.html
```diff
@@ -0,0 +1,51 @@
...
+
+<script>
+
+
+</script>
+</body>
+</html>
+
+
+                    for log_entry in log_entries:
```
rfrenchseti: I think these lines shouldn't be here

rfrenchseti (changes requested):

log_analyzer/templates/error_analysis.html
fyellin: Fixed.

This file is just nonsense anyway.  It generates text with some html headers and footers thrown in for fun.  I wouldn't try to actually look at the output in a browser.

rfrenchseti (approved):

PR #1018, opened 2020-03-24
Sort on multiple fields

### LINKED ISSUE #362: Sort on multiple fields
Currently it's possible to sort on a single field by using the "view table" view and clicking on a column header. However, there are times you want to sort on multiple columns, since columns like "planet" or "mission" can have a large number of entries that are all identical. For example, you might want to sort on "Target Body" in ascending order, then on "Filter" in ascending order, then on "Observation Duration" in descending order.

The JPL Imaging Node "Image Atlas" allows this kind of searching. We should too.

dstopp:
- Fixes #362
  - All Django tests pass: NA
  - Tested on Chrome, Firefox, Safari, and iOS: Yes

Description of changes:
-Add ability to sort on multiple fields:
  -add a 'plus' indicator to allow user to add a selected metadata field to sort on;
  -click on table header sort arrow will ask the user if they want to modify or append to current sort order behaviour
  -Limit total number of sortable fields to 9; disable button when limit is reached
-Add ability to reorder the sort fields
  -only rerender if the sort order was actually modified
-Add a small sort order indicator adjacent to the sort arrow that indicates the order in sorting
-Keep the last word of the header + sort arrow + sort number on same line
-Responsive behaviour on small screens
-Apply to both browse and cart pages
-Allow metadata fields in the sort order category that are not currently selected metadata fields to view <cannot add them but they will not disappear from the sort order>

Known problems:
 -As there is ongoing responsive work for mobile devices in the pipeline, this pull request may need to be updated w/those sorts of changes
-on Safari, small screen, lots of metadata fields in 'add' menu, i noticed that the scroll bar doesn't appear unless you try to add.  Not sure how to fix that.


opus/application/apps/ui/templates/ui/base.html
```diff
@@ -114,10 +114,10 @@
                             </ul>
                         </div>
                         <div class="op-sort-order-container">
-                            <span class="op-sort-order-icon" title="Results are sorted by these metadata fields">
+                            <span class="op-sort-order-icon" title="Results are sorted by these metadata fields&#10;Drag to reorder the sort fields">
                                 <i class="fas fa-random"></i>
-                            </span>&nbsp;
-                            <ul class="sort-contents list-inline"></ul>
+                            </span>
+							<div class="op-sort-contents"></div>
```
rfrenchseti: Bad indent

opus/application/apps/ui/templates/ui/base.html
```diff
@@ -183,8 +183,8 @@ <h2>Your search produced no results</h2>
                             <div class="op-sort-order-container">
                                 <span class="op-sort-order-icon" title="Results are sorted by these metadata fields">
                                     <i class="fas fa-random"></i>
-                                </span>&nbsp;
-                                <ul class="sort-contents list-inline"></ul>
+                                </span>
+								<div class="op-sort-contents"></div>
```
rfrenchseti: Bad indent

opus/application/apps/ui/templates/ui/base.html
```diff
@@ -278,6 +278,12 @@ <h2 class="dropdown-header"></h2>
             <a class="dropdown-item" data-action="downloadURL" href="#" download><i class="fas fa-download"></i>Download zipped URL archive&nbsp;</a>
         </div>
 
+        <!-- dropdown menu for sort metadata columns -->
+        <div id="op-add-sort-metadata" class="dropdown-menu dropdown-menu-sm" data-toggle="dropdown" aria-labelledby="op-add-sort-metadata" aria-haspopup="true" aria-expanded="false">
+            <h2 class="dropdown-header">Add metadata field to sort</h2>
+			<div class="op-sort-list"></div>
```
rfrenchseti: Bad indent

opus/application/apps/ui/templates/ui/base.html
```diff
@@ -316,7 +322,7 @@ <h3 class='op-header-text'></h3>
         </div>
 
         <!--modal dialog for the metadata selector -->
-        <div id="op-select-metadata" class="modal fade" role="dialog" tabbindex="-1" data-backdrop="false" aria-labelledby="op-select-metadataLabel" aria-hidden="true">
+        <div id="op-select-metadata" class="modal fade" role="dialog" tabbindex="-1" data-backdrop="false" aria-labelledby="op-select-metadataLabel" ariaaria-hidden="true">
```
rfrenchseti: "ariaaria"

opus/application/static_media/css/opus.css
```diff
@@ -1134,13 +1151,14 @@ a:-webkit-any-link, a::any-link {
...
-    background-color: #303030; /* #383838; */
+.op-sort-add-disabled {
+    cursor: default;
+    opacity: 0.30 !important;
 }
 
-#cart .thumb-page {
-    padding: 0 1.0em;
+#cart .op-cart-gallery-side {
+    /* margin-top:.25em; */
```
rfrenchseti: Remove comment

opus/application/static_media/js/browse.js
```diff
@@ -1308,6 +1156,10 @@ var o_browse = {
 
         let suppressScrollY = false;
 
+        if (opus.prefs.order.length === 0) {
```
rfrenchseti: This does nothing...

opus/application/static_media/js/browse.js
```diff
@@ -1532,14 +1382,39 @@ var o_browse = {
...
+                columnSorting = "desc";
+                icon = "-up";
+                columnOrderPostion = `<span class="${positionIndicatorClasses}">${positionDesc+1}</span>`;
+            }
+
+            let headerArr = header.split(" ");
+            let lastWord = headerArr.pop();
+            let spacing = headerArr.length ? "&nbsp;" : "";
+            let lastWordWrappingGroup = `${headerArr.join(" ")}${spacing}<span class="op-last-word-group">${lastWord}` +
+                            `<span data-sort='${columnSorting}' ${orderToolTip} class='op-column-ordering fas fa-sort${icon}'>${columnOrderPostion}</span>`;
```
rfrenchseti: Indent?

rfrenchseti: Also we normally use double quotes in HTML, and you can do that here since you're using template strings.

opus/application/static_media/js/browse.js
```diff
@@ -1532,14 +1382,39 @@ var o_browse = {
...
+            let columnOrdering = `<a href='' data-slug='${slug}' data-label='${label}'>${lastWordWrappingGroup}</a>`;
 
             $(`${tab} .op-data-table-view thead tr`).append(`<th id='${slug} 'scope='col' class='sticky-header'><div>${columnOrdering}</div></th>`);
```
rfrenchseti: Use double quotes in HTML

opus/application/static_media/js/opus.js
```diff
@@ -859,6 +860,10 @@ var opus = {
                         case "op-reset-opus-modal":
                             location.assign("/opus");
                             break;
+                        case "op-overwrite-sort-order":
+                            let addToSort = false;
+                            o_sortMetadata.onClickSortOrder($(`#${target}`).data("slug"), addToSort);
```
rfrenchseti: Why have a variable addToSort?

opus/application/static_media/js/opus.js
```diff
@@ -891,6 +896,10 @@ var opus = {
                         case "op-close-metadata-modal":
                             o_selectMetadata.discardChanges();
                             break;
+                        case "op-overwrite-sort-order":
+                            let addToSort = true;
+                            o_sortMetadata.onClickSortOrder($(`#${target}`).data("slug"), addToSort);
```
rfrenchseti: Why have a variable addToSort?

opus/application/static_media/js/sortMetadata.js
```diff
@@ -0,0 +1,280 @@
...
+                    o_hash.updateURLFromCurrentHash(); // This makes the changes visible to the user
+                    o_sortMetadata.renderSortedDataFromBeginning();
+                }
+            },
+        });
+
+        // click table column header to reorder by that column
+        // On click, if there are already 2 items in the sort <opusId + slug>, display overwrite modal option
+        // On ctrl click will append the selected sort
+        $("#browse, #cart").on("click", ".op-data-table-view th a",  function(e) {
```
rfrenchseti: Two space before function

opus/application/static_media/js/sortMetadata.js
```diff
@@ -0,0 +1,280 @@
...
+            o_browse.hideMenus();
+            let inOrderList = $(this).find(".op-column-ordering").data("sort") !== "none";
+            let slug = $(this).data("slug");
+            if (inOrderList || opus.prefs.order.length < 2 || (e.ctrlKey || e.metaKey)) {
```
rfrenchseti: I think this is the logic you need it's just <=2

opus/application/static_media/js/sortMetadata.js
```diff
@@ -0,0 +1,280 @@
...
+            $(this).closest(".list-inline-item").remove();
+
+            o_hash.updateURLFromCurrentHash();
+            o_sortMetadata.renderSortedDataFromBeginning();
+        });
+
+        $(".op-sort-contents").on("click", ".op-sort-order-add-icon", function(e) {
+            // allow the hover to work but the click appear to be disabled
+            if ($(".op-sort-order-add-icon").hasClass("op-sort-add-disabled")) {
+
```
rfrenchseti: Extra blank line

opus/application/static_media/js/sortMetadata.js
```diff
@@ -0,0 +1,280 @@
...
+            }
+            return false;
+        });
+    }, // end edit sort metadata behaviours
+
+    onClickSortOrder: function(orderBy, addToSort) {
+        $("body").addClass("op-prevent-pointer-events");
+        o_browse.showPageLoaderSpinner();
+
+        addToSort = (addToSort === undefined ? true : addToSort);
```
rfrenchseti: With ES6 you can use a default argument.... function(orderBy, addToSort=true).

opus/application/static_media/js/sortMetadata.js
```diff
@@ -0,0 +1,280 @@
...
+                fullSlug = "-" + slug;
+            }
+            opus.prefs.order.push(fullSlug);
+            if (tableColumnFields[slug] !== undefined) {
+                delete tableColumnFields[slug];
+            }
+        });
+
+        listHtml += `<div class="op-sort-order-add-icon list-inline-item" title="Edit sort by adding metadata fields">`+
+					   `<i class="fas fa-plus"></i>`+
```
rfrenchseti: Bad indent?

opus/application/static_media/js/sortMetadata.js
```diff
@@ -0,0 +1,280 @@
...
+            if (isDescending) {
+                fullSlug = "-" + slug;
+            }
+            opus.prefs.order.push(fullSlug);
+            if (tableColumnFields[slug] !== undefined) {
+                delete tableColumnFields[slug];
+            }
+        });
+
+        listHtml += `<div class="op-sort-order-add-icon list-inline-item" title="Edit sort by adding metadata fields">`+
```
rfrenchseti: How about just "Add metadata fields to sort order"?

opus/application/static_media/js/sortMetadata.js
```diff
@@ -0,0 +1,280 @@
...
+					   `<i class="fas fa-plus"></i>`+
+					`</div>`;
+
+        $(".op-sort-contents").html(listHtml);
+
+        // if all the metadata field columns are already in the sort list, disable the add button
+        // limit the total number of sort columns to 9
+        if (Object.keys(tableColumnFields).length === 0 || opus.prefs.order.length === 9) {
+            $(".op-sort-order-add-icon").addClass("op-sort-add-disabled");
+            $(".op-sort-order-add-icon").attr("title", "The maximun of 9 metadata fields to sort on has already been selected");
```
rfrenchseti: Change to: "The maximum of nine metadata sort fields has been reached"

rfrenchseti (changes requested):
- Always required OPUS ID to be the last item in the list.
- Put some kind of vertical divider in front of OPUS ID to make it clear there is something different going on.
- Make OPUS ID non-draggable.
- When there are <= 2 sort fields and the user clicks on a table header, do an "overwrite" without asking first. If there are >2 sort fields, do the current behavior.
- When there are no items to put in the dropdown list, the + has the wrong tooltip.
- I think it would be nice to add "Drag to reorder" to the tooltips for each of the individual pills.
- Currently the tooltip for a ^/v in the table header says "Change to ascending(descending) sort". Since you can click anywhere on the header, I think that tooltip should show up no matter where you hover. Also, I think the tooltip should change based on what's actually going to happen. If the field is already in the sort order, then it should say what it currently says. Right now, if the field is not in the sort order, there is no tooltip. It should say "Sort on this field" if there are currently <=2 sort fields or "Sort on this field or append to current sort<br>Ctrl+click to append" if there are >2 sort fields.
- I don't see the "+" being marked as inactive (greyed out) when it isn't going to do anything.
- It is possible for the "+" to wrap around and be on a line by itself. I think it should always stay with the last pill.
- All of our other modal-looking elements have title case in their header. Should we do that for this dropdown too?

opus/application/apps/ui/templates/ui/base.html
rfrenchseti: This is still the wrong indent (although a different one)

opus/application/apps/ui/templates/ui/base.html
rfrenchseti: This is still the wrong indent (although a different one)

opus/application/apps/ui/templates/ui/base.html
rfrenchseti: Still bad indent

opus/application/static_media/js/opus.js
rfrenchseti: This one is still there

opus/application/static_media/js/sortMetadata.js
rfrenchseti: This is still weirdly indented

rfrenchseti (changes requested):
- Still need normalizeurl tests for missing opusid or opusid in the middle of the sort order.
- Think about moving + to the left of opusid and greying out and putting distance before opusid to make it look different than the other pills
- Make sure + always stays with the pill immediately to its left when wrapping

opus/application/static_media/js/sortMetadata.js
dstopp: The indent here is intentional; I am mimicking the html indent in js.

rfrenchseti:
Suggested text for apps/help/templates/help/gettingstarted.html:

```
<p>Results may be sorted in either ascending or descending order by one or more
    metadata fields. The current sort fields and their associated sort
    directions are always displayed at the top after the
    <i class="fas fa-random"></i> icon (the sort order must always include OPUS
    ID as the final field to ensure a deterministic ordering). The sort fields
    may be reordered by dragging the field name and the sort direction for a
    field may be changed by clicking on the field name. Sort fields may be
    added from the currently selected metadata fields by clicking
    <i class="fas fa-plus"></i>; to change the list of available fields use
    Select Metadata. A sort field may be removed by clicking its
    <i class="fas fa-times-circle"></i> icon. In the table view, if a column is
    being used to sort it will display the sort direction
    (<i class="fas fa-sort-up"></i> or <i class="fas fa-sort-down"></i>) and a
    number in a circle showing its position in the sort order. Clicking on the
    header of a column that is used in the current sort will change the
    direction for that field. Clicking on the header of a column that is not
    currently being used will replace the current sort if there is only one
    field there (other than OPUS ID), or ask if you want to overwrite or append
    to the end of the current sort otherwise. Ctrl+clicking on a table header
    will always append that field to the end of the sort (but before OPUS ID).
    The same sort order applies to the <b>Browse</b> tab and the <b>Cart</b>
    tab.</p>
```

rfrenchseti:
- The tooltip for the random icon has "drag to reorder" on a separate line and I think that looks nice. It's also consistent with tooltips on thumbnails. For each sort pill, the "Drag to reorder" (capitalized) should be on a second line with no "-".
- This change has not been done:
Currently the tooltip for a ^/v in the table header says "Change to ascending(descending) sort". Since you can click anywhere on the header, I think that tooltip should show up no matter where you hover. Also, I think the tooltip should change based on what's actually going to happen. If the field is already in the sort order, then it should say what it currently says. Right now, if the field is not in the sort order, there is no tooltip. It should say "Sort on this field" if there are currently <=2 sort fields or "Sort on this field or append to current sort
Ctrl+click to append" if there are >2 sort fields.

rfrenchseti:
- Start with a fresh OPUS. Flip the direction of OPUS ID. Then flip the direction of Start Time. Boom.
- Also dragging pills never does anything. They just snap back to where they started.
- You can drag the last pill (before OPUS ID) to the right of the +.
- Sort by can split across lines. The + doesn't stay with the last pill.
- Sort by also needs to stay with the random icon
- There's still a problem with clicking on the OPUS ID column in the table header.

rfrenchseti (approved):

PR #1048, opened 2020-06-06
Issue 1042 and 1037

### LINKED ISSUE #1037: Using ? as a string search causes an error
Invoke any string search widget. Type "?" and hit enter. You will get an error dialog and OPUS with restart.

### LINKED ISSUE #1042: JavaScript error when entering Cassini orbit range
* Open mission widget and select Cassini
* In Cassini mission mission constraints, open Saturn orbit number (by RANGE) widget
* In the min input, enter some number, like 3, and leave focus or hit enter, error will show up in JS console. There is some issue in URL encoding so the URL won't get updated and no search will be triggered.

juzen2003:
- Fixes #1037, #1042 
  - All Django tests pass: Y
  - JSHINT run on all affected files: Y (widgets.js)

Description of changes:
- Fixed #1037 by encoding the input value before passing into stringsearchchoices API URL.
- Fixed #1042 by making sure when RANGE widgets (without qtype & unit inputs) just get open, the corresponding slugs are updated in opus.selections properly. Previously they are not in opus.selections. 

Known problems:
None

### DIFF
--- a/opus/application/static_media/js/widgets.js
+++ b/opus/application/static_media/js/widgets.js
@@ -1366,14 +1366,18 @@ var o_widgets = {
                 o_search.getHinting(slug);
             }
 
-            // if a RANGE widget that has no qtype & unit inputs just got open, we need to
-            // update opus.selections to make sure input slugs exist.
-            if (widgetInputs.hasClass("RANGE") && !qtypeInputs.length && !unitInput.length) {
+            // if an input widget just got opened and input slugs are not in opus.selections, 
+            // we need to update opus.selections to make sure input slugs exist.
+            if (widgetInputs.hasClass("RANGE")) {
                 if (!opus.selections[`${slug}1`]) {
                     opus.selections[`${slug}1`] = [null];
                 }
                 if (!opus.selections[`${slug}2`]) {
                     opus.selections[`${slug}2`] = [null];
+                }
+            } else if (widgetInputs.hasClass("STRING")) {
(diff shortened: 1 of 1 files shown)


opus/application/static_media/js/widgets.js
```diff
@@ -1365,6 +1365,18 @@ var o_widgets = {
             } else {
                 o_search.getHinting(slug);
             }
+
+            // if a RANGE widget that has no qtype & unit inputs just got open, we need to
```
rfrenchseti: Why is this limited to widgets with no qtype and units? Even if the other cases are taken care of in alignDataInSelectionsAndExtras it's very confusing to have this limitation here with no explanation. Wouldn't it be simpler to just always do it?

rfrenchseti: open => opened

rfrenchseti (changes requested):

opus/application/static_media/js/widgets.js
juzen2003: Just updated it to always check & update opus.selections when input widgets just got opened. Previously I limited to widgets with no qtype & units because opus.selections of other cases will be updated in alignDataInSelectionsAndExtras. But you are right, it's confusing and looks like a special case by only doing this step for widgets with no qtype & units there. Now we do the step (update opus.selections) for all cases of input widgets.

rfrenchseti (approved):

PR #1054, opened 2020-06-13
Fixed the issue that browse view causes API calls past end of observations

### LINKED ISSUE #782: Browse view causes API calls past end of observations
- Go to the table view
- Move the slider to the far right
- Scroll down until you are at the very bottom of the table
- Scroll up one line at a time (by using the arrow keys or mouse wheel)
- Each scroll up will cause an API call beyond the max number of results

The same thing happens with in the gallery view for the law row of thumbnails.

Note the user doesn't really notice anything, but it's annoying to the backend and has extra network delays.

juzen2003:
- Fixes #782
  - All Django tests pass: Y

Description of changes:
- In infiniteScroll path, when the data reaches to the end, disable infiniteScroll load data by setting loadOnScroll to false so it won't keep calling dataimages api with startObs larger the search results.

Known problems:
None

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -556,6 +556,19 @@ var o_browse = {
             "scrollbarOffset": 0
         });
 
+        // Properly set loadOnScroll. This will avoid loadOnScroll being false when the user
+        // drags the slider from the most right end to the middle.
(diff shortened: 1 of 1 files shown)


rfrenchseti (approved):
- All Django tests pass: Y

PR #1083, opened 2020-07-10
417 - Add edit and reorder metadata within galleryView modal

### LINKED ISSUE #417: Allow dragging of table view headers to reorder fields
From Mark Showalter:

You are looking at a table view. Each column has a header. The header is actually a pop-up menu. When you click on the header, you can select any other quantity that we can tabulate. When you release, the content of that column is updated based on the new selection.

In addition there's a widget at right that let's you add another column (or maybe another N columns all at once). They appear, and then you can select each header.

We would also need options to close columns and to sort by columns. I bet they could be integrated into the table headers as additional widgets.

We also need to be able to reorder columns. Maybe that could be done by dragging columns right a left, Otherwise, there could at least be a way to select one or more columns and click a button to move them right or left.

As a user, I find it a bit tedious to go to a separate page or an overlay when I want to choose columns. This UI would make the table view much more interactive and responsive.

dstopp:
- Fixes #417
  - All Django tests pass: Y/N/NA
  - Tested for race conditions (with delayed API calls if applicable): Y/N/NA

Description of changes:
-add an 'edit' button on the galleryView modal to allow user to:
   - add a metadata field from the menu list;
   - remove a metadata field;
   - reorder the metadata fields
-Fixed race condition between search widget and selectMetadata modal that cause unselected fields to be highlighted
-Fixed missing scrollbar <bug> in modal

Known problems:
-There is some code that is unused in this pull that will eventually be part of the table header draggable code.  I did not remove it for this pull.
-There are still 3 django tests failing that appear to not be related to these changes but are still under investigation.


rfrenchseti:
Current issues:

- On top of the field name, the pointer is a "drag hand". But if you keep moving the pointer to the right, it eventually turns into a vertical "text" bar, even though in that region you can still drag the field. It should stay a "drag hand" all the way across.
- When you have a lot of fields and there is a scrollbar in the slideshow view, and then you drag an item, you can't drag it "off the edge" and have it automatically scroll.
- When a field is added at the bottom of the list, and the new field will end up being off the bottom of the slideshow, it only scrolls to show the new field, not the two lines with the new field and its "Add field here" line, so you have to manually scroll a little to be able to add another field at the bottom.
- When the browser is narrow enough that the font size decreases, there is a large margin between the bottom of the field list and the tool icons at the bottom of the slideshow dialog.
- If you add really long field names (like things from Ring Geometry), and then make the browser really narrow, the field names wrap around making the field+data more than two lines. But then when you go into edit mode, it restricts them to be on a single line, moves the big thumbnail left, and shrinks it. We don't do this with long field data that wraps around (like "Note") - that stays multiple line, which means the whole system works fine when there are more than two lines per field.
- Independent of this pull, when the browser is really narrow and there needs to be a horizontal scrollbar, it has an ugly white background.
- There seems to be a noticeable delay between starting a drag motion and the field actually moving. This makes me doubt whether the drag is actually happening. We don't have that delay when dragging search widgets, but we do when dragging "Select Metadata" fields. It seems like a little too long a delay to me, especially in the slide show.
- When you add or remove a field, the main gallery scrollbar does this weird thing where it jumps to the top and then returns back to the middle.
- Moving with the keyboard left/right arrows keeps you in edit mode, but clicking on the left/right arrows in the toolbar exits edit mode.

opus/application/apps/ui/templates/ui/add_field.html
```diff
@@ -0,0 +1,100 @@
+{% block menu %}
+    <ul class="op-search-menu list-group list-group-flush">
+        {% for div in menu.divs %}
+            {% if div.table_name != 'search_fields' %}
+            <li class="list-group-item py-0 op-search-menu-category">
+                        <a href="#{{ which }}-mini-{{ div.table_name }}" data-toggle="collapse" aria-expanded="false" class="dropdown-toggle py-2 op-submenu-category {{ div.collapsed }}" data-cat="{{ div.table_name }}">
+                            <span class="title">{{ div.label }}&nbsp;<span class="op-menu-arrow"></span></span>
+                        </a>
+                <ul class="list-group list-group-flush submenu collapse {{ div.show }}" id="{{ which }}-mini-{{ div.table_name }}">
+
```
juzen2003, line 10: The indentation of ```<a>``` here should be aligned with ```<ul >```.

opus/application/apps/ui/templates/ui/add_field.html
```diff
@@ -0,0 +1,100 @@
...
+                                                        {% else %}
+                                                            {{ p.label }}
+                                                        {% endif %}
+                                                        " data-qualifiedlabel="
+                                                        {% if menu.data.labels_view == 'selector' %}
+                                                            {{ p.body_qualified_label_results }}
+                                                        {% else %}
+                                                            {{ p.body_qualified_label }}
+                                                        {% endif %}
+                                                      " data-slug="{{ p.slug }}" class="op-search-param" href="#">
```
juzen2003: Need 2 more spaces here to aligned with other codes, same at line 72 and 78.

opus/application/static_media/js/browse.js
```diff
@@ -301,6 +304,91 @@ var o_browse = {
...
+            // save anything that was changed via sort/trash before the dropdown is displayed
+            o_browse.onDoneUpdateMetadataDetails();
+            o_browse.showMetadataList(e);
+            $("#op-add-metadata-fields").data("slug", slug);
+
+            return false;
+        });
+
+        $("#galleryView").on("click", "a.op-metadata-detail-remove", function(e) {
+            let slug = $(this).parent().parent().data('slug');
```
juzen2003: Just a note here, we can also use .closest("ul") as the replacement for .parent().parent()

opus/application/static_media/js/browse.js
```diff
@@ -2069,23 +2179,137 @@ var o_browse = {
...
         opus.metadataDetailOpusId = opusId;
+        $(`.op-metadata-details > .loader`).hide();
 
         // list columns + values
-        let html = "<dl>";
+        let html = "";
+        let selectMetadataTitle = "Add metadata field after the current field";
+        let removeTool = `<li class="op-metadata-details-tools list-inline-item">` +
+                         `<a href="#" class="op-metadata-detail-remove" title="Remove selected metadata field"><i class="far fa-trash-alt"></i></a></li>`;
+        //let addTool = `<a href="#" class="op-metadata-details-tools op-metadata-detail-add dropdown" title="${selectMetadataTitle}" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" "><i class="fas fa-plus pr-1"> Add field here</i></a>`;
```
juzen2003: Please remove the commented out variable (not used).

opus/application/static_media/js/browse.js
```diff
@@ -301,6 +304,91 @@ var o_browse = {
...
+            let parentSlug = $("#op-add-metadata-fields").data("slug");
+            let index = $.inArray(parentSlug, opus.prefs.cols);
+            if (index >= 0) {
+                index++;
+                opus.prefs.cols.splice(index, 0, slug);
+                // save the last slug that was added so that we can make sure it is visible on scroll
+                $("#op-add-metadata-fields").data("last", slug);
+                $(`.op-metadata-details > .loader`).show();
+                $((`#op-add-metadata-fields .op-select-list a[data-slug="${slug}"]`)).hide();
+                //o_menu.markMenuItem(`#op-add-metadata-fields .op-select-list a[data-slug="${slug}"]`);
```
juzen2003: Please remove unused function call.

opus/application/static_media/js/browse.js
```diff
@@ -301,6 +304,91 @@ var o_browse = {
...
+                o_browse.initEditMetadataDetails();
+            } else {
+                o_browse.removeEditMetadataDetails();
+            }
+            return false;
+        });
+
+        $(".op-select-list").on("click", ".dropdown-item", function(e) {
+            o_browse.hideMenus();
+            //o_sortMetadata.onClickSortOrder($(this).data("slug"));
```
juzen2003: Please remove unused function call.

juzen2003 (changes requested):
- All Django tests pass: Y

- Issues:
    - In metadata detail edit menu, when dragging around an item, release and quickly drag any item again, there will be some console errors. Here are steps to reproduce:
        1. In edit detail menu, drag an item, and move around.
        2. Release the item at any position (by releasing the mouse), and quickly hit the mouse again to drag any item.
        3. Move around for some time and the console will have the following errors from jquery ui:
            ```
            Uncaught TypeError: Cannot read property 'insertBefore' of null
            Uncaught TypeError: Cannot read property 'removeChild' of null
            ```
        (Issue can be reproduced by repeating the above steps)
    - In metadata detail slide show, when screen is small, metadata detail info will overlap a little bit with the bottom row (the row of cart, prev/next, and hamburger icons).

opus/application/static_media/js/browse.js
dstopp: done

opus/application/static_media/js/browse.js
dstopp: done

opus/application/static_media/js/browse.js
dstopp: done

opus/application/static_media/js/browse.js
dstopp: thanks, good suggestion.  done

opus/application/apps/ui/templates/ui/add_field.html
dstopp: done

opus/application/apps/ui/templates/ui/add_field.html
dstopp: done

dstopp:
still outstanding:

- In metadata detail slide show, when screen is small, metadata detail info will overlap a little bit with the bottom row (the row of cart, prev/next, and hamburger icons).

juzen2003:
I create a pull request to Debby's repo (https://github.com/dstopp/pds-opus/pull/9) with these fixes:
- Dynamically update the height of "#galleryViewContents .op-metadata-details" to make sure items in the right columns of metadata detail modal will not overlap with the bottom icon row.
- Change ".op-metatdata-detail-edit" to ".op-metadata-detail-edit".
- Add styling to make sure info with long string like "Multiple Target List" is wrapped in metadata detail modal.

The remaining issue is this one mentioned in previous comment:
- In metadata detail edit menu, when dragging around an item, release and quickly drag any item again, there will be some console errors. Here are steps to reproduce:
        1. In edit detail menu, drag an item, and move around.
        2. Release the item at any position (by releasing the mouse), and quickly hit the mouse again to drag any item.
        3. Move around for some time and the console will have the following errors from jquery ui:
```
Uncaught TypeError: Cannot read property 'insertBefore' of null
Uncaught TypeError: Cannot read property 'removeChild' of null
```

opus/application/static_media/js/utils.js
```diff
@@ -10,6 +10,8 @@
 var o_utils = {
 /* jshint varstmt: true */
 
+    numberOfDisableUserInteractionRequests: 0,
```
rfrenchseti: This variable is never used anywhere.

opus/application/static_media/js/browse.js
```diff
@@ -1235,6 +1326,10 @@ var o_browse = {
         let selector = `${tab} ${contentsView}`;
         let infiniteScrollDataObj = $(selector).data("infiniteScroll").options;
 
+        let hashArray = o_hash.getHashArray();
+        let slugs = hashArray.cols.split(",");
+
+
```
rfrenchseti: Do you want two blank lines here?

rfrenchseti (changes requested):
- If you resize the browser while the dropdown is open, you can easily make it run off the top or bottom of the browser. I think it would make more sense if on resize we just closed the dropdown.
- As soon as you start dragging a field, a "hole" opens up below the field and all the lower fields move down one slot. This is not how dragging anywhere else works.
- Because of the above, you can end up with a weird visual like shown below (note duplicated field name).

Screenshot from 2020-07-21 17-33-16

- Once all of the fields in a category are selected, the category name is still present. Wouldn't it be better to remove the category name from the list entirely? The way it is right now you just keep clicking on the category name expecting something to happen, and it never does. Don't forget to handle the case where there are no fields left to add.
- I think there needs to be a little more vertical space between "Edit" and the first metadata line. Perhaps you could just move "Edit" up slightly and leave everything else where it is.
- If you have a very wide field contents, the default browser X scrollbar will show up. I thought Dave had fixed this - he and I talked about it yesterday.

Screenshot from 2020-07-21 17-39-04

- I thought you were going to disable interaction while the new data was loading after adding/dragging/removing a field? Right now there's a race condition where if you add a field, and then add a new field while the previous adding is processing in the backend, only the final field will get added and the earlier ones will be thrown away. It's easy to reproduce with backend API delays.

juzen2003:
- One weird issue found in metadata detail slideshow:
1. In edit detail menu, drag an item, and move around.
2. Release the item at any position (by releasing the mouse), and stay in edit detail menu.
3. Click either prev/next, there will be a server error.
Screen Shot 2020-07-21 at 5 49 03 PM

In step 2, if we go back to normal metadata detail menu by clicking "Done", and then click prev/next icons, we will not see the server error.

rfrenchseti:
Another bug...If you add a field using the Select Metadata dialog, and then pull up "Add field" in the slideshow, the field you added is still in the list. The dropdown hasn't been refreshed.

rfrenchseti:
**New issue:**

- The dropdown box for both "add field" and the sort order pills can be "selected" with Ctrl+A. They should be marked as not-selectable.

**Previous issues:**

- "As soon as you start dragging a field, a "hole" opens up below the field and all the lower fields move down one slot. This is not how dragging anywhere else works."

The huge hole is gone, but now there's a smaller hole. As soon as you start dragging, all of the fields below the current one shift down about 1/2 a field height.

- "If you have a very wide field contents, the default browser X scrollbar will show up. I thought Dave had fixed this - he and I talked about it yesterday."

This still happens, plus it does something weird with the field height

Screenshot from 2020-07-22 11-07-58

- "Another bug...If you add a field using the Select Metadata dialog, and then pull up "Add field" in the slideshow, the field you added is still in the list. The dropdown hasn't been refreshed."

This still happens.

- From Dave: "One weird issue found in metadata detail slideshow:
In edit detail menu, drag an item, and move around.
Release the item at any position (by releasing the mouse), and stay in edit detail menu.
Click either prev/next, there will be a server error."

This still happens.

opus/application/static_media/js/utils.js
dstopp: done

opus/application/static_media/js/browse.js
dstopp: done

juzen2003:
- When all fields are added in the in metadata detail slideshow, and the user clicks the "+ Add field here", the dropdown will be empty with the last category left inside. Do we want to disable the dropdown or have a full blank dropdown in this case?
Here is the screen capture of the current behavior:
1. Add all fields in metadata detail slideshow.
2. Click any "+ Add field here" to open the dropdown
Screen Shot 2020-07-24 at 10 53 01 PM
3. In this case, fields in "Ring Geometry Constraints" category are added at the last, so the dropdown is open with that category title displayed (with empty contents).

rfrenchseti:
- I think the right behavior is to disable all the Add Field buttons, just like we disable the “+” in the sort pills.
- I can still reproduce the problem with moving fields quickly, hitting next, and having it throw a server backend error. It's hard to do, but I've done it twice.
- The subheadings under Ring Geometry should be indented
- If all the fields under Ring Geometry get used, all the subheadings go away, but Ring Geometry itself still stays around.

rfrenchseti (approved):

Issue #1084, opened 2020-07-10
Look into upgrading to Django 4.x and Python 3.10


rfrenchseti:
https://docs.djangoproject.com/en/3.0/releases/3.0/

juzen2003:
- Known problems of upgrading to either Django 3.1.2 or 3.0.10
  - Version 3.1.2: (Need to wait for update on sql_flush call in django_nose for python manage.py test to work)
    - python manage.py test can't be run. We need to wait for the update on sql_flush call in django_nose.
    - Reference: wemake-services/django-test-migrations#117
    - Workaround, modify line 405 in runner.py (under site-pacakges/django_nose) by changing from:
```flush_statements = connection.ops.sql_flush(style, tables, connection.introspection.sequence_list())```
to:
```flush_statements = connection.ops.sql_flush(style, tables)```
  - Version 3.0.10: (Need to add a workaround to remove session from cache in base.py to make it work)
    - To have python manage.py runserver to work, we need to add a workaround to remove session from cahce under python site-packages "django/contrib/sessions/backends/base.py", we don't need this workaround in 3.1.x. After workaround, runserver will work, and test will pass.
    - Reference: https://code.djangoproject.com/ticket/31592
    - Workaround, add these in base.py:
       ```
       from django.contrib.sessions.models import Session
       Session.objects.all().delete()
       ```
- Requirement for Django version 3.x:
  - Change "render_to_response" to "render"
    - Reference: https://stackoverflow.com/questions/55911435/django-importerror-cannot-import-name-render-to-response-from-django-shortcu
  - Dependencies:
    - djangorestframework==3.11.1
    - mysqlclient==2.0.1

rfrenchseti:
Dave says:

I think it's ok for us to stay on Django 2.2 since it's still releasing new update (the latest 2.2.17 is expected to release on Nov). And at the same time we can wait for django_nose to release its new support on 3.1.x for the sql_flush, once it supports 3.1.x, we can upgrade to Django 3.1.x right away.

Note: To make django_nose sql_flush call work with Django 3.1.x, we can modify line 405 in runner.py (under site-pacakges/django_nose) by changing from:
```flush_statements = connection.ops.sql_flush(style, tables, connection.introspection.sequence_list())```
to:
```flush_statements = connection.ops.sql_flush(style, tables)```
With above modification, python manage.py test will run and pass, runserver will work fine too.

rfrenchseti:
Deferring until django_nose is updated.

[...]

PR #1104, opened 2020-10-14
Dummy pull with notes on upgrading to Django 3.x

juzen2003:
- Addresses: #1084 
- Known problems of upgrading to either Django 3.1.2 or 3.0.10
    - Version 3.1.2: (Need to wait for update on sql_flush call in django_nose for python manage.py test to work)
        - python manage.py test can't be run. We need to wait for the update on sql_flush call in django_nose.
        - Reference: https://github.com/wemake-services/django-test-migrations/issues/117
    - Version 3.0.10: (Need to add a workaround to remove session from cache in base.py to make it work)
        - To have python manage.py runserver to work, we need to add a workaround to remove session from cahce under python site-packages "django/contrib/sessions/backends/base.py", we don't need this workaround in 3.1.x. After workaround runserver will work, and test will pass.
        - Reference: https://code.djangoproject.com/ticket/31592

- Requirement for Django version 3.x:
    - Change "render_to_response" to "render"
    - Reference: https://stackoverflow.com/questions/55911435/django-importerror-cannot-import-name-render-to-response-from-django-shortcu

- Note: mysqlclient==2.0.1 is required for runserver to work on mac.

### DIFF
--- a/opus/application/apps/metadata/views.py
+++ b/opus/application/apps/metadata/views.py
@@ -30,7 +30,7 @@ from django.core.cache import cache
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import Max, Min, Count
 from django.http import Http404, HttpResponseServerError
-from django.shortcuts import render_to_response
+from django.shortcuts import render
 from django.views.decorators.cache import never_cache
 
 from cart.models import Cart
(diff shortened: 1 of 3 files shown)


rfrenchseti:
So based on your analysis and reading up on what they've changed, do you think it's worth upgrading to 3.x or is it OK to stay on 2 for now?

juzen2003:
I think it's ok for us to stay on Django 2.2 since it's still releasing new update (the latest 2.2.17 is expected to release on Nov). And at the same time we can wait for django_nose to release its new support on 3.1.x for the sql_flush, once it supports 3.1.x, we can upgrade to Django 3.1.x right away.

Note: To make django_nose sql_flush call work with Django 3.1.x, we can modify line 405 in runner.py (under site-pacakges/django_nose) by changing from:
```flush_statements = connection.ops.sql_flush(style, tables, connection.introspection.sequence_list())```
to:
```flush_statements = connection.ops.sql_flush(style, tables)```
With above modification, python manage.py test will run and pass, runserver will work fine too.

PR #1105, opened 2020-10-16
-implement the up/down keys for the gallery view/slide modal

### LINKED ISSUE #715: Use up/down arrows with slideshow view to move binoculars
B043: It would be nice to be able to use the up and down arrows to move the binoculars in a 2-D grid.

dstopp:
- Fixes #715
  - If YES:
    - Database used: opus3_test_200903
    - All Django tests pass: NA
    - FLAKE8 run on modified code: NA
  - If YES:

Description of changes:
-Catch the up and down arrows when the modal is open; in gallery view, the up/down will move the binoculars up or down one row and update the modal w/the new opusID; in table view, the up/down will do the same as next/prev and move one up or one down, then update the modal.
-if there are no more rows up or down, the up/down arrow will do nothing.

Known problems:
 - there is a bug in the table view that causes the highlighted row to always be at the bottom of the screen once it reaches a certain part of the visible screen.
 - there is a bug for a corner case when the binoculars or highlighted row are scrolled up off the screen, edit of the metadata fields will cause the metadata on the slide view modal to disappear

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -2027,7 +2027,6 @@ var o_browse = {
             o_browse.renderGalleryAndTable(data, this.url, view);
 
             if (opus.metadataDetailOpusId !== "") {
-                let obsNum =  o_browse.getGalleryElement(opus.metadataDetailOpusId).data("obs");
                 o_browse.metadataboxHtml(opus.metadataDetailOpusId, view);
             }
             o_sortMetadata.updateSortOrder(data);
@@ -2568,7 +2567,7 @@ var o_browse = {
             let buttonInfo = o_browse.cartButtonInfo(action);
 
             // prev/next buttons - put this in galleryView html...
-            html = `<div class="col"><a href="#" class="op-cart-toggle" data-id="${opusId}" title="${buttonInfo[tab].title} (spacebar)"><i class="${buttonInfo[tab].icon} fa-2x float-left"></i></a></div>`;
+            let html = `<div class="col"><a href="#" class="op-cart-toggle" data-id="${opusId}" title="${buttonInfo[tab].title} (spacebar)"><i class="${buttonInfo[tab].icon} fa-2x float-left"></i></a></div>`;
             html += `<div class="col text-center op-obs-direction">`;
             let opPrevDisabled = (nextPrevHandles.prev == "" ? "op-button-disabled" : "");
             let opNextDisabled = (nextPrevHandles.next == "" ? "op-button-disabled" : "");
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -553,7 +553,6 @@ var o_browse = {
                         }
                         break;
                     case 38:  // up
-                        // note that if there is no where to go in the up direction, nothing will change ... it could go to the very first, tho
                         offset = (o_browse.isGalleryView() ? viewNamespace.galleryBoundingRect.x : 1);
                         if (obsNum > offset) {
                             obsNum -= offset;
@@ -567,7 +566,6 @@ var o_browse = {
                         }
                         return false;
                     case 40:  // down
-                        // note that if there are not enough left to go down, nothing will change ... it could go to the very last tho
                         offset = (o_browse.isGalleryView() ? viewNamespace.galleryBoundingRect.x : 1);
                         if (obsNum + offset <= viewNamespace.totalObsCount) {
                             obsNum += offset;
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -514,72 +514,50 @@ var o_browse = {
                 o_browse.undoRangeSelect();
             }
             if ($("#galleryView").hasClass("show")) {
-                /*  Catch the right/left arrow and spacebar while in the modal
-                    Up: 38
-                    Down: 40
-                    Right: 39
-                    Left: 37
-                    Space: 32 */
-
-                let tab = opus.getViewTab();
-                let viewNamespace = opus.getViewNamespace();
-                let detailOpusId;
-                let offset = 0;
-                let obsNum = $("#galleryViewContents .op-obs-direction a").data("obs");
-                let lastCachedObsNum = $(`${tab} .op-thumbnail-container`).last().data("obs");
-                // the || is for cross-browser support; firefox does not support keyCode
-                switch (e.which || e.keyCode) {
-                    case 32:  // spacebar
-                        if (opus.metadataDetailOpusId !== "" && opus.metadataDetailOpusId !== undefined && !$("#galleryViewContents").hasClass("op-disabled")) {
-                            o_browse.undoRangeSelect();
-                            o_cart.toggleInCart(opus.metadataDetailOpusId);
-                        }
-                        break;
-                    case 39:  // next
-                        obsNum++;
-                        // if this new observation is in the last row of cached observations, don't allow the user to arrow there
-                        if (lastCachedObsNum === viewNamespace.totalObsCount || obsNum + viewNamespace.galleryBoundingRect.x <= lastCachedObsNum) {
-                            detailOpusId = $("#galleryView").find(".op-next").data("id");
-                            o_browse.removeEditMetadataDetails();
-                            o_browse.loadPageIfNeeded("next", detailOpusId);
-                        } else {
-                            return false;
-                        }
-                        break;
-                    case 37:  // prev
-                        obsNum--;
-                        detailOpusId = $("#galleryView").find(".op-prev").data("id");
-                        o_browse.removeEditMetadataDetails();
-                        o_browse.loadPageIfNeeded("prev", detailOpusId);
-                        break;
-                    case 38:  // up
-                        offset = (o_browse.isGalleryView() ? viewNamespace.galleryBoundingRect.x : 1);
-                        if (obsNum > offset) {
-                            obsNum -= offset;
-                            detailOpusId = (o_browse.isGalleryView() ? $(tab).find(`.op-thumbnail-container[data-obs='${obsNum}']`).data("id") : $(tab).find(`tr[data-obs='${obsNum}']`).data("id"));
-                            o_browse.removeEditMetadataDetails();
-                            o_browse.loadPageIfNeeded("prev", detailOpusId);
-                        }
-                        break;
-                    case 40:  // down
-                        offset = (o_browse.isGalleryView() ? viewNamespace.galleryBoundingRect.x : 1);
-                        if (obsNum + offset <= viewNamespace.totalObsCount) {
-                            obsNum += offset;
-                            // if this new observation is in the last row of cached observations, don't allow the user to arrow there
-                            if (lastCachedObsNum === viewNamespace.totalObsCount || obsNum + viewNamespace.galleryBoundingRect.x <= lastCachedObsNum) {
-                                detailOpusId = (o_browse.isGalleryView() ? $(tab).find(`.op-thumbnail-container[data-obs='${obsNum}']`).data("id") : $(tab).find(`tr[data-obs='${obsNum}']`).data("id"));
-                                o_browse.removeEditMetadataDetails();
-                                o_browse.loadPageIfNeeded("next", detailOpusId);
-                            } else {
-                                return false;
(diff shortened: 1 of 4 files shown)


opus/application/static_media/js/browse.js
```diff
@@ -559,23 +536,69 @@ var o_browse = {
...
                         o_browse.loadPageIfNeeded("next", detailOpusId);
                         break;
                     case 37:  // prev
                         detailOpusId = $("#galleryView").find(".op-prev").data("id");
+                        obsNum--;
                         o_browse.removeEditMetadataDetails();
                         o_browse.loadPageIfNeeded("prev", detailOpusId);
                         break;
+                    case 38:  // up
+                        // note that if there is no where to go in the up direction, nothing will change ... it could go to the very first, tho
```
rfrenchseti: I don't understand this comment. What does "it could go to the very first" mean? Also "nowhere" is one word.

rfrenchseti (changes requested):

opus/application/static_media/js/browse.js
```diff
@@ -548,34 +521,102 @@ var o_browse = {
...
+                let tab = opus.getViewTab();
+                let viewNamespace = opus.getViewNamespace();
                 let detailOpusId;
+                let offset = 0;
+                let obsNum = $("#galleryViewContents .op-obs-direction a").data("obs");
                 // the || is for cross-browser support; firefox does not support keyCode
                 switch (e.which || e.keyCode) {
                     case 32:  // spacebar
-                        if (opus.metadataDetailOpusId !== "") {
+                        if ((opus.metadataDetailOpusId !== "" && opus.metadataDetailOpusId !== undefined) && !$("#galleryViewContents").hasClass("op-disabled")) {
```
rfrenchseti: No reason to have the first set of nested parens. (A & B) & C is the same as A & B & C.

opus/application/static_media/js/browse.js
```diff
@@ -548,34 +521,102 @@ var o_browse = {
...
                             o_cart.toggleInCart(opus.metadataDetailOpusId);
                         }
                         break;
                     case 39:  // next
-                        detailOpusId = $("#galleryView").find(".op-next").data("id");
-                        o_browse.removeEditMetadataDetails();
-                        o_browse.loadPageIfNeeded("next", detailOpusId);
+                        obsNum++;
+                        // if this new observation is in the last row of screen, don't allow the user to arrow there
+                        if (!o_browse.isLastRowOfObservations(tab, viewNamespace, obsNum)) {
```
rfrenchseti: This should only matter when loading is in progress. The way it's written right now, I can never "arrow" to the last row of observations at the end of the result set.

opus/application/static_media/js/browse.js
```diff
@@ -548,34 +521,102 @@ var o_browse = {
...
+                        if (obsNum > offset) {
+                            obsNum -= offset;
+                            detailOpusId = (o_browse.isGalleryView() ? $(tab).find(`.op-thumbnail-container[data-obs='${obsNum}']`).data("id") : $(tab).find(`tr[data-obs='${obsNum}']`).data("id"));
+                            o_browse.removeEditMetadataDetails();
+                            o_browse.loadPageIfNeeded("prev", detailOpusId);
+                        }
+                        break;
+                    case 40:  // down
+                        offset = (o_browse.isGalleryView() ? viewNamespace.galleryBoundingRect.x : 1);
+                        if (obsNum + offset <= viewNamespace.totalObsCount) {
```
rfrenchseti: Same here - this should only be checked when loading is in progress.

opus/application/static_media/js/browse.js
```diff
@@ -548,34 +521,102 @@ var o_browse = {
...
-                    o_browse.updateGalleryView(detailOpusId);
+                    o_browse.updateGalleryView(detailOpusId, obsNum);
+                    return false;
                 }
             }
             // don't return false here or it will snatch all the user input!
         });
     }, // end browse behaviors
 
+    isLastRowOfObservations: function(tab, viewNamespace, obsNum) {
```
rfrenchseti: I'm confused by this function. The point is not to check for the last row on the current screen. The point is to check for the last row in the cached data regardless of whether it's displayed or not. This appears to be doing the former.

opus/application/static_media/js/browse.js
```diff
@@ -548,34 +521,102 @@ var o_browse = {
...
+        let firstObsOnScreen = $(`${tab} .op-observation-slider`).slider("option", "value");
+        let obsPerRow = viewNamespace.galleryBoundingRect.x;
+        let totalObsOnScreen = obsPerRow * viewNamespace.galleryBoundingRect.yFloor;
+        let beginningOfLastRow = firstObsOnScreen + (totalObsOnScreen - obsPerRow);
+        let lastObs = $(`${tab} .op-thumbnail-container`).last().data("obs");
+        return (lastObs - firstObsOnScreen < o_browse.getLimit() && obsNum >= beginningOfLastRow);
+    },
+
+    onGalleryOrRowClick: function(obj, e) {
+        // make sure selected modal thumb is unhighlighted, as clicking on this closes the modal
```
rfrenchseti: I really hate the use of the word "modal" everywhere. There are dozens of "modals" in OPUS. We should always specify which one we're talking about. Do you mean the "slide show modal"? (I know this was an existing comment that got moved, but still we should fix it)

rfrenchseti (changes requested):
I would still like the slide show dialog not to recenter when it's open and you click on a new thumbnail. With the full arrows available, moving the dialog out of the way will be more common, and this will show up more.

opus/application/static_media/js/browse.js
```diff
@@ -548,34 +521,94 @@ var o_browse = {
...
                             o_browse.undoRangeSelect();
                             o_cart.toggleInCart(opus.metadataDetailOpusId);
                         }
                         break;
                     case 39:  // next
-                        detailOpusId = $("#galleryView").find(".op-next").data("id");
-                        o_browse.removeEditMetadataDetails();
-                        o_browse.loadPageIfNeeded("next", detailOpusId);
+                        obsNum++;
+                        // if this new observation is in the last row of cached observations, don't allow the user to arrow there
```
rfrenchseti: Isn't this comment wrong? I thought we only disallowed the last row if a load was in progress. In fact I don't understand this at all - I can move the binocs to the last row, but this code would seem to prevent that. What am I missing?

opus/application/static_media/js/browse.js
```diff
@@ -548,34 +521,94 @@ var o_browse = {
...
+                        if (lastCachedObsNum === viewNamespace.totalObsCount || obsNum + viewNamespace.galleryBoundingRect.x <= lastCachedObsNum) {
+                            detailOpusId = $("#galleryView").find(".op-next").data("id");
+                            o_browse.removeEditMetadataDetails();
+                            o_browse.loadPageIfNeeded("next", detailOpusId);
+                        } else {
+                            return false;
+                        }
                         break;
                     case 37:  // prev
+                        obsNum--;
```
rfrenchseti: Why don't we check for obsNum < 1 here and return false like everywhere else in this switch?

opus/application/static_media/js/browse.js
```diff
@@ -548,34 +521,94 @@ var o_browse = {
...
                         o_browse.loadPageIfNeeded("prev", detailOpusId);
                         break;
+                    case 38:  // up
+                        offset = (o_browse.isGalleryView() ? viewNamespace.galleryBoundingRect.x : 1);
+                        if (obsNum > offset) {
+                            obsNum -= offset;
+                            detailOpusId = (o_browse.isGalleryView() ? $(tab).find(`.op-thumbnail-container[data-obs='${obsNum}']`).data("id") : $(tab).find(`tr[data-obs='${obsNum}']`).data("id"));
+                            o_browse.removeEditMetadataDetails();
+                            o_browse.loadPageIfNeeded("prev", detailOpusId);
+                        }
```
rfrenchseti: Some other cases return false when we don't do anything. Why not here?

opus/application/static_media/js/browse.js
```diff
@@ -548,34 +521,94 @@ var o_browse = {
...
+                        break;
+                    case 40:  // down
+                        offset = (o_browse.isGalleryView() ? viewNamespace.galleryBoundingRect.x : 1);
+                        if (obsNum + offset <= viewNamespace.totalObsCount) {
+                            obsNum += offset;
+                            // if this new observation is in the last row of cached observations, don't allow the user to arrow there
```
rfrenchseti: Same comment as above about this comment and only doing on load in progress.

opus/application/static_media/js/browse.js
rfrenchseti: Also, if you go to the table view, make enough columns that there's a horizontal scrollbar, go to the very last observation, and hit the right arrow, the scrollbar will move rather than just ignoring the arrow like we should because it's now captured by the slide.

rfrenchseti (changes requested):
1) There's an existing bug that has gotten worse with these changes:
- Click on a thumbnail
- Enter "Edit" mode on the slide
- Add or remove a field
- While the new metadata is being retrieved, click on a different background thumbnail
- The binoculars will disappear completely (up until now, same behavior as production)
- On the production version, you can use the left/right arrows to get the binocs back
- In the new version, the left arrow works, but right/up/down don't do anything

2) The up/down keys do the same thing (move the binocs) whether the mouse is on top of the slide dialog or the background thumbnails. However, the spacebar (toggle cart) both toggles the cart AND scrolls the browser window if the mouse is over the thumbnails. The default behavior of spacebar needs to be disabled with the slide is up.

opus/application/static_media/js/browse.js
dstopp: Even though the check is unnecessary as there are other checks in place for this, i have added a check to the code for consistency.

opus/application/static_media/js/browse.js
dstopp: It's superfluous - there's no change in behaviour if there is a return false here or not, it only matters at the end of the cache where the right arrow or down can start scrolling in some direction that needs to be controlled.

opus/application/static_media/js/browse.js
dstopp: comment clarified in both places.

juzen2003:
- There are two issues:
1. Binoculars stay on the gallery view after some changes in select metadata menu.
    - Steps to reproduce the issue:
        1. Start OPUS freshly, and go to gallery view.
        2. Click on any thumbnail to open the metadata modal, and then close it.
        3. Go to select metadata menu and add something like volume id, and then save.
        4. Binoculars will stick to the thumbnail we clicked at step b with metadata modal closed. Hitting left/right arrows won't do anything and up/down arrows will move the scrollbar without moving the binoculars. Need to click on any thumbnail again to open the metadata modal and resume the correct behavior of binoculars.
Screen Shot 2020-11-04 at 8 22 15 AM

2. In console, error message: metadataboxHtml: in each, observationData may be out of sync with colLabels;
    - Steps to reproduce the issue:
        1. Add some API delay.
        2. In gallery view, open up a metadata detail modal, and hold the down arrow key and it will move down to trigger initInfiniteScroll load.
        3. When spinner is still running, click edit in the modal and delete a metadata col. At the same time, keeping hitting down key maybe for about 3 to 4 times, then we can see the error message in the console. Most of the time, all metadata info will still be loaded fine, but sometimes it will have a blank right pane without any metadata info.
Screen Shot 2020-11-04 at 7 06 24 AM
Screen Shot 2020-11-04 at 7 59 28 AM

dstopp:
re: 1.	Binoculars stay on the gallery view after some changes in select metadata menu - that has been resolved.

re: 2.	In console, error message: metadataboxHtml: in each, observationData may be out of sync with colLabels;   This issue is in production code and is an unrelated bug.  It is easy to dup in production code w/out any delays by using the table view, edit metadata, etc, similar as described above.

juzen2003 (changes requested):
- Database used: opus3_test_200903
- All Django tests pass: Y
    - In browse.js
        - undefined variable ```html``` in ```updateGalleryView```
        - unused variable ```obsNum``` at line 2030

- Only minor changes needed for jshint test on browse.js

rfrenchseti (approved):

PR #1111, opened 2020-12-02
Make sure highlighted table item is full displayed and scrollbar is properly scrolled when hitting up/down arrows

### LINKED ISSUE #1107: In the table view, if the slide view modal highlighted is close to the bottom, left arrow will not move up in the table list.
To reproduce:
Move the slider to the middle of the set of observations and set the view to table view.  Select the last observation in the table list to show a slide view modal; then left arrow to the previous observation.
Notice that the highlighted observation will stay at the bottom of the table instead of just moving up as expected.

This appears to be a bug w/the variable "bottom" in the function $.fn.isOnScreen = function(scope, slop).

juzen2003:
- Fixes #1107 
  - If YES:
    - Database used: opus3_test_200903
    - All Django tests pass: Y
  - If YES:

Description of changes:
- Update the offset and footer top position for isOnScreen to make sure highlighted table item is fully displayed.
- Update the calculation of tr count to get the correct number of table rows on the screen. This will make sure scrollbar is scrolled down by one table row height when we keep hitting down arrow to highlight the next hidden row. Without this change, scrollbar could scroll down by two table row heights in Firefox & Safari.

Known problems:
NA

### DIFF
--- a/opus/application/static_media/js/browse.js
+++ b/opus/application/static_media/js/browse.js
@@ -2133,7 +2133,7 @@ var o_browse = {
             let tableRowHeight = ($(`${tab} .op-data-table tbody tr[data-obs]`).length === 1 ?
                                   $(`${tab} .op-data-table tbody tr[data-obs]`).outerHeight() :
                                   $(`${tab} .op-data-table tbody tr[data-obs]`).eq(1).outerHeight());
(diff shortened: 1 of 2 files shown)


dstopp (approved):
tested and works

PR #1125, opened 2021-03-02
Add ability for placing a notification/message into OPUS

### LINKED ISSUE #1115: Add mechanism for short-term notifications
OPUS needs a mechanism to display a banner of some kind when people start OPUS, for announcements such as upcoming service interruptions. This would be done by reading a local file on the server, and, if it exists, displaying it in a pretty manner.

dstopp:
- Fixes #1115
  - If YES:
    - Database used: XXX
    - All Django tests pass: Y/NA
    - FLAKE8 run on modified code: Y/NA
  - If YES:

Description of changes:

- on open or tab change, check the file notification.html to see if it is empty or has changed; display contents in a modal if so. Contents should be regular html; no error checking is done on contents of file.
- User can choose to not see the modal again, which will store a cookie that contains the modification date of the file. When the file is modified, the contents will be displayed again until the user requests to not see it.
- the secrets file has been modified to point to the location of the notification.html file

Known problems:
 - No error checking on the contents of the file notifications to verify that it contains valid html.


opus/application/apps/ui/test_ui.py
```diff
@@ -31,38 +31,47 @@ def tearDown(self):
 
 
             ###################################################
```
rfrenchseti: Fix #### above and below to have the correct length

opus/application/apps/ui/test_ui.py
```diff
@@ -31,38 +31,47 @@ def tearDown(self):
...
+    def test__api_notifications_ok(self):
+        "[test_ui.py] api_notifications: normal"
         settings.OPUS_LAST_BLOG_UPDATE_FILE = 'test_api/data/lastblogupdate.txt'
-        request = self.factory.get('__lastblogupdate.json')
-        ret = api_last_blog_update(request)
+        request = self.factory.get('__notifications.json')
+        ret = api_notifications(request)
         print(ret)
-        self.assertEqual(ret.content, b'{"lastupdate": "2019-JAN-01"}')
+        self.assertEqual(ret.content, b'{"lastupdate": "2019-JAN-01", "notification": null, "notification_mdate": null}')
```
rfrenchseti, line 59: Wouldn't it be better to update the test file with a notification, or have two test files one with and one without? The mdate is a problem, of course - you'd have to override that in the returned value or test it separately.

opus/application/apps/ui/views.py
```diff
@@ -80,22 +81,23 @@ def get_context_data(self, **kwargs):
         return context
 
 @never_cache
-def api_last_blog_update(request):
-    """Return the date of the last blog update.
+def api_notifications(request):
+    """Return the HTML for any pending notifications and the date of the last blog update.
```
rfrenchseti: Comment exceeds 80 character limit - shorten the first line to fit within 80 character, and fix rest of comment to fit as well.

opus/application/apps/ui/views.py
```diff
@@ -80,22 +81,23 @@ def get_context_data(self, **kwargs):
...
-    Format: __lastblogupdate.json
+    Format: __notifications.json
 
     JSON return:
-        {'lastupdate': '2019-01-31'}
-      or if none available:
-        {'lastupdate': None}
+        {'lastupdate': '2019-01-31',                (or if none available 'None')
+         'notification': '<html code>',             (or if none available 'None')
+         'notification_mdate': '<file mod date>'   (store as cookie)
```
rfrenchseti: "Store as cookie" is not relevant here - it's not part of the API. What the caller decides to do with this information is up to them.

rfrenchseti: Also give an example of the mod date - what is the format?

opus/application/apps/ui/views.py
```diff
@@ -105,12 +107,31 @@ def api_last_blog_update(request):
...
+            log.error('api_notifications: Failed to read file UNKNOWN')
 
-    ret = json_response({'lastupdate': lastupdate})
+    notification = None
+    notification_modify = None
+    try:
+        with open(settings.OPUS_NOTIFICATION_FILE, 'r') as fp:
+            notification = fp.read().strip()
+            try:
+                notification_modify = os.path.getmtime(settings.OPUS_NOTIFICATION_FILE)
```
rfrenchseti: This line and two below exceed 80 chars

opus/application/apps/ui/views.py
```diff
@@ -105,12 +107,31 @@ def api_last_blog_update(request):
             lastupdate = fp.read().strip()
     except:
         try:
-            log.error('api_last_blog_update: Failed to read file "%s"',
+            log.error('api_notifications: Failed to read file "%s"',
                       settings.OPUS_LAST_BLOG_UPDATE_FILE)
         except:
-            log.error('api_last_blog_update: Failed to read file UNKNOWN')
+            log.error('api_notifications: Failed to read file UNKNOWN')
```
rfrenchseti: Change this line and the other "read file UNKNOWN" error to say whether it's a blog or notification file, since now there are two possibilities.

opus/application/apps/ui/views.py
```diff
@@ -105,12 +107,31 @@ def api_last_blog_update(request):
...
 
-    ret = json_response({'lastupdate': lastupdate})
+    notification = None
+    notification_modify = None
+    try:
+        with open(settings.OPUS_NOTIFICATION_FILE, 'r') as fp:
+            notification = fp.read().strip()
```
rfrenchseti, line 121: I think it would be good to make sure that a completely empty file, and a non-existent file, look the same to an outside user. So perhaps a line after this that says:
    if not notification: notification = None
(this will catch empty strings)
Could do the same for the blog date above to be consistent.

opus/application/apps/ui/test_ui.py
```diff
@@ -31,38 +31,47 @@ def tearDown(self):
...
+        self.assertEqual(ret.content, b'{"lastupdate": null, "notification": null, "notification_mdate": null}')
+
+    def test__api_notifications_bad_notification_file(self):
+        "[test_ui.py] api_notifications: missing notifications.html file"
+        settings.OPUS_NOTIFICATIONS_FILE = 'test_api/data/xyxyxyxyxyx.html'
+        request = self.factory.get('__notifications.json')
+        ret = api_notifications(request)
+        print(ret)
+        print(ret.content)
+        self.assertEqual(ret.content, b'{"lastupdate": "2019-JAN-01", "notification": null, "notification_mdate": null}')
```
rfrenchseti: This test fails because it doesn't override LAST_BLOG_UPDATE_FILE with the good test copy.

opus/application/apps/ui/test_ui.py
```diff
@@ -31,38 +31,47 @@ def tearDown(self):
...
+        "[test_ui.py] api_notifications: missing last blog update file"
         settings.OPUS_LAST_BLOG_UPDATE_FILE = 'test_api/data/xyxyxyxyxyx.txt'
-        request = self.factory.get('__lastblogupdate.json')
-        ret = api_last_blog_update(request)
+        request = self.factory.get('__notifications.json')
+        ret = api_notifications(request)
         print(ret)
         print(ret.content)
-        self.assertEqual(ret.content, b'{"lastupdate": null}')
+        self.assertEqual(ret.content, b'{"lastupdate": null, "notification": null, "notification_mdate": null}')
```
rfrenchseti, line 68: This test fails because it doesn't override OPUS_NOTIFICATIONS_FILE with the good test copy.

opus/application/apps/ui/views.py
```diff
@@ -105,12 +107,31 @@ def api_last_blog_update(request):
...
+            try:
+                notification_modify = os.path.getmtime(settings.OPUS_NOTIFICATION_FILE)
+            except:
+                log.error('api_notification: Failed to read the modify date of file "%s"',
+                          settings.OPUS_NOTIFICATION_FILE)
+    except:
+        try:
+            log.error('api_notifications: Failed to read file "%s"',
```
rfrenchseti: I think we don't want this to be an error. The blog update date should always be there, but it's perfectly normal for there to be no notification file. We don't want that condition logged as an error in the logs. Perhaps make it debug level instead (here and below).

rfrenchseti (changes requested):

opus/application/apps/ui/test_ui.py
rfrenchseti: This was not fixed.

opus/application/apps/ui/test_ui.py
```diff
@@ -31,38 +32,60 @@ def tearDown(self):
...
+
+    def test__api_notifications_html_set(self):
+        "[test_ui.py] api_notifications: notification html set"
+        settings.OPUS_NOTIFICATION_FILE = "test_api/data/test_ui_notification.html"
+        request = self.factory.get('__notifications.json')
+        ret = api_notifications(request)
+        data = json.loads(ret.content)
+        print(data)
+        self.assertEqual(data['notification'], "<div><p>test</p></div>")
+        message = "Test value is none."
```
rfrenchseti: There's no reason to include a message. The assert failing makes it obvious what the problem is.

rfrenchseti (changes requested):
Think a little harder about the test cases. The possible combinations are:

- Blog update file
  - Good
  - Empty
  - Missing
- Notifications file
  - Good
  - Empty
  - Missing

I think you want a minimum of 6 test cases to cover these, perhaps more to cover combinations.

Every test function needs to override both the blog file and the notifications file, to avoid things going bad if the user has something set in opus_settings.

You have a notifications.html file, and you refer to it in a comment, but you don't actually use it in the code.

Make sure the main test comment and function name make sense. What does "default - normal" mean? Describe what you're testing instead.

rfrenchseti (approved):

PR #1152, opened 2021-04-23
Add formats DMS/HMS and datetime

### LINKED ISSUE #599: Add units for RA and DEC to be input as H:M:S and D:M:S
It would be nice to be able to specify RA as H:M:S and DEC as D:M:S

### LINKED ISSUE #1030: Add units for clock time
Currently "clock time" (Observation Start/Stop, Earth Received Time, etc) does not have units in OPUS. While I'm looking at all the UVIS problems, I've founding myself wanting to be able to input and display the time in other formats. I suggest adding units to all time fields:

YMDHMS (this already parses)
YDHMS (this already parses)
JD (in seconds)
TDB (in seconds)

rfrenchseti:
- Fixes #599. Fixes #1030
  - If YES:
    - Database used: New
    - All Django tests pass: Y
  - If YES:
    - Tested on Chrome, Firefox, Safari, Edge, and iOS: N

Description of changes:

- Change the way units are handled
- Add support for new datetime formats
- Add support for new latitude/longitude/hour_angle formats
- There is a new test suite for lib/opus_support.py. Just run "python opus_support.py" in the lib directory.
- You can run this version with the most recent test database, but you have to update the param_table first with "python main_opus_import.py --create-param".
- I slightly expanded the width of input fields so that a complete datetime string would fit in Linux Chrome. I have not been able to test on other browsers because I'm having trouble with BrowserStack, so testing would be greatly appreciated.
- The changes to the requirements are not caused by these changes, but by recent changes to pdsfile.

Known problems:

None.


rfrenchseti:
- Date/time formats:
  - The "julian" module handles the following formats: YYYY-MM-DD[THH:MM:SS.sss], YYYY-DDD[THH:MM:SS.sss], JDxxxx, JEDxxxx, MJDxxxx, MJEDxxxx.
  - All of those formats are supported by just typing them raw into a date/time field, regardless of the currently selected units. In other words, if the current unit is "JD", you can still type "2010-01-01".
  - If you just type a number, it is interpreted according to the current unit selection.
  - Dates are limited to the years 1000-2999; anything outside of that range is considered an error
- RA/DEC/longitude/latitude formats:
  - Nd Nm Ns (degrees-minutes-seconds) or Nh Nm Ns (hours-minutes-seconds).
  - The last field (d/h, m, or s) can have fractions, but none of the earlier fields can. So "1.23d" and "1d 2.34m" are legal but "1.23d 10m" is not.
  - You can type a degrees string during "hours" units or vice versa and it will be interpreted properly.
  - You can type either a degrees or hours string during radians and it will be interpreted properly.
  - If you type a raw number, it will be interpreted according to the current units: degrees or DMS means degrees, hours or HMS means hours, radians mean radians.
- Numeric fields that support other types of units (duration, distance, resolution, etc.) can take an optional suffix to override the unit. So when "seconds" is selected, you can type "1" which means 1 second. Or you can type "1 day" or "1 hour" and it will convert for you. There are a wide range of things that will be understood.
- Spacing and capitalization shouldn't matter for any of the above as long as it doesn't introduce an ambiguity.

rfrenchseti:
One more change: I increased the number of significant digits for wave number and wave number resolution.

juzen2003:
UPDATE: after updating my local copy of pds-tools to have the latest version julian.py, JED & MJED work fine now.

lib/opus_support.py
```diff
@@ -402,178 +686,1252 @@ def format_voyager_sclk(value, sep=':', fields=3):
...
+    if format_func is None:
+        if numerical_format is None:
+            return str(val)
+        if abs(val) >= 1e8:
+            numerical_format = numerical_format.replace('f', 'e')
+        new_format = adjust_format_string_for_units(numerical_format,
+                                                    unit_id, unit)
+        ret = ('{:'+new_format+'}').format(val)
+        if not keep_trailing_zeros and '.' in ret:
+            ret = ret.rstrip('0').rstrip('.')
```
juzen2003: I think the following issue is caused by this, it will strip the last 0 when number is in the format like 8e+10 (got converted to 8e+1). Will have to set keep_trailing_zeros to True when the input has scientific format.
Issue:
- Open Observation duration widget.
- Change unit to days, and type 1 in any input.
- Change unit to usecs, and it will convert 1 day to 8.64e+1 instead of 8.64e+10

Or call this API for the same action:
- http://127.0.0.1:8000/opus/__api/normalizeinput.json?reqno=1&observationduration2_101=1&unit-observationduration_101=microseconds&sourceunit-observationduration_101=days&unit-observationduration_102=microseconds&sourceunit-observationduration_102=days

juzen2003 (changes requested):
Input width for datetime text (Test on latest version of each browser):
- Test example: put "1981-08-12T14:55:10.080" in min input.
- MAC (Big Sur): 
    - Chrome, Safari, Opera, and Edge will have text overflow
    - Firefox will not have text overflow.
- Window 10:
    - Chrome, Firefox, Edge, and Opera will not have overflow.

lib/opus_support.py
rfrenchseti: Fixed. I also fixed the case of 1.000e+10 not turning into 1e+10. New tests have been added.

juzen2003 (approved):
- Database used: opus3_test_210331
- All Django tests pass: Y

PR #1157, opened 2021-05-06
1071 allow selection of download file internal structure

### LINKED ISSUE #1071: Allow selection of download file internal structure
We could permit the user to choose a flat file format or hierarchical mirroring the original PDS directory hierarchy. The GEO Node does this already.

juzen2003:
- Fixes #1071 
  - If YES:
    - Database used: opus3_test_210331
    - All Django tests pass: Y
  - If YES:
    - JSHINT run on all affected files: Y (cart.js)
    - Tested on Chrome, Firefox, Safari, Edge, and iOS: Y (No iOS)

Description of changes:
- Add a "Hierarchical" checkbox below the "Download Options" in the cart pane. 
    - If checkbox is selected, the downloaded zip file will have all files stored with hierarchy tree.
    - If checkbox is not selected, in the zip file:
        - files with the same basename + same logical path (duplicated) will only be stored once.
        - files with the same basename + different logical paths (not duplicated) will be stored within its own hierarchy tree.
- API tests for /api/download/<opusid>.zip & /__cart/download.json are added.
- For local testing purpose, the default selected files (0 byte dummy files) for these opus ids are stored in Shared OPUS dropbox:  
    - co-iss-n1460973661
    - co-iss-n1481265970
    - co-iss-n1462840881
  
Known problems:
N/A


opus/application/apps/cart/views.py
```diff
@@ -516,6 +516,8 @@ def api_create_download(request, opus_id=None):
         or: [__]api/download/(?P<opus_id>[-\w]+).zip
     Arguments: types=<PRODUCT_TYPES>
                urlonly=1 (optional) means to not zip the actual data products
+               hierarchical=0 (optional) means files in zip are not stored with
```
rfrenchseti: Since hierarchical=0 is the default if nothing is specified, this should instead say what happens if you specify hierarchical=1 (like urlonly=1)

opus/application/static_media/css/opus.css
```diff
@@ -3217,6 +3223,12 @@ when wrapping into the 2nd line */
     .op-download-options-title {
         display: none !important;
     }
+
+    /* Make sure when "Download Options" is gone, we create some space to between
```
rfrenchseti: Remove "to"

opus/application/test_api/api_test_helper.py
```diff
@@ -166,3 +167,25 @@ def _run_csv_equal_file(self, url, exp_file):
...
         self.assertEqual(resp, expected)
+
+    def _run_zip_equal(self, url, expected):
+        print(url)
+        response = self._get_response(url)
+        self.assertEqual(response.status_code, 200)
+        if '__cart/download.json' in url:
+            jdata = json.loads(response.content)
+            file = jdata['filename']
+            filename = file[file.rindex('/'):]
```
rfrenchseti: Since we aren't guaranteed what the user has set for TAR_FILE_URL_PATH, how about just stripping off TAR_FILE_URL_PATH (e.g. replace(TAR_FILE_URL_PATH, '') instead of looking for the last "/"?

opus/application/test_api/api_test_helper.py
```diff
@@ -166,3 +167,25 @@ def _run_csv_equal_file(self, url, exp_file):
         print('Expected:')
         print(expected)
         self.assertEqual(resp, expected)
+
+    def _run_zip_equal(self, url, expected):
+        print(url)
+        response = self._get_response(url)
+        self.assertEqual(response.status_code, 200)
+        if '__cart/download.json' in url:
```
rfrenchseti: Rather than check explicitly for download,json and download/ how about passing in a flag that says what kind of response you expect? Then it's still useable if we change the API in the future.

opus/application/test_api/test_cart_api.py
```diff
@@ -2373,22 +2373,97 @@ def test__api_cart_datacsv_multi(self):
...
-            ################################################
-
-    # XXX Need to implement tests
-
-
             ####################################################
             ######### /__cart/download.json: API TESTS #########
             ####################################################
+    def test__api_cart_download_single_no_hierarchical(self):
+        "[test_cart_api.py] /__cart/download.json: single opus id  & no hierarchical"
```
rfrenchseti: Extra space before &

opus/application/test_api/test_cart_api.py
```diff
@@ -2373,22 +2373,97 @@ def test__api_cart_datacsv_multi(self):
...
+        url = '/__cart/add.json?opusid=co-iss-n1462840881&reqno=456'
```
rfrenchseti, line 2384: Should call reset.json at the beginning of each test like the other existing tests do

opus/application/test_api/test_cart_api.py
```diff
@@ -2373,22 +2373,97 @@ def test__api_cart_datacsv_multi(self):
...
+        expected = {'recycled_count': 0, 'count': 1, 'error': False, 'reqno': 456}
+        self._run_json_equal(url, expected)
+        url = '/__cart/download.json?types=coiss_raw,coiss_calib,browse_full&hierarchical=0'
+        expected = ['N1462840881_1.IMG', 'N1462840881_1.LBL', 'prefix2.fmt', 'tlmtab.fmt', 'N1462840881_1_CALIB.IMG', 'N1462840881_1_CALIB.LBL', 'N1462840881_1_full.png', 'manifest.csv', 'data.csv', 'urls.txt']
+        self._run_zip_equal(url, expected)
```
rfrenchseti, line 2389: All of these tests leave .zip files behind in the TAR_FILE_PATH. They need to clean up after themselves by removing the zip file.

rfrenchseti (changes requested):
Now that I look at the wording, I think it would make more sense to have the checkbox say "Flat zip file structure" and be checked by default. We have plenty of room for that with this UI layout. You can keep the flag name and whatnot and just change the meaning of the text.

Also maybe the line that says "Select which product types..." should have a little more visual contrast. Perhaps make it a heavier character weight? Something that makes it look more like a heading, which it really is.

opus/application/test_api/api_test_helper.py
```diff
@@ -166,3 +168,29 @@ def _run_csv_equal_file(self, url, exp_file):
...
+        if response_type == 'json':
+            jdata = json.loads(response.content)
+            file = jdata['filename']
+            filename = file.replace(settings.TAR_FILE_URL_PATH, '')
+            path = settings.TAR_FILE_PATH + filename
+            zip_file = zipfile.ZipFile(path, mode='r')
+        elif response_type == 'binary':
+            binary_stream = BytesIO(response.content)
+            zip_file = zipfile.ZipFile(binary_stream, mode='r')
+        resp = zip_file.namelist()
```
rfrenchseti, line 185: If you're going to use a series of "if...elif" then there needs to be something to catch the last clause, like an "else: assert False" or something. Otherwise if you pass in the wrong argument it will simply crash with an undefined variable. Alternatively you can just replace the "elif" with a plain "else".

opus/application/test_api/api_test_helper.py
```diff
@@ -166,3 +168,29 @@ def _run_csv_equal_file(self, url, exp_file):
...
+
+    def _run_zip_equal(self, url, expected, response_type='json'):
+        print(url)
+        response = self._get_response(url)
+        self.assertEqual(response.status_code, 200)
+        if response_type == 'json':
+            jdata = json.loads(response.content)
+            file = jdata['filename']
+            filename = file.replace(settings.TAR_FILE_URL_PATH, '')
+            path = settings.TAR_FILE_PATH + filename
```
rfrenchseti: Can't you just say file.replace(settings.TAR_FILE_URL_PATH, settings.TAR_FILE_PATH)?

opus/application/static_media/css/opus.css
```diff
@@ -2914,6 +2914,17 @@ when wrapping into the 2nd line */
...
 }
 
+/* This is to have "Flat zip file structure" checkbox to fit in at the bottom of
+"Download Options" */
+.op-download-options-title {
+    margin-bottom: 0 !important
+}
+
+/* Make "Select all product types" and "Deselect all product types" more visible */
+.op-cart-select-all-btn, .op-cart-deselect-all-btn {
```
rfrenchseti: I meant the line above the buttons, not the buttons themselves.

rfrenchseti (changes requested):

rfrenchseti (approved):

PR #1162, opened 2021-05-14
1153 update import to handle new pdsfile changes

### LINKED ISSUE #1153: Update import to handle new COCIRS product types etc
Changes to the data that need to be supported (from Mark):

As we discussed, there are a couple of OPUS re-imports required. 1. Matt and I fixed the exponent error in STIS wavelengths, so a re-import of HSTO* is needed. 2. There was an error in the opus_products rule for COCIRS that overlooked some files in the BROWSE subdirectories. 3. As I mentioned to you, there were a few duplicated records in the GO_0xxx indices. Also, with my latest pdsfile checkin, it now prioritizes new versions of GO images relative to old ones, labeling the latter "Superseded Processing". 4. Not urgent but I noticed that the primary_filespec for NH files is raw in pdsfile, whereas your "whoever loads last wins" rule means that the calibrated are the primary filespecs inside OPUS. This has not mattered yet but it will eventually. I think a simple solution would be to re-import NH but use the 2001 indices first and the 1001 indices second.

juzen2003:
- Fixes #1153 
  - If YES:
    - Database used: re-import on a test database using import_for_tests.sh
    - All Django tests pass: Y
  - If YES:

Description of changes:
- Update tests results for STIS (wavelength values is changed after re-import)
- Make sure all the opus_type & opus_prioritizer in pdsfile rules (latest master branch in pds-webtools) are in internal_def_product_types.json. The newly added opus_type is "gossi_raw_alternate". Tooltips in detail tab for GOSSI superseded processing files shows up after re-import.
- For NH volset, update to make sure 2001 is imported first and then 1001 after that. Update the tests results for NH after re-import.

Known problems:
NA


lib/opus_support.py
```diff
@@ -1947,7 +1947,25 @@ def parse_form_type(s):
 
     return form_type, form_type_format, form_type_unit
 
-
+def get_single_parse_function(form_type_unit_id):
+    range_func = None
```
rfrenchseti: We don't use the term "range_func" anymore so this one should be called "parse_func' and the next should be "format_func".

lib/opus_support.py
```diff
@@ -1947,7 +1947,25 @@ def parse_form_type(s):
 
     return form_type, form_type_format, form_type_unit
 
-
+def get_single_parse_function(form_type_unit_id):
```
rfrenchseti: Add docstring: """Return the parse func for a unit_id with a single non-displayed unit."""

lib/opus_support.py
```diff
@@ -1947,7 +1947,25 @@ def parse_form_type(s):
...
+    range_func = None
+    if form_type_unit_id and not display_unit_ever(form_type_unit_id):
+        default_unit = get_default_unit(form_type_unit_id)
+        range_func = (UNIT_FORMAT_DB[form_type_unit_id]
+                                    ['conversions']
+                                    [default_unit]
+                                    [2])
+    return range_func
+
+def get_single_format_function(form_type_unit_id):
```
rfrenchseti: Add docstring: """Return the format func for a unit_id with a single non-displayed unit."""

lib/opus_support.py
rfrenchseti: Likewise just call it "unit_id" not "form_type_unit_id" to match the rest of this file.

opus/import/do_import.py
```diff
@@ -431,12 +431,10 @@ def update_mult_table(table_name, field_name, table_column, val, label,
...
         # No disp_order specified, so make one up
         # Update the display_order
-        form_type = table_column['pi_form_type']
-        range_func = None
-        if form_type is not None and form_type.startswith('GROUP:'):
-            range_func_name = form_type.replace('GROUP:', '')
-            if range_func_name in opus_support.RANGE_FUNCTIONS:
-                range_func = opus_support.RANGE_FUNCTIONS[range_func_name][1]
+        (form_type, form_type_format,
+         form_type_unit_id) = opus_support.parse_form_type(
```
rfrenchseti: You removed the `import opus_support` so you definitely shouldn't reference `opus_support` like this. I'm actually surprised this works. I would do what the other files do. See, for example, the top of `apps/search/views.py`. Then remove the `opus_support.` entirely.

opus/import/do_import.py
```diff
@@ -431,12 +431,10 @@ def update_mult_table(table_name, field_name, table_column, val, label,
...
+                            table_column['pi_form_type'])
+        range_func = get_single_parse_function(form_type_unit_id)
```
rfrenchseti: Again let's not call this a "range_func" but instead a "parse_func".

rfrenchseti (changes requested):
See if you can find a GOSSI observation that has superseded data to include in test_results_contents. When you run create_test_results.py it will print out the Python code for the associated tests which you can then just copy into test_results_contents.py as the appropriate location.

rfrenchseti (approved):

PR #1176, opened 2021-07-06
1144 indicator for detail

### LINKED ISSUE #1144: Add indicator on gallery view of what OPUS ID is on detail tab
When there is an OPUS ID present for the Detail Tab, then whenever that OPUS ID shows up in the gallery, put a (i) icon in the corner to indicate that's the observation being "detailed".

dstopp:
- Fixes #1144 
  - If YES:
    - Database used: opus3_test_210526
    - All Django tests pass: Y
    - FLAKE8 run on modified code: N
  - If YES:

Description of changes:
-add 'DETAIL' tag to current selected detail opus object
-add mini-icon/image to the Detail tab to show detail has a selected value; hover to get opus id
-fix the regression tests for windows delete

Known problems:


opus/application/static_media/js/detail.js
```diff
@@ -128,6 +128,23 @@ var o_detail = {
         o_detail.adjustDetailHeight();
     },
 
+    showDetailThumbInNav: function(imageHtml) {
+        if (imageHtml === undefined) {
+            if (opus.prefs.detail != "") {
+                let url = "/opus/api/image/thumb/" + opus.prefs.detail + ".json";
```
rfrenchseti: This should use the internal API URL "__api/image/thumb/" so that we don't count these as uses by an external API user. We haven't needed this before so you'll have to add it to urls.py by just copying the normal "api/image" line and adding a new one with the "__" in front. Everything else can stay the same. The internal functionality doesn't change - we just use it when parsing the Apache logs.

opus/application/static_media/js/browse.js
```diff
@@ -1242,6 +1242,15 @@ var o_browse = {
 
     showDetail: function(e, opusId) {
         o_browse.hideMenus();
+
+        // if the item is selected as 'detail', add the icon in the corner
+        // hide all first... this will hide both on browse and cart
```
rfrenchseti: Hide all what first? And why are you hiding anything? This comment is confusing. And what "icon" in the corner? Do you mean the word "detail"?

rfrenchseti (changes requested):

opus/application/static_media/js/browse.js
```diff
@@ -1242,6 +1242,15 @@ var o_browse = {
...
+        $(`.op-thumbnail-container .op-detail-overlay`).addClass("op-hide-element");
```
juzen2003: Please change `` to "" since there is no string interpolation here.

juzen2003 (changes requested):

rfrenchseti (approved):

Issue #972, opened 2020-01-31
Switch to using a non-native tooltip library (and add links)


rfrenchseti:
Under Mojave/Safari, tooltips disappear in 10 seconds, but they don't under Catalina. Michael Aye would like them to not disappear at all under any browser or OS, which I agree with.

matthewtiscareno:
What would one do about it?

rfrenchseti:
There are always solutions. Tooltips are not actually a browser feature, they are implemented using the "jQuery UI" library: https://jqueryui.com/tooltip/

Thus we can override that library's behavior, use another third-party library, or implement our own version of tooltips. In fact there is already a need for a different type of tooltip, because "mouse hover" doesn't exist on mobile devices:

https://github.com/SETI/pds-opus/issues/796

A fancier library would also allow things like links inside the tooltips that can be clicked on:

http://qtip2.com/

matthewtiscareno:
Okay, so the current problem is that different browsers do different things when they see a jQueryUI, and in some cases we don't like the results.  The answer is to use something other than jQueryUI, since that doesn't work for mobiles anyway.  Have I got it?

rfrenchseti:
In the long term, yes. In the short term we'll try to find a workaround using jQuery UI for this one particular problem because it will be easier if we can fix a user's complaint without redesigning the whole tooltip system.

juzen2003:
I think we can't fix the disappearing native tooltip from title attribute in Safari 12 without converting our tooltip design to either jQuery, Bootstrap or qTip. If we tried native tooltips from other sites in Safari 12, it will disappear too. For example, the tooltips in https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title

I tried with jQuery & Bootstrap last night and see some difference between native tooltips and ones from libraries. To create tooltips with libraries, we need to fix at least the following issues:
1. Tooltips can't go off the viewport. Unlike the native ones, tooltips from libraries will have to stay inside the body tag (or html tag) and can't go off the screen.
2. Since tooltips can't go off the screen, we need to make sure all our tooltips will fit in the screen when the screen is very small. Some of our tooltips have lots of texts. 
3. Sometimes tooltips need to cover the (i) icon to have the full tooltips info displayed on the screen, this will cause the tooltips flickering issue due to these:
      a. When mouseover (i) icon and leave the mouse within (i) icon, tooltips pop out.
[...]

Issue #425, opened 2018-08-15
Need tooltips for mult options


rfrenchseti:
The dictionary supports hover-over tool tips for mult options, but we don't have any definitions or support in the OPUS UI.

rfrenchseti:
BYBTW 6/15/2021 - Mark wants the priority on this increased

PR #1179, opened 2021-07-20
Add tooltips on mults inputs

juzen2003:
Progress of implementation:
- Use tooltipster to replace the default tooltips
- Currently only mission id widget has tooltips on mult inputs. 
    - Tooltips there are having dark background (can be changed later) with a link pointing to https://pds-rings.seti.org/. It will be opened in a new tab when being clicked.
- When adding tooltipster to mult widget inputs, we will write our own html for mult widget inputs instead of using Django form. It's easier to do the customization in this way.
- Tooltips are stored in the preprogrammed "mult_options" in table schema.

Pending item:
- Implement tooltipster on one of mult widgets with grouping.
- For mult tables without tooltips, maybe we don't want to have tooltips field (with all null values) in the database. (?)
- Do we want all mult widgets to have tooltips on each input?

Note: Need to import a test db because we store tooltips in the database.


juzen2003:
- Fixes #972 #425 
  - If YES:
    - Database used: Import a new database
    - All Django tests pass: Y
  - If YES:

Description of changes:
- Replace the default tooltips with tooltipster
- Make the ability to add tooltips on mult options available.
- To update the options of all tooltipster instances, change the variables from line 155-157 in ```opus.js```

Note: 
To put on random tooltips on some options in coiss filter and coiss target description mult options, please do the followings:
- Before running import, uncomment ```TOOLTIPS_FOR_MULT``` in ```config_data.py```
- In ```settings.py```, uncomment line 379 and comment out line 377
After import is done, the tooltips in ```TOOLTIPS_FOR_MULT``` will show up.

Known problems:
- Current theme is light, will need to finalize the theme or fine tune the tooltip to what we like.

juzen2003:
Close and move to https://github.com/SETI/pds-opus/pull/1239

PR #1225, opened 2022-03-15
Pull request to track the progress of converting to use Django 4

juzen2003:
Related to https://github.com/SETI/pds-opus/issues/1084

Changes done to work with Django 4:
- Change ```from django.conf.urls import url``` to ```from django.urls import re_path```
    - change the ```url``` call to ```re_path``` call
- Change ```from django.shortcuts import render_to_response``` to ```from django.shortcuts import render```
    - change ```render_to_response``` call to ```render``` call and make sure to pass in ```request```
- Add DEFAULT_AUTO_FIELD = 'django.db.models.AutoField' to settings.py to get rid of the warnings. (https://koenwoortman.com/python-django-auto-created-primary-key-used-when-not-defining-primary-key-type/)
- Add ```'django.template.context_processors.request'``` in context processors in settings.py
- Get rid of django-nose. Django nose will only support up to Django 2.2 and Python 3.7. (https://pypi.org/project/django-nose/) We will removed it from ```INSTALLED_APPS``` and remove ```TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'``` in settings.py.
    - Note: we can still make django-nose work with Django 4 by changing line 406 in ```runner.py``` from: ```flush_statements = connection.ops.sql_flush(
        style, tables, connection.introspection.sequence_list())``` to ```flush_statements = connection.ops.sql_flush(
        style, tables)```

Pending items:
- RadioSelect and CheckboxSelectMultiple widgets are now rendered in ```<div>``` instead of ```<li>```
    - Two ways to solve this:
        1. Override the widget template as suggested from https://docs.djangoproject.com/en/4.0/releases/4.0/ to keep the previous form behavior.
        2. Update/modify our current styling and codes so that new template will work.
- Modify tests with no request passed in. Now never cache will raise a type error when no request is passed in. Here is the code differece between 2.2.6 & 4.0.3:
2.2.6:
```
def never_cache(view_func):
    """
    Decorator that adds headers to a response so that it will never be cached.
    """
    @wraps(view_func)
    def _wrapped_view_func(request, *args, **kwargs):
        response = view_func(request, *args, **kwargs)
        add_never_cache_headers(response)
        return response
    return _wrapped_view_func
```
4.0.3
```
def never_cache(view_func):
    """
    Decorator that adds headers to a response so that it will never be cached.
    """

    @wraps(view_func)
    def _wrapped_view_func(request, *args, **kwargs):
        # Ensure argument looks like a request.
        if not hasattr(request, "META"):
            raise TypeError(
                "never_cache didn't receive an HttpRequest. If you are "
                "decorating a classmethod, be sure to use @method_decorator."
            )
        response = view_func(request, *args, **kwargs)
        add_never_cache_headers(response)
        return response

    return _wrapped_view_func
```
We can do something like this: (use ```test__api_notifications_no_request``` as an example)
```
    def test__api_notifications_no_request(self):
        "[test_ui.py] api_notifications: no request"
        try:
            with self.assertRaisesRegex(Http404,
                r'Internal error \(No request was provided\) for /__notifications.json'):
                api_notifications(None)
        except TypeError as e:
            # In Django 4.0, never_cache will raise an error when no request is passed in.
            if "didn't receive an HttpRequest" in str(e):
                assert True
            else:
                assert False, "Unexpected error when no request was provided."
```

### DIFF
--- a/opus/application/apps/cart/test_cart.py
+++ b/opus/application/apps/cart/test_cart.py
@@ -37,9 +37,11 @@ class cartTests(TestCase):
 
     def test__api_view_cart_no_request(self):
         "[test_cart.py] api_view_cart: no request"
+        no_request = self.factory.get('dummy')
+        no_request.META = None
         with self.assertRaisesRegex(Http404,
             r'Internal error \(No request was provided\) for /__cart/view.html'):
-            api_view_cart(None)
+            api_view_cart(no_request)
 
     def test__api_view_cart_no_get(self):
         "[test_cart.py] api_view_cart: no GET"
@@ -56,9 +58,11 @@ class cartTests(TestCase):
 
     def test__api_cart_status_no_request(self):
         "[test_cart.py] api_cart_status: no request"
+        no_request = self.factory.get('dummy')
+        no_request.META = None
         with self.assertRaisesRegex(Http404,
             r'Internal error \(No request was provided\) for /__cart/status.json'):
-            api_cart_status(None)
+            api_cart_status(no_request)
 
     def test__api_cart_status_no_get(self):
         "[test_cart.py] api_cart_status: no GET"
@@ -75,9 +79,11 @@ class cartTests(TestCase):
 
     def test__api_get_cart_csv_no_request(self):
         "[test_cart.py] api_get_cart_csv: no request"
+        no_request = self.factory.get('dummy')
+        no_request.META = None
         with self.assertRaisesRegex(Http404,
             r'Internal error \(No request was provided\) for /__cart/data.csv'):
-            api_get_cart_csv(None)
+            api_get_cart_csv(no_request)
 
     def test__api_get_cart_csv_no_get(self):
         "[test_cart.py] api_get_cart_csv: no GET"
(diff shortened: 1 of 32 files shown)


juzen2003:
Override the widget templates for RadioSelect and CheckboxSelectMultiple by doing the followings:
- Add ```'django.forms'``` to INSTALLED_APPS in settings.py
- Use templatesSetting as form renderer by adding ```FORM_RENDERER = 'django.forms.renderers.TemplatesSetting'``` in settings.py.
- Change ```self.fields.keyOrder``` to ```self.field_order```
- Create widgets folder (```application/templates/django/forms/widgets```) to store Django 3.2 templates. This is to override the Django 4.0 RadioSelect
 and CheckboxSelectMultiple widgets.

Pending item:
- Modify tests with no request passed in.

juzen2003:
- Update tests with no http request to take in a dummy request with META set to None. This is to avoid the TypeError raised by never_cache (detailed code in the first comment) in Django 4.
- Pending Item: None

juzen2003:
- Verified with Python ```3.10.2```
- For tests to be discovered by ```python manage.py test```, we have to make ```test_api``` and ```test_db_data``` to be modules by adding ```__init__.py```. 
- Please use ```python manage.py test -b``` to run test, this is to suppress the sys.stdout, and if there is an error or a failure, the detailed logs will still show up.

rfrenchseti (approved):
Tested with Python 3.10 and compared with old version for # tests, test results, and pixel-by-pixel comparison of various displayed pages.

dstopp (approved):

PR #1232, opened 2022-05-06
Jinja generators can generate Markup

fyellin:
Jinja generators can produce Markup in addition to strings.  Fix code to handle this.  Also fix a typo.


fyellin:
I'm closing this pull request.  Not sure what happened, but it shouldn't be this big.

PR #1239, opened 2022-06-09
Add tooltips on mult inputs

### LINKED ISSUE #425: Need tooltips for mult options

### LINKED ISSUE #972: Switch to using a non-native tooltip library (and add links)

juzen2003:
- Fixes #972. Fixes #425.
- This is the continuation of https://github.com/SETI/pds-opus/pull/1179
- Use the master as the base and merge the old tooltips branch.

- Current Implementation:
    - Create mult_tooltips.json to store tooltips (HTML markup are allowed in the content) for mult options.
    - Add mult tooltips context name to contexts.csv
    - Store mult tooltips to dictionary (definitions table)
    - In definitions table, the term will be the same as the mult option value in the mult tables, and context will be ```MULT_{slug.upper()}```. And the definitions (tooltips) will be fetched by ```get_def_for_tooltip later```.
    - Since tooltips are stored in definitions table, we can update tooltips by ```python main_opus_import.py --import-dictionary``` instead of re-import the whole database.

Note: for now testing toolstips are added for the options in mission widget. We can see the tooltips with different HTML markup when hovering over Galileo, New Horizons, and Voyager


opus/import/config_data.py
```diff
@@ -152,3 +152,21 @@
     63: 'Madrid 70m',
     65: 'Madrid 34m'
 }
+
```
rfrenchseti: This can go away, right?

opus/import/import_util.py
```diff
@@ -243,6 +243,17 @@ def safe_column(row, column_name, idx=None):
         return None
     return row[column_name][idx]
 
+def get_mult_tooltip(table_name, field_name, option_val):
```
rfrenchseti: This can go away, right?

opus/import/do_dictionary.py
```diff
@@ -31,6 +31,10 @@ def create_import_definitions_table():
     json_def_path = (opus_secrets.DICTIONARY_JSON_SCHEMA_PATH +
                      '/internal_def*.json')
     json_list += glob.glob(json_def_path)
+    # Tooltips for mult
```
rfrenchseti: for *mults*

opus/import/do_dictionary.py
```diff
@@ -31,6 +31,10 @@ def create_import_definitions_table():
     json_def_path = (opus_secrets.DICTIONARY_JSON_SCHEMA_PATH +
                      '/internal_def*.json')
     json_list += glob.glob(json_def_path)
+    # Tooltips for mult 
+    json_tooltips_def_path = (opus_secrets.DICTIONARY_JSON_SCHEMA_PATH +
```
rfrenchseti, line 35: I think we want to do the same thing here that's done for obs*.json above . Allow multiple mult_tooltips files like mult_tooltips_mission and mult_tooltips_instrument. Otherwise the file is going to get enormous, and this allows multiple people to work on different widgets at the same time. Since you're already using glob this just means changing the filename to mult_toolips*.json.

opus/import/table_schemas/mult_template.json
```diff
@@ -37,6 +37,11 @@
         "field_type": "char100",
         "field_default": null
     },
+    {
```
rfrenchseti: This can go away, right? We don't store the tooltip in the mult table anymore.

opus/import/do_import.py
```diff
@@ -1204,17 +1205,17 @@ def import_observation_table(instrument_obj,
             if field_type.startswith('flag'):
                 if column_val in [0, 'n', 'N', 'no', 'No', 'NO', 'off', 'OFF']:
                     if field_type == 'flag_onoff':
-                        column_val = 'Off'
```
rfrenchseti: What changed to require this? Isn't mult_label set further down below based on colum_val.title()?

opus/application/apps/dictionary/views.py
```diff
@@ -13,8 +13,10 @@ def get_def_for_tooltip(term, context):
         entry = Definitions.objects.get(context__name=context, term=term)
         return entry.definition
     except Definitions.DoesNotExist:
-        log.error('No tooltip definition for context "%s" term "%s"',
-                  context, term)
+        # We allow mult tooltips to be None
+        if not context.startswith('MULT'):
```
rfrenchseti: Change to MULT_ just to be cautious

dictionary/contexts.csv
```diff
@@ -21,3 +21,4 @@ NEWHORIZONS,New Horizons Mission,OPUS_GENERAL
 NHLORRI,New Horizons Long-Range Reconnaissance Imager (LORRI),NEWHORIZONS
 NHMVIC,New Horizons Multispectral Visible Imaging Camera (MVIC),NEWHORIZONS
 OPUS_PRODUCT_TYPE,OPUS Product Types,NULL
+MULT_MISSION,OPUS Mission Ids,NULL
```
rfrenchseti: This isn't needed for now since we never actually use the contexts table. Eventually we'll want to add contexts automatically as the mult tooltips are loaded.

opus/application/apps/ui/templates/ui/customized_input.html
```diff
@@ -0,0 +1,28 @@
...
+    <ul id="{{ slug }}" class="multichoice">
+        {% for id, choice, tooltip, tp_id in options %}
+            <li>
+                <label for="{{ slug }}_{{ id }}">
+                    {% if choice in selections %}
+                        <input type="checkbox" name="{{ slug }}" value="{{ choice }}" class="multichoice" id="{{ slug }}_{{ id }}" checked>
+                    {% else %}
+                        <input type="checkbox" name="{{ slug }}" value="{{ choice }}" class="multichoice" id="{{ slug }}_{{ id }}">
+                    {% endif %}
+                <!-- Assign an unique html structure to data-tooltip-content for each different individual tooltip-->
```
rfrenchseti: a unique

opus/application/apps/ui/templates/ui/customized_input.html
```diff
@@ -0,0 +1,28 @@
...
+                <span class="op-choice-label-name op-mult-tooltip-{{ slug }}" data-tooltip-content="#tooltip-content-{{ slug }}-{{ tp_id }}">
+                    {{ choice }}
+                </span>
+                <!-- Customized tooltip contents -->
+                {% if tooltip %}
+                    <div class="tooltip-templates">
+                        <span id="tooltip-content-{{ slug }}-{{ tp_id }}">
+                            <strong>{{ tooltip|safe }}</strong>
+                            <!-- <a class="op-mult-tooltip-link" href="https://pds-rings.seti.org/" target="_blank"><strong>{{ tooltip }}</strong></a> -->
```
rfrenchseti: Remove comment

opus/application/apps/ui/views.py
```diff
@@ -283,6 +284,14 @@ def _update_form_with_grouping(form, form_vals, glabel, gvalue):
     auto_id = True
     selections = {}
     extras = {}
+    customized_input = slug in settings.MULT_WIDGETS_WITH_TOOLTIPS
```
rfrenchseti: We shouldn't need MULT_WIDGETS_WITH_TOOLTIPS anymore. Instead look through the mult values and see if any of them have tooltips. If they do, then you need a customized input. If they're all None, then you don't need one.

opus/application/settings.py
```diff
@@ -373,3 +373,8 @@
 }
 
 DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
+
+# Mult widgets that required tooltips
+MULT_WIDGETS_WITH_TOOLTIPS = ['mission']
```
rfrenchseti: Get rid of

opus/application/static_media/js/browse.js
```diff
@@ -65,6 +70,22 @@ var o_browse = {
     addBrowseBehaviors: function() {
         // note: using .on vs .click allows elements to be added dynamically w/out bind/rebind of handler
 
+        // Get the x coordinate of the current cursor when moving mouse in the table view.
+        // Also when moving around in the same table row, reposition the tooltip so that
+        // it stays right next to the cursor. Work for both browse/cart data table.
+        $(".op-data-table tbody").on("mousemove", function(e) {
+            // We will wait for the stop of mouse moving, and then show the tooltip
```
rfrenchseti: for mouse moving to stop

opus/application/static_media/js/browse.js
```diff
@@ -1858,6 +1883,46 @@ var o_browse = {
...
+            contentAsHTML: true,
+        });
+        $(".op-browse-table-tooltip").tooltipster({
+            maxWidth: opus.tooltips_max_width,
+            theme: opus.tooltips_theme,
+            delay: opus.tooltips_delay,
+            debug: false,
+            contentAsHTML: true,
+            functionBefore: function(instance, helper){
+                // Make sure all other tooltips is closed before a new one is open
```
rfrenchseti: are closed. is opened.

opus/application/static_media/js/browse.js
```diff
@@ -1858,6 +1883,46 @@ var o_browse = {
...
+                    inst.close();
+                });
+            },
+            // Make sure the tooltip position is next to the cursor when users
+            // move around the same row in the browse table view.
+            functionPosition: function(instance, helper, position){
+                let tooltipWidth = position.size.width;
+                let offsetToRightWindow = 5;
+                let windowWidth = helper.geo.window.size.width;
+                // When the cursor is very close to the right edget of the window, we have
```
rfrenchseti: right edge.

opus/application/static_media/js/opus.js
```diff
@@ -151,6 +151,11 @@ var opus = {
 
     oldSurfacegeoTarget: null,
 
+    // Tooltipster settings
+    tooltips_max_width: 300,
+    tooltips_theme: "tooltipster-light",
+    tooltips_delay: 1500,
```
rfrenchseti: I think this is pretty slow. Normal tooltips pop up a lot faster. Like maybe a half second? Is there any way to have the (i) info buttons show up fast and the mult values take longer?

opus/application/static_media/js/widgets.js
```diff
@@ -437,6 +437,10 @@ var o_widgets = {
             } else {
                 lastExistingSetOfInputs.append(orLabel);
                 cloneInputs.find(".op-or-labels").remove();
+                // remove the cloned trash icon, and attach a new so that the tooltipster
```
rfrenchseti: a new <what> so...

rfrenchseti (changes requested):

opus/application/static_media/js/opus.js
juzen2003: we can assign different delays for different tooltips. I modified it to have 1.5 second delay for mult tooltips and 0.5 second delay for the rest of tooltips.

dictionary/contexts.csv
juzen2003: There is a constraint that requires the context field in definitions to have a foreign key to contexts table name field. If we remove the entry in context.csv, we will get the following error when running import.
```
Failed to insert row into "imp_definitions": Cannot add or update a child row: a foreign key constraint fails (`opus3_test_group_inst_id`.`imp_definitions`, CONSTRAINT `imp_definitions_ibfk_1` FOREIGN KEY (`context`) REFERENCES `imp_contexts` (`name`) ON UPDATE CASCADE)
```

rfrenchseti: OK, but then this isn't the way to do it because you'd have to add a new context to this file every time you wanted to add tooltips to a new mult. That is an unnecessary dependency. Instead, import_dictionary should create the contexts for each mult context as it's needed.

juzen2003:
Update ```create_import_definitions_table``` to have steps to create a mult widget entry in ```contexts``` table before we create tooltips entries in ```definitions``` table for that mult widget. By adding this step, ```contexts.csv``` will remain untouched.
Here is the current entry in ```contexts``` table:
```
new_row = {
    'name': f'MULT_{slug.upper()}',
    'description': f'OPUS {slug.title()}',
    'parent': 'NULL'
}
```

opus/application/static_media/css/opus.css
```diff
@@ -3475,3 +3475,23 @@ when screen is narrow (cart tab) */
...
+}
+
+/* Make sure tooltipster tooltip is not covered by popover */
+.tooltipster-show {
+    z-index: 2000000002 !important;
+}
+
+/* set font size and color of the tooltips contents */
+.tooltipster-content {
+    font-size: 13px;
```
dstopp, line 3622: I think the font size should be responsive; looks like it stays at 13px even when the screen size is small.

opus/application/static_media/js/browse.js
```diff
@@ -1878,7 +1943,7 @@ var o_browse = {
         // check all box
         // let addallIcon = "<button type='button' data-toggle='modal' data-target='#op-addall-to-cart-modal' " +
         let addallIcon = "<button type='button'" +
-                         "class='op-table-header-addall btn btn-link'" +
+                         "class='op-table-header-addall btn btn-link op-addall-icon-tooltip'" +
```
dstopp: Please consider renaming "op-addall-icon-tooltip" to "op-icon-tooltip-addall" for class naming consistency

opus/application/static_media/js/menu.js
```diff
@@ -68,6 +68,13 @@ var o_menu = {
             clearTimeout(spinnerTimer);
 
             o_menu.wrapTriangleArrowAndLastWordOfMenuCategory("#search");
+            // Initialize all tooltips using tooltipster in menu.html
+            $(".op-menu-tooltip").tooltipster({
+                maxWidth: opus.tooltips_max_width,
+                theme: opus.tooltips_theme,
+                delay: opus.tooltips_delay,
+                debug: false,
```
dstopp: do the other tipsters need "debug: false" as well, or can this be removed?

opus/application/static_media/js/opus.js
```diff
@@ -1495,4 +1509,11 @@ $(document).ready(function() {
     // Call normalized url api first
     // Rest of initialization process will be performed afterwards
     opus.normalizedURLAPICall();
+    // Initialize all tooltips using tooltipster in base.html
+    $(".op-base-tooltip, .op-browse-view-tooltip").tooltipster({
+        maxWidth: opus.tooltips_max_width,
+        theme: opus.tooltips_theme,
+        delay: opus.tooltips_delay,
+        debug: false,
```
dstopp: same question; does "debug: false" need to be here or in any of the others?

opus/application/static_media/js/widgets.js
```diff
@@ -1442,6 +1461,21 @@ var o_widgets = {
             if (!opus.isAnyNormalizeInputInProgress()) {
                 opus.updateOPUSLastSelectionsWithOPUSSelections();
             }
+
+            // TESTING
```
dstopp: is this still test code?

dstopp (changes requested):

juzen2003:
- Renamed ```op-addall-icon-tooltip``` to ```op-icon-tooltip-addall```
- Shrink the tooltips font size to 90% (original size: 13px) when screen is small
- Remove all ```debug: false``` in tooltipster initialization
- Rename ```op-metadatabox-tooltip``` to ```op-metadatabox-edit-tooltip``` in ```metadataboxHtml```
- Fixed the issue of missing metadata preview image by changing selector from ```#galleryViewContents``` to ```.op-metadata-detail-view-body``` in ```updateMetadataDetailView```
- Add tooltips for "x", maximize and minimize icons in metadatabox

rfrenchseti:
- The style of the tooltips is inconsistent. Some places, like the Search tab, have black on white with the tooltip appearing at the cursor location. Others have black on grey, with the tooltip appearing above the item. I think this is confusing. At the very least, they should have the same color style. But unless there's a good reason, I think they should all have the same behavior relative to the mouse. An example of "good reason" might be on the gallery tab, where having the tooltip not block the thumbnail is actually quite nice.
- The cursor should change to an arrow when over the (i) in the title of a search widget.
- When hovering over the image in the slideshow view, the tooltip shows up at the very top of the image, far away from the cursor position. This makes it hard to notice.
- On the cart tab, the tooltips for "Data Archive" etc are hard to see above the buttons.

juzen2003:
- Reduce tooltip max width to 350px.
- Update styling of tooltips (op-custom-tooltip-theme) to have black font on white background with black border.
- Make sure the cursor on (i) of search widgets is the default style.
- Make sure all tooltips are converted to use tooltipster.
- Make sure when hovering over metadatabox preview image, detail preview image, and table thumbnail image, the tooltips will follow the cursor instead of staying on the top edge of the element.

rfrenchseti:
On the Browse Results tab, the "Sort by" and actual sort pills are supposed to have two-line tooltips but they're jammed together on a single line.

rfrenchseti:
For some reason the "Sort by" tooltip on the Cart tab only shows one of the two lines. The part that says "Click to reset" is missing entirely. It's there on the production version.

rfrenchseti (changes requested):
When you click on the "cart add" or "cart trash" icon in either the gallery view or the slideshow, OPUS tries to change the tooltip text but ends up adding an old-style tooltip so there are now two tooltips on that icon, one saying "Add" and one saying "Remove".

opus/import/table_schemas/mult_tooltips_mission.json
```diff
@@ -0,0 +1,32 @@
+[
```
rfrenchseti: This file should just be removed entirely.

rfrenchseti (changes requested):

rfrenchseti (approved):

opus/application/apps/cart/templates/cart/cart.html
```diff
@@ -20,16 +20,16 @@
...
                 </li>
                 <li class="list-inline-item">
                     <div id="create_zip_url_file">
-                        <a href="#" class="downloadURL btn btn-outline-primary" title="Download a zipped URL archive of all observations in the cart using the selected products"><i class="fas fa-download"></i>&nbsp;URL Archive</a>
+                        <a href="#" class="downloadURL btn btn-outline-primary op-cart-tooltip" title="Download a zipped URL archive of all observations in the cart using the selected products"><i class="fas fa-download"></i>&nbsp;URL Archive</a>
                     </div>
                 </li>
                 <li class="list-inline-item">
-                    <a href="#" class="op-download-csv btn btn-outline-primary" download title="Download a CSV of selected metadata for all observations in the cart"><i class="fas fa-file-csv"></i>&nbsp;Metadata CSV</a>
+                    <a href="#" class="op-download-csv btn btn-outline-primary op-cart-tooltip" download title="Download a CSV of selected metadata for all observations in the cart"><i class="fas fa-file-csv"></i>&nbsp;Metadata CSV</a>
```
dstopp, line 32: I don't believe "download" belongs (by itself) in the html part of this <a> tag

opus/application/settings.py
```diff
@@ -373,3 +373,6 @@
 }
 
 DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
+
+# We don't want to have these characters in HTML class or ID
+INVALID_CLASS_CHAR = '~!@$%^&*()+=,./;:"?><[]\{}|`# '
```
dstopp, line 381: Perhaps a more detailed explanation as to why we care about this, ie., "for tooltips" or something of that nature.

opus/application/static_media/css/opus.css
```diff
@@ -489,6 +489,10 @@ ul.op-search-menu {
     cursor: grab;
 }
 
+.widget .fa-info-circle:hover {
+    cursor: default;
```
dstopp, line 501: I think all the .fa-info-circle:hover <and perhaps other tooltip locations> should have the same cursor

opus/application/static_media/js/opus.js
```diff
@@ -151,6 +151,19 @@ var opus = {
...
 
+    // Tooltipster settings
+    tooltipsMaxWidth: 680,
+    downloadPaneTooltipsMaxWidth: 350,
+    tooltipsTheme: "op-custom-tooltip-theme",
+    tooltipsDelay: 500,
+    multTooltipsDelay: 1500,
+
+    // the x & y coordinate of the current cursor when moving mouse
+    mouseX: 0,
```
dstopp, line 164: is this specifically for tooltips or general purpose?

dstopp (changes requested):

opus/application/apps/cart/templates/cart/cart.html
rfrenchseti: Actually it does. See, for example: https://www.w3schools.com/TAGS/att_a_download.asp

opus/application/static_media/js/opus.js
juzen2003: This is tracking the mouse position, and that mouse position will be used to set the proper position for the tooltips. It can be used to other places where we need to track the position of the mouse, but for now, it's only used for setting the the position of the tooltips.

dstopp: perhaps it would be useful to add just a bit more detail per your comment above.

juzen2003:
- Add spaces between tooltips box and left & right edges of the browser window.
- Increase left & right paddings of the tooltips contents.

rfrenchseti (approved):

PR #1241, opened 2022-06-24
1216 include all preview browse products in detail tab

### LINKED ISSUE #1216: Include all browse/preview products on Detail tab
Right now only the main preview image is shown on the Detail tab (the same one shown on the Browse tab). All of the browse/preview images should be shown instead.

juzen2003:
- Fixes #1216
- Were any Django, import pipeline, or table_schema files modified? N
  - If YES:
    - Database used: opus3_test_220530
    - All Django tests pass: Y
  - If YES:
- Was the documentation reviewed for necessary changes or additions? NA
  - About
  - Getting Started
  - FAQ
  - API Guide
  - Tooltips

Description of changes:
- Use jquery plugin FlexSlider to display multiple preview images in detail tab
- Update CSS for responsiveness 
- In detail tab, wrap the overflow text of the detail entry values when window size is small. This will solve the issue of left metadata container taking up too much space and preview image is cut off when the browser is narrow. (ex: co-rss-occ-2008-039-rev058c-x43-i)


rfrenchseti:
Is there some way to make the dots a little closer together so when we look at RSS occultations they don't go to two lines?

juzen2003:
Now the dots will stay in one line when the screen is narrow and preview image is in the right container.

juzen2003:
Fixed the twitching issue when the cursor is at the border of preview image and "<" & ">" buttons by updating the stylings to make sure "<" & ">" stay in the same container of the preview image. The twitching issue is the caused by the transition animation. When the positions of "<" & ">" at the beginning of the animation are outside of the main container of the preview image (the target we want to apply the carousel),  if we mouse over the cursor to those non-overlapping area, the twitching issue will happen.

dstopp:
- Is there a way to make the prev/next arrows smaller when the screen is resized?  the dots as well?  they seem disproportionately big.

- When the screen is approx 730x560, the arrows are not evenly placed on either side and they are on top of the image

juzen2003:
- Since we don't want "<" & ">" buttons to overlap with detail preview image, now "<" & ">" are positioned at both sides of the image without transition animation. (With transition animation, those buttons will easily overlapped with the image, there is no enough width for animation when screen is narrow). The size of "<" & ">" will also change based on the screen width.

- When detail preview image moves to the top, position "<" & ">" near the image
instead of the edges of the screen. Shrink the size of "..." when screen is getting narrower.

juzen2003:
Show the preview detail image arrow buttons after they are properly positioned.

opus/application/apps/ui/views.py
```diff
@@ -55,7 +55,8 @@
                              HTTP404_BAD_OR_MISSING_REQNO,
                              HTTP404_NO_REQUEST)
 from tools.file_utils import (get_pds_preview_images,
-                              get_pds_products)
+                              get_pds_products,
+                              get_displayed_browse_products)
```
rfrenchseti: get_displayed_browse_products should be first since imports are kept in alphabetical order.

opus/application/apps/tools/file_utils.py
```diff
@@ -237,3 +237,24 @@ def get_pds_preview_images(opus_id_list, preview_jsons, sizes=None,
         image_list.append(data)
 
     return image_list
+
+def get_displayed_browse_products(opus_id, version_name='Current'):
+    """Given an opus_ids, return a list of browse products urls displayed
```
rfrenchseti: "Given an opus_id, return a list of browse product URLs to display in the detail tab."

opus/application/apps/tools/file_utils.py
```diff
@@ -237,3 +237,24 @@ def get_pds_preview_images(opus_id_list, preview_jsons, sizes=None,
...
+       in detail tab.
+    """
+    browse_products = get_pds_products(opus_id_list=opus_id,
```
rfrenchseti: opus_id_list isn't a keyword argument so you should just give it as the first argument without opus_id_list= in front.

opus/application/apps/tools/file_utils.py
```diff
@@ -237,3 +237,24 @@ def get_pds_preview_images(opus_id_list, preview_jsons, sizes=None,
...
+    # When there is no preview image, we return settings.THUMBNAIL_NOT_FOUND
+    if len(selected_browse_products) == 0:
+        return [(settings.THUMBNAIL_NOT_FOUND, settings.THUMBNAIL_NOT_FOUND)]
+    res = []
+    # One opus id could have multiple previews, for example:
+    # co-rss-occ-2008-039-rev058c-x43-i
+    # co-uvis-occ-2005-232-alpsco-i
+    for p in selected_browse_products:
+        for browse_med_url in selected_browse_products[p]:
+            browse_full_url = browse_med_url.replace('_med.', '_full.')
```
rfrenchseti, line 257: This really isn't the right way to do this, because there's no guarantee that preview/browse filenames will follow this format. We should be using the PdsViewable capability somehow, but I don't think it currently supports what we want to do and would require changes to the import pipeline. You should submit a new GitHub issue after this pull is merged saying this eventually needs to be fixed.

opus/application/static_media/css/opus.css
```diff
@@ -2281,12 +2288,14 @@ to the left pane when screen is getting too wide */
...
-    /* Set detail image size when it's above the left pane */
+    /* Set detail image size when it's above the left pane, use !important to override
+    the styling from flexslider */
     .op-detail-img img {
-        height: 150px;
+        height: 150px !important;
     }
 
-    /* Reduce the top space above the detail image when screen */
+    /* Reduce the top space above the detail image when screen, use !important to override
```
rfrenchseti: I'm not sure what this comment was supposed to mean originally - "when screen" doesn't make any sense. Can you fix it?

opus/application/static_media/js/opus.js
```diff
@@ -767,6 +767,7 @@ var opus = {
         let adjustTableSizeDB = _.debounce(o_browse.adjustTableSize, 200);
         let adjustProductInfoHeightDB = _.debounce(o_cart.adjustProductInfoHeight, 200);
         let adjustDetailHeightDB = _.debounce(o_detail.adjustDetailHeight, 200);
+        let adjustDetailImgNavBtnsDB = function() {o_detail.adjustDetailImgNavBtns();};
```
rfrenchseti: Isn't this just:
let adjustDetailImgNavBtnsDB = o_detail.adjustDetailImgNavBtns();

rfrenchseti (changes requested):

rfrenchseti (approved):

opus/application/apps/ui/templates/ui/base.html
```diff
@@ -638,6 +638,9 @@ <h3 class='op-header-text'></h3>
     <script src="https://code.jquery.com/jquery-3.3.1.min.js" integrity="sha256-FgpCb/KJQlLNfOu91ta32o/NMZxltwRo8QtmkMRdAu8=" crossorigin="anonymous"></script>
     <script>window.jQuery || document.write('<script src="{{ STATIC_URL }}cdn_fallback/jquery/js/jquery-3.3.1.min.js"><\/script>')</script>
 
+    <!-- flexslider -->
+    <script src="{{ STATIC_URL }}flexslider/jquery.flexslider.js"></script>
```
dstopp: why did we decide not to make use of a CDN copy of flexslider?

opus/application/static_media/js/detail.js
```diff
@@ -208,4 +217,19 @@ var o_detail = {
         }
     },
 
+    // Adjust the positions of detail preview image nav buttons "<" & ">" to make them
+    // stay right next to the image at both sides.
+    adjustDetailImgNavBtns: function() {
+            let detailPreviewImgWdith = $(".op-detail-img img").width();
+            let offset = 40;
```
dstopp, line 229: why is offset 40 (40... pixels...?)

dstopp (changes requested):

opus/application/static_media/js/detail.js
juzen2003: By default when css left & right of "<" & ">" are 0, those two buttons stay within the image container with their arrow tips touching the left & right borders of the container (see the attached pic). 
The width of "<" & ">" is bit more than 25px.
To move buttons out of the image, we need to set css left & right to at least -(26px + some values) so that buttons won't be too close the image.
Therefore we pick 40 here.
Screen Shot 2022-07-18 at 3 10 33 PM

opus/application/apps/ui/templates/ui/base.html
juzen2003: Updated to use CDN for flexslider now. Local files have been moved to cdn_fallback folder.

opus/application/static_media/js/detail.js
dstopp: I think a brief comment would be great regarding your reasoning.

dstopp (approved):

opus/application/static_media/js/detail.js
rfrenchseti: It seems like Dave already added these comments a couple of days ago

rfrenchseti (approved):

Issue #1076, opened 2020-07-02
Implement file versions in OPUS API


rfrenchseti:
Currently product_types gives information on all product_type / version combinations. But there is no way to select any version other than "current" for actual download. This should be added to the API.

rfrenchseti:
This goes along with #289

Issue #289, opened 2018-03-14
Allow selection of production version in cart download


rfrenchseti:
For some datasets there are multiple versions of the products available. We should make it possible to download the old versions as part of the shopping cart process.

The OPUS search criteria will always be based on the most recent version.

See
    pdsf.associated_parallel(..., rank=something)
to get access to the different 'version ranks' of a file or volume.

A version rank is an integer that encodes the version number or status of a volume or product. The latest version is 999999. In lien resolution translates to 990400. For other than the latest version, the rank is an encoded form of the version number, e.g., _v2.1 = 201000.

Each PdsFile object has a property version_ranks that returns the sorted list of all versions of this volume or file.

rfrenchseti:
See also issue #288, which also requires use of ViewMaster routines.

rfrenchseti:
Prior version  downloads are now available on the Detail page.

rfrenchseti:
Idea from Mark 1/25/2022: Instead of specifying versions by number, allow a download of the version that was present "on a specific date" and separately anything that has changed "after a specific date".

PR #1243, opened 2022-07-30
Implement files versions in Cart Download

juzen2003:
- Work in progress related to #289 
- Update the results of product_types.json by appending @version to the "product_type" value.
- Update files api. If there is no types parameter specified, all files will be returned. If types parameter is specified, only the specified types & versions will be displayed.
- Pending item:
    - Update download api
    - Add/Fix tests


opus/application/apps/results/views.py
```diff
@@ -1155,11 +1155,13 @@ def api_get_product_types_for_opus_id(request, opus_id):
     cursor.execute(sql, values)
 
     results = cursor.fetchall()
-    product_types = [{'category': x[0],
-                      'product_type': x[1],
-                      'description': x[2],
-                      'version_number': x[3],
-                      'version_name': x[4]} for x in results]
+    for x in results:
+        prod_type = f'{x[1]}@v{x[4]}' if x[4] != 'Current' else f'{x[1]}@{x[4].lower()}'
```
rfrenchseti: You could still do this using the list comprehension above. Also I think it would make a little more sense to say:
`f'{x[1]}@current' if x[4] == 'Current' else f'{x[1]}@v{x[4]}`

opus/application/apps/results/views.py
```diff
@@ -1233,11 +1235,14 @@ def api_get_product_types_for_search(request):
...
 
     results = cursor.fetchall()
-    product_types = [{'category': x[0],
-                      'product_type': x[1],
-                      'description': x[2],
-                      'version_number': x[3],
-                      'version_name': x[4]} for x in results]
+    product_types = []
+    for x in results:
+        prod_type = f'{x[1]}@v{x[4]}' if x[4] != 'Current' else f'{x[1]}@{x[4].lower()}'
```
rfrenchseti: Same as above

opus/application/apps/tools/file_utils.py
```diff
@@ -85,10 +85,40 @@ def get_pds_products(opus_id_list,
         sql += ', '+q('obs_files')+'.'+q('checksum')
 
     sql += ' FROM '+q('obs_files')
-    sql += ' WHERE '
+    sql += ' WHERE ('
     if product_types != ['all']:
-        sql += q('obs_files')+'.'+q('short_name')+' IN %s AND '
-        values.append(product_types)
+        # Because we didn't store @version to the database, when multiple versions (or
```
rfrenchseti: in the database
product types
are passed in
query the database
in the product_types list
. The query condition will be

opus/application/apps/tools/file_utils.py
```diff
@@ -85,10 +85,40 @@ def get_pds_products(opus_id_list,
...
+        for i, p in enumerate(product_types):
+            # Uncomment this code if we would like to display all files when the user
+            # passes empty string to types parameter. Otherwise we assume the user is
+            # specifying empty string for the types.
+            # if not p:
+            #     continue
+
+            # Check and see if the product type has a version specified (look for '@')
+            if '@' in p:
+                idx = p.index('@')
```
rfrenchseti: Cleaner way to do this:
`prod_type, _, version = p.partition('@')`

opus/application/apps/tools/file_utils.py
```diff
@@ -85,10 +85,40 @@ def get_pds_products(opus_id_list,
...
+                prod_type = p[:idx]
+                version = p[idx+1:]
+                if version == 'current':
+                    version = 'Current'
+                else:
```
rfrenchseti: Ideally we would check for "v" here instead of just assuming it

opus/application/apps/tools/file_utils.py
```diff
@@ -125,6 +155,10 @@ def get_pds_products(opus_id_list,
 
         if version_name not in results[opus_id]:
             results[opus_id][version_name] = OrderedDict()
+        if version_name != 'Current':
+            short_name = f'{short_name}@v{version_name}'
+        else:
+            short_name = f'{short_name}@{version_name.lower()}'
```
rfrenchseti: Again cleaner to say:
```
if version_name == 'Current'
    short_name = f'{short_name}@current'
else:
    ...
```

juzen2003:
- Fixes #1076 #289 
  - If YES:
    - Database used: re-import to make the default select unchecked for non "Current" versions of files.
    - All Django tests pass: Y
  - If YES:
  - About
  - Getting Started
  - FAQ
  - API Guide
  - Tooltips

Description of changes:
- New download behaviors:
    - If no types parameter specified, download all files of the "Current" version
    - If a product type is passed in without "@" + version, download the "Current" version of the specified product type
    - If a product type is passed in with "@" + version, download files of the specified version & product type
    - If types=all or type=, download all files of all versions.
    - By default, non "Current" versions of files are not selected in download pane.
    - Add non "Current" versions of files in download pane (UI). 
- For testing purpose, the files for these two opus ids are uploaded to Dropbox:
    - co-iss-n1460973661
    - co-iss-n1462840881

Known problems:
None

rfrenchseti (changes requested):
- Version names are different on the Detail tab and the Cart tab. For example, Cassini ISS says "Version: 1" on the Detail tab, but "Cassini ISS-Specific Products V1.0" on the Cart tab.
- Non-current version products are still selected by default on the Cart tab.
- Is there a way to make the version headings on the Cart tab less emphasized? So they're more of a sub-category under the main "Cassini ISS-Specific" heading? This would also serve to group them visually.
- You don't actually need to add files to DropBox to test this, since the shelve files contain all the filenames. Creating a zip file without files present will just create files with zero length.

juzen2003:
Update:
  - Add "Show previous versions" checkbox to show/hide the non-current versions of the files.
  - When select all button is clicked, the non-current versions of the files won't
  be selected.
  - When deselect all button is clicked, the non-current versions of the files will
  be de-selected.

opus/application/settings.py
```diff
@@ -379,3 +379,6 @@
 
 # We don't want to have these characters in HTML class or ID for customized tooltips.
 INVALID_CLASS_CHAR = '~!@$%^&*()+=,./;:"?><[]\{}|`# '
+
+# The symbol is appened after product_type to specify the files version
+FILE_VERSION_MODIFIER = '@'
```
rfrenchseti: This doesn't really seem like worth including in settings if we're making it a permanent part of the API. We don't do this with anything else in the API.

opus/application/apps/ui/views.py
```diff
@@ -636,7 +636,9 @@ def api_init_detail_page(request, **kwargs):
         products = {}
     new_products = OrderedDict()
     for version in products:
-        new_products[version] = OrderedDict()
+        # Display "Version: 1.0" instead of "Version: 1" in detail tab
+        pretty_ver = version if version == 'Current' else f'{float(version)}'
```
rfrenchseti: Instead of using the f'' notation it's more clear just to say str(float(version)) unless you're planning to actually use the f'' formatting in some way.

opus/application/apps/tools/file_utils.py
```diff
@@ -87,8 +90,45 @@ def get_pds_products(opus_id_list,
...
+
+            # Check and see if the product type has a version specified (look for '@')
+            if settings.FILE_VERSION_MODIFIER in p:
+                prod_type, _, version = p.partition(settings.FILE_VERSION_MODIFIER)
+                if version == 'current':
+                    version = 'Current'
+                else:
+                    # Support cases like @v1.0, @v1, @v1.2, @1.0, @1, and @1.2
+                    if 'v' in version:
+                        version = version[1:]
```
rfrenchseti: This doesn't necessarily work, since the v could be somewhere else but then we always strip off the first character. How about
version = version.lstrip('v')

opus/application/apps/tools/file_utils.py
```diff
@@ -87,8 +90,45 @@ def get_pds_products(opus_id_list,
...
+        # Because we didn't store @version in the database, when multiple versions (or
+        # product types) are passed in, we need to query the database by both short name
+        # and version name for each product type in the product_types list. The query
+        # condition will be something like:
+        # (obs_files.short_name='coiss_calib' AND obs_files.version_name='Current') OR
+        # (obs_files.short_name='coiss_calib' AND obs_files.version_name='1') OR
+        # (obs_files.short_name='coiss_raw' AND obs_files.version_name='Current') OR
+        # ...
+        for i, p in enumerate(product_types):
+            # Uncomment this code if we would like to display all files when the user
```
rfrenchseti: Is this needed?

opus/application/apps/tools/file_utils.py
```diff
@@ -87,8 +90,45 @@ def get_pds_products(opus_id_list,
...
+                prod_type, _, version = p.partition(settings.FILE_VERSION_MODIFIER)
+                if version == 'current':
+                    version = 'Current'
+                else:
+                    # Support cases like @v1.0, @v1, @v1.2, @1.0, @1, and @1.2
+                    if 'v' in version:
+                        version = version[1:]
+                    float_ver = float(version)
+                    if float_ver.is_integer():
+                        version = f'{int(float_ver)}'
```
rfrenchseti: I don't understand this. Why are we converting version to a float then back to an integer string? Isn't that just leaving it the way it started?

opus/application/apps/tools/file_utils.py
```diff
@@ -87,8 +90,45 @@ def get_pds_products(opus_id_list,
...
+                sql += '('+q('obs_files')+'.'+q('short_name')+'=%s AND '
+                values.append(prod_type)
+                sql += q('obs_files')+'.'+q('version_name')+'=%s)'
+                values.append(version)
+            else:
+                # When there is no modifier "@" in types, we will display "Current"
+                # version of the files, this will match the behavior of api/download
```
rfrenchseti: . This

opus/application/apps/help/api_guide.md
```diff
@@ -606,10 +606,48 @@ Supported return formats: `zip`.
...
+    %EXTLINK%%HOST%/opus/api/download/co-iss-n1460973661.zip?types=coiss_calib@current,coiss_calib@v2.0%ENDEXTLINK%
+
+    Return value is a zip archive containing the files:
+
+%CODE%
+calibrated/COISS_2xxx/COISS_2002/data/1460960653_1461048959/N1460973661_1_CALIB.IMG
+calibrated/COISS_2xxx/COISS_2002/data/1460960653_1461048959/N1460973661_1_CALIB.LBL
+calibrated/COISS_2xxx/COISS_2002/data/1462783195_1462915477/N1462840881_1_CALIB.IMG
+calibrated/COISS_2xxx/COISS_2002/data/1462783195_1462915477/N1462840881_1_CALIB.LBL
+calibrated/COISS_2xxx_v1/COISS_2002/data/1460960653_1461048959/N1460973661_1_CALIB.IMG
```
rfrenchseti: Why are the V1 calibrated files here if we only asked for current and v2?

opus/application/apps/cart/views.py
```diff
@@ -120,13 +120,21 @@ def api_view_cart(request):
...
-    for name, details in info['product_cat_list']:
-        for type in details:
-            if (type['slug_name'] in not_selected_product_types or
-                not type['default_checked']):
-                type['selected'] = ''
-            else:
-                type['selected'] = 'checked'
+    for name, product_versions in info['product_cat_dict'].items():
+        for ver, types in product_versions.items():
+            for type in types:
```
rfrenchseti: `type` is the name of a Python built-in function and should not be used as the name of a variable.

opus/application/apps/cart/views.py
```diff
@@ -120,13 +120,21 @@ def api_view_cart(request):
...
+                if (type['slug_name'] in not_selected_product_types or
+                    not type['default_checked']):
+                    type['selected'] = ''
+                else:
+                    type['selected'] = 'checked'
+    # for name, details in info['product_cat_dict'].items():
```
rfrenchseti: Remove comments?

opus/application/apps/cart/views.py
```diff
@@ -167,7 +175,7 @@ def api_cart_status(request):
             'total_download_count':       Total number of unique files
             'total_download_size':        Total size of unique files (bytes)
             'total_download_size_pretty': Total size of unique files (pretty format)
-            'product_cat_list':           List of categories and info:
+            'product_cat_dict':           List of categories and info:
```
rfrenchseti: The example seems to be wrong now, since it's not showing a dict

opus/application/apps/cart/views.py
```diff
@@ -432,7 +440,7 @@ def api_reset_session(request):
...
                 [
```
rfrenchseti: Same problem with example

opus/application/apps/cart/views.py
```diff
@@ -540,8 +548,12 @@ def api_create_download(request, opus_id=None, fmt=None):
     if product_types is None or product_types == '':
         product_types = []
     else:
-        product_types = product_types.split(',')
-
+        product_types = product_types.lower().split(',')
+    # By default, we want to download all files of the "Current" version if types
+    # parameter is not specified.
+    downloadCurrentOnly = False
```
rfrenchseti: Don't use mixed-cap variable names in Python

opus/application/apps/cart/views.py
```diff
@@ -540,8 +548,12 @@ def api_create_download(request, opus_id=None, fmt=None):
...
+    if product_types == []:
+        downloadCurrentOnly = True
```
rfrenchseti: Easier to write `download_current_only = (product_types == [])`

opus/application/apps/cart/views.py
```diff
@@ -673,62 +685,68 @@ def api_create_download(request, opus_id=None, fmt=None):
...
-                    files_info[pretty_name].append(logical_path)
+        for version_name in files[f_opus_id]:
+            if downloadCurrentOnly and version_name != 'Current':
+                continue
+            files_version = files[f_opus_id][version_name]
+            for product_type in files_version:
+                for file_data in files_version[product_type]:
+                    path = file_data['path']
+                    pretty_name = path.split('/')[-1]
+                    logical_path = path[path.index('/holdings')+9:]
```
rfrenchseti: I'm not thrilled with hardcoding `/holdings` in the OPUS source code. That's really something that is the domain of PdsFile and OPUS shouldn't know about the directory structure. If we have to do this, let's at least put this string in settings.py.

opus/application/apps/cart/views.py
```diff
@@ -781,22 +799,24 @@ def _get_download_info(product_types, session_id):
         'total_download_count':       Total number of unique files
         'total_download_size':        Total size of unique files (bytes)
         'total_download_size_pretty': Total size of unique files (pretty format)
-        'product_cat_list':           List of categories and info:
-            [
+        'product_cat_dict':           Dict of categories and info:
+            {
              [<Product Type Category>,
```
rfrenchseti: There should be a key here...

juzen2003:
- Make sure the non-current version of the files of a category can be selected when the "v" button of the category is clicked.
- Make sure the results in cart download pane is sorted from the current to the oldest version.

opus/application/apps/cart/views.py
```diff
@@ -812,27 +832,28 @@ def _get_download_info(product_types, session_id):
...
     sql += 'FROM '+q('obs_files')+' '
     sql += 'INNER JOIN '+q('cart')+' ON '
     sql += q('cart')+'.'+q('obs_general_id')+'='
     sql += q('obs_files')+'.'+q('obs_general_id')+' '
     sql += 'WHERE '+q('cart')+'.'+q('session_id')+'=%s '
     values.append(session_id)
-    sql += 'AND '+q('obs_files')+'.'+q('version_number')+' >= 900000 '
-    sql += 'ORDER BY '+q('sort')
+    # Put "Current" version on top of others
+    sql += 'ORDER BY '+q('sort')+', FIELD('+q('ver')+' ,"Current") DESC'
```
rfrenchseti: Can you just say `ver == 'Current'` (and let it sort on True/False) instead of using the `FIELD` function, which is probably slower?

opus/application/apps/cart/views.py
```diff
@@ -965,25 +1011,49 @@ def _get_download_info(product_types, session_id):
...
+        # info
+        is_adding_up_to_total = False
+        for p in product_types:
+            if settings.FILE_VERSION_MODIFIER in p:
+                prod_type, _, p_version = p.partition(settings.FILE_VERSION_MODIFIER)
+                if p_version.lower() == 'current':
+                    p_version = 'current'
+                else:
+                    # Support cases like @v1.0, @v1, @v1.2, @1.0, @1, and @1.2
+                    if 'v' in p_version:
```
rfrenchseti: p_version = p_version.lstrip('v')

opus/application/apps/cart/views.py
```diff
@@ -965,25 +1011,49 @@ def _get_download_info(product_types, session_id):
...
+                        p_version = p_version[1:]
+                    float_ver = float(p_version)
+                    if float_ver.is_integer():
+                        p_version = f'{int(float_ver)}'
```
rfrenchseti: str(int(float_ver))

rfrenchseti (changes requested):
Some initial comments

juzen2003:
- Update the code for types parameter
    - Only accept cases like @1, @1.0, @1.2, or @current
    - If no version specified, and "@" is not seen in the types, will return the current version
    - If we passed in something like "coiss_calib@", it means the version is an empty string for that specific product type, will return nothing for that product type.
    - If "types=" is passed in, we treat it as an empty string, will return nothing
- Treat "1" and "1.0" differently:
    - In detail tab:
        - If version is 1 (store "1" in db), Version: 1 is displayed, ex: co-iss-n1462840881
        - If version is 1.0 (store "1.0" in db), Version: 1.0 is displayed. ex: hst-13667-wfc3-icok01h9q
    - In cart download pane:
        - If version is 1 (store "1" in db), Version: 1 is displayed, ex: co-iss-n1462840881
        - If version is 1.0 (store "1.0" in db), Version: 1.0 is displayed. ex: hst-13667-wfc3-icok01h9q

opus/application/apps/cart/views.py
```diff
@@ -673,62 +682,68 @@ def api_create_download(request, opus_id=None, fmt=None):
...
-                    files_info[pretty_name].append(logical_path)
+        for version_name in files[f_opus_id]:
+            if download_current_only and version_name != 'Current':
+                continue
+            files_version = files[f_opus_id][version_name]
+            for product_type in files_version:
+                for file_data in files_version[product_type]:
+                    path = file_data['path']
+                    pretty_name = path.split('/')[-1]
+                    logical_path = path[path.index(settings.PDS_HOLDINGS_DIR)+9:]
```
rfrenchseti: The 9 should be len(settings.PDS_HOLDINGS_DIR)

opus/application/apps/cart/views.py
```diff
@@ -673,62 +682,68 @@ def api_create_download(request, opus_id=None, fmt=None):
...
+                    path = file_data['path']
+                    url = file_data['url']
+                    category = file_data['category']
+                    product_type = file_data['full_name']
+                    product_abbrev = file_data['short_name']
+                    version_name = file_data['version_name']
+                    checksum = file_data['checksum']
+                    size = file_data['size']
+                    pretty_name = path.split('/')[-1]
+                    logical_path = path[path.index(settings.PDS_HOLDINGS_DIR)+9:]
```
rfrenchseti: The 9 should be len(settings.PDS_HOLDINGS_DIR)

opus/application/apps/cart/views.py
```diff
@@ -845,11 +864,19 @@ def _get_download_info(product_types, session_id):
...
+
         key = (category, pretty_name)
         if key not in product_cats:
             product_cats.append(key)
             cur_product_list = []
-            product_cat_list.append((pretty_name, cur_product_list))
+            product_cat_dict[pretty_name] = {}
+            product_cat_dict[pretty_name][ver] = cur_product_list
+        else:
+            try:
```
rfrenchseti: This is not a good use of try/except because the KeyError could be from either `[pretty_name]` or [`ver`]. Better to just check using `in` explicitly.

opus/application/apps/help/api_guide.md
```diff
@@ -877,6 +885,45 @@ Examples:
 }
 %ENDCODE%
 
+* Retrieve raw images ("Current" version) and calibrated images (version 1.0) for a Cassini ISS observation in JSON format.
```
rfrenchseti: This should say "version 1"

opus/application/apps/help/api_guide.md
```diff
@@ -836,7 +852,7 @@ Examples:
 }    
 %ENDCODE%
 
-* Retrieve raw images only for a Galileo SSI observation in JSON format.
+* Retrieve raw images ("Current" version) only for a Galileo SSI observation in JSON format.
```
rfrenchseti, line 855: This is pointing out that we are losing functionality. There is no longer a way to ask for "all" versions of a product type. It used to be when you specified a product type, you got all the versions, but now you only get the "current" version. Perhaps it would be worth adding an "@all" feature to the `files` API?

rfrenchseti (changes requested):
- "Select All Product Types" disables all of the checkmarks at the top of sections, even for the sub-versions. But in those cases the product types weren't actually selected so it's the wrong thing to do.
- Changing any of the "current" version checkboxes causes the total size and number of files  at the top to go to zero. Only selecting a sub-version gives a good total size and number of files.

opus/application/static_media/js/cart.js
```diff
@@ -253,16 +260,31 @@ var o_cart = {
          * Update checkboxes and disable/enable select all & deselect all
          * buttons correspondingly.
          */
-        if (($(target).hasClass("op-cart-select-btn") ||
-             $(target).hasClass("op-cart-select-all-btn"))) {
+        if ($(target).hasClass("op-cart-select-all-btn")) {
+            // When "Select all product types" is clicked, only select the current version
+            // of the files
+            for(let input of checkboxesOptions) {
```
rfrenchseti: Add space after `for`

opus/application/static_media/css/opus.css
```diff
@@ -1485,6 +1485,17 @@ panel */
     max-width: 36em;
 }
 
+.op-cart-ver-sub-header {
+    /* font-size: 0.8rem; */
```
rfrenchseti: Remove comment?

opus/application/test_api/test_cart_api.py
```diff
@@ -311,7 +311,7 @@ def test__api_cart_add_download_one(self):
...
...ltip': 'Medium-size (often 512x512) non-linearly stretched preview JPEGs (*_med.jpg) of observations created by the Ring-Moon Systems Node. Previews of images are colored according to the filter used. Previews from non-imaging instruments attempt to represent the contents of observations in a visual way. Previews are not appropriate for scientific use.', 'product_type': 'Browse Image (medium)', 'product_count': 1, 'download_count': 1, 'download_size': 7892, 'download_size_pretty': '7K', 'default_checked': 0, 'product_type_with_version': 'browse_medium@Current'}, {'slug_name': 'browse_full', 'tooltip': 'Full-size non-linearly stretched preview JPEGs (*_full.jpg) of observations created by the Ring-Moon Systems Node. Previews of images are not colored. Previews from non-imaging instruments attempt to represent the contents of observations in a visual way. Previews are not appropriate for scientific use.', 'product_type': 'Browse Image (full)', 'product_count': 1, 'download_count': 1, 'download_size': 215079, 'download_size_pretty': '210K', 'default_checked': 1, 'product_type_with_version': 'browse_full@Current'}]}}, 'error': False, 'count': 1, 'recycled_count': 0, 'reqno': 101010101}
```
rfrenchseti, line 338: I just noticed there's an entire class of tests missing that you could add now. All of the cart operators like add, remove, and reset take an optional types= parameter when download=1 is specified, but we never test this. This would be a good time to add a few tests that add that parameter and also check versions at the same time.

opus/application/apps/cart/views.py
```diff
@@ -965,25 +1007,41 @@ def _get_download_info(product_types, session_id):
...
+                prod_type, _, p_version = p.partition('@')
+                if p_version.lower() == 'current':
+                    p_version = 'current'
+                if short_name == prod_type and version_name.lower() == p_version:
+                    is_adding_up_to_total = True
+                    break
+            elif short_name == p:
+                is_adding_up_to_total = True
+                break
+        if (product_types == ['all'] and checked) or is_adding_up_to_total:
```
rfrenchseti, line 1032: This doesn't seem right to me. _get_download_info should not be checking whether an item is checked by default. "all" should really mean all. Perhaps we need a function flag that says something like "count_default_checked_only"? Or is there a way to pass in the list of default-checked product types?

rfrenchseti (changes requested):

opus/application/apps/cart/views.py
rfrenchseti: I thought about this some more. Perhaps this is OK after all, since it's only for internal use. But there needs to be a comment added to the function header saying that "all" doesn't really mean ALL.

rfrenchseti:
- Show previous versions. Click Select all product types. Now deselect one of the "current" types. The Select all button will become active. Select the "current" type again. The Select all button will stay active.
- Show previous versions. Click Select all product types. The "x" above additional versions will become active even though there are no type selected under that heading.

rfrenchseti:
Also I'm bothered by the button saying "Select all product types". Maybe something like "Select all current version product types"? I think there might be room.

juzen2003:
- Update "Select all current version product types" & "Deselect all product types" behaviors:
    - When all current version inputs (non-current version inputs are not included) are checked, "Select all current version product types" is disabled
    - When all inputs (both current & non-current version inputs are included) are unchecked, "Deselect all product types" is disabled.
- Update "v" & "x" behaviors for each category:
    - When all inputs in a category (not including the non-current version inputs) are checked, "v" of that category will be disabled. Non-current version inputs are under their own category, Version: x.
    - When all inputs in a category (not including the non-current version inputs) are unchecked, "x" of that category will be disabled. Non-current version inputs are under their own category, Version: x.

opus/application/apps/cart/views.py
rfrenchseti: I think what you want to say is:
The resulting totals are limited to the given product_types. ['all'] means include all product types that are "checked by default" in the database.

opus/application/static_media/js/cart.js
```diff
@@ -142,8 +142,10 @@ var o_cart = {
             let isAllCatOptionsChecked = o_cart.isAllOptionStatusTheSame(productInputs);
             let isAllCatOptionsUnchecked = o_cart.isAllOptionStatusTheSame(productInputs, false);
 
-            let allCheckboxesOptions = $(".op-download-options-product-types input");
+            // let allCheckboxesOptions = $(".op-download-options-product-types input");
```
rfrenchseti: Remove comment?

opus/application/static_media/js/cart.js
```diff
@@ -175,14 +179,21 @@ var o_cart = {
 
             o_cart.updateCheckboxes(e.currentTarget, productList);
 
-            let allCheckboxesOptions = $(".op-download-options-product-types input");
+            // let allCheckboxesOptions = $(".op-download-options-product-types input");
```
rfrenchseti: Remove comment?

opus/application/test_api/test_cart_api.py
```diff
@@ -2163,6 +2202,27 @@ def test__api_cart_recyclebin_reset(self):
...
+        self._run_json_equal(url, expected)
+        url = '/__cart/status.json?reqno=456'
+        expected = {'recycled_count': 0, 'count': 1, 'reqno': 456}
+        self._run_json_equal(url, expected)
+        url = '/__cart/reset.json?reqno=42&recyclebin=0'
```
rfrenchseti: Maybe added types= here too?

opus/application/test_api/test_cart_api.py
```diff
@@ -778,7 +802,22 @@ def test__api_cart_remove_one_download(self):
...
         self._run_json_equal(url, expected)
         url = '/__cart/remove.json?opusid=co-iss-n1460961026&reqno=123456&download=1'
-        expected = {'recycled_count': 0, 'count': 0, 'error': False, 'reqno': 123456, "total_download_count": 0, "total_download_size": 0, "total_download_size_pretty": "0B", "product_cat_list": []}
+        expected = {'recycled_count': 0, 'count': 0, 'error': False, 'reqno': 123456, "total_download_count": 0, "total_download_size": 0, "total_download_size_pretty": "0B", "product_cat_dict": {}}
+        self._run_json_equal(url, expected)
+        url = '/__cart/status.json?reqno=789'
+        expected = {'recycled_count': 0, 'count': 0, 'reqno': 789}
+        self._run_json_equal(url, expected)
+
+    def test__api_cart_remove_download_ver(self):
```
rfrenchseti, line 811: This is sort of a boring test, since the cart is empty after the remove, so the list of product types doesn't matter. Is there another case you can do where something is left in the cart after the remove?

rfrenchseti (changes requested):

rfrenchseti (approved):

PR #1249, opened 2022-08-26
Support multiple values in mult fields

### LINKED ISSUE #1154: PDS4: Support multiple-valued mult properties associated with a single OPUS ID
- There can be more than one "filespec" for an OPUS ID 
- PDS4 allows multiple values for "intended target name"

rfrenchseti:
- Fixes #1154 
  - If YES:
    - Database used: NEW IMPORT
    - All Django tests pass: Y
  - If YES:
    - JSHINT run on all affected files: N
  - API Guide

Description of changes:

Major changes to the handling of "mult" database fields. Previously a mult obs_* entry, such as Intended Target Name, would contain a string which was the abbreviated version of the displayed value (e.g. "CO" vs "Cassini"). There would also be a hidden mult_* field that contained an id. This id would then be joined with a mult_obs_* table to look up the displayed value (and its sort order, etc). Now mult obs_* entries are integers containing the id, and the hidden mult_* fields have been removed. The value field in the mult_obs_* tables are almost unused at this point, except they are still used to distinguish the NULL database value. At some point we should remove the use of the value field entirely.

A second major change is that some mult obs_* fields can contain lists of values instead of just a single mult id. Currently that is obs_general target_name and target_class. These fields are marked with the "MULTIGROUP" form type. Their contents is a JSON encoding of a list of integers, each of which is an index into a mult_obs_* table. When searching, if the user-requested search value is anywhere in the list, the search is True. When returning metadata, the list is converted to a comma-separate list of display labels by looking up each id one at a time.

A third major change is I took the opportunity to get rid of the ill-conceived /api/metadata_v2 API call. This was implemented originally because the /api/metadata API had a bug in it, but we wanted to maintain backwards compatibility at all costs. As a result we had this weird _v2 version of this single API. However, in the past two years not a single user has accessed either of these APIs, so I feel it is safe to remove the bad /api/metadata and rename /api/metadata_v2 to /api/metadata. The internal version is now /__api/metadata and the JS has been changed accordingly. /api/metadata_v2 is still provided for backwards compatibility, but is undocumented.

A new database needs to be imported for this version to work. As checked in, the new database will have the normal correct data, which means there are no examples of multiple targets. To create a database containing multiple targets, comment out line 28 in obs_instrument_cassini_occ.py. Cassini occultation observations are faked to have three Intended Target Names: Saturn Rings, Enceladus, and Tethys. This causes two Nominal Target Classes, Ring and Regular Satellite.  You can check these results by looking at the Detail tab or adding Intended Target Name or Nominal Target Class to the selected metadata. With this database, nine tests will fail because they expect the correct data for occultation observations. 

Known problems:

The performance of searching on target_name has been degraded by ~30%. The performance of returning metadata with target_name has been degraded by ~10%.

The second part of #1154, allowing text fields to have multiple values, has not been implemented.

At some point we will have a dataset that supports multiple targets and we should add new tests at that point to explicitly test the multi-mult functionality.


opus/application/test_perf/test_perf_target.py
```diff
@@ -0,0 +1,71 @@
...
+            start_time = f'{yr:04d}-{month:02d}-{day:02d}T{hr:02d}:{min:02d}:{sec:02d}.'
+            start_time += f'{iteration:03d}'
+            presearch_params = f'time1={start_time}&'
+        url = f'{HOST}/api/data.json?{presearch_params}{search_params}'
+        url += f'&cols={columns}&limit=10000'
+        # print(url)
+        start_time = time.time()
+        with urllib.request.urlopen(url) as response:
+            html = response.read()
+            # print(html)
```
juzen2003, line 47: Do we want to remove these debugging print statements (same at line 43)?

opus/import/do_import.py
```diff
@@ -1195,140 +1216,160 @@ def import_observation_table(instrument_obj,
...
+                        'has NULL value but NOT NULL is set')
+            else:
+                if field_type.startswith('flag'):
+                    if column_val in [0, 'n', 'N', 'no', 'No', 'NO', 'off', 'OFF']:
+                        if field_type == 'flag_onoff':
+                            column_val = 'Off'
+                        else:
+                            column_val = 'No'
+                    elif column_val in [1, 'y', 'Y', 'yes', 'Yes', 'YES', 'on',
+                                        'ON']:
```
juzen2003: This list can be just one line.

juzen2003 (changes requested):
some warnings from flake 8
```
./run_flake8_application.sh 
application/apps/tools/file_utils.py:10:1: F401 'json' imported but unused
application/apps/tools/db_utils.py:7:1: F401 'json' imported but unused
application/apps/tools/db_utils.py:43:1: E305 expected 2 blank lines after class or function definition, found 1
```

opus/import/do_import.py
rfrenchseti: Fixed

opus/application/test_perf/test_perf_target.py
rfrenchseti: This program is a bit of a hack at the moment, just used for this one purpose. I think leaving them readily available is nice because it's good for future debugging. Also there is a top-level "perf_test" system that this should really be integrated into. I'll make that a separate issue.

juzen2003 (approved):

PR #1254, opened 2022-09-06
Add units to API metadata retrieval

### LINKED ISSUE #1031: Allow selection of units when retrieving metadata in API
I'd really like to be able to display metadata using different units. For example, right now you can enter wavelength in microns of cm, but you can only display it in microns. There should be a way to allow the user to select units for displayed metadata. I was thinking perhaps the Select Metadata dialog should allow selecting units for those fields that support it. You might also be able to select units from the header in table view. How to encode it in the URL is TBD.

rfrenchseti:
- Fixes #1031 
  - If YES:
    - Database used: opus3_test_220906
    - All Django tests pass: Y
  - If YES:
- Was the documentation reviewed for necessary changes or additions? Y/NA
  - About
  - Getting Started
  - FAQ
  - API Guide - YES
  - Tooltips

Description of changes:

Any retrieval of metadata that has a `cols=` parameter now accepts an optional unit-specifier on each slug name like `time1:jd`. One side effect of this is date/time fields now have visible units `(YMDhms)`. Also improved diffs in test output.

Known problems:

None.


opus/application/apps/results/views.py
```diff
@@ -1340,7 +1350,9 @@ def error_return(s, e): return (None, None, None, None, None, None, (s,e))
     mult_tables = set()
     for slug in cols_to_slug_list(cols):
         # First try the full name, which might include a trailing 1 or 2
-        pi = get_param_info_by_slug(slug, 'col')
+        # Allow allow the caller to specify desired units for the retrieved metadata
```
juzen2003: redundant "allow"

juzen2003 (changes requested):
- Database used: opus3_test_220906
- All Django tests pass: Y
- Tested on Chrome, Firefox, Safari, and iOS: NA

juzen2003 (approved):

PR #1256, opened 2022-09-08
Detail view COISS doesn't allow opening of full preview image

### LINKED ISSUE #1255: Detail view of COISS does not allow opening of full-size preview image
The implementation of get_displayed_browse_products doesn't work for COISS. The problem is that the medium preview image has a .jpg extension and the full preview images has a .png extension. Thus you can't do the string replace in this code. The effect is you can't click on a COISS image on the detail page and get the full image. The code will need to be changed to get both the medium and full browse products from get_pds_products.

juzen2003:
- Fixes #1255 
- Were any Django, import pipeline, or table_schema files modified? N
  - If YES:
    - Database used: opus3_test_220906
    - All Django tests pass: Y
  - If YES:
- Was the documentation reviewed for necessary changes or additions? NA
  - About
  - Getting Started
  - FAQ
  - API Guide
  - Tooltips

Description of changes:
Update ```get_displayed_browse_products``` to obtain both displayed (med) and full size browse products. When iterating through the displayed browse products, we will search for the corresponding full size image in full size browse products instead of replacing "_med." with "_full". That way the displayed images can find the corresponding full size image url even if the file extensions are different.

Known problems:

### DIFF
--- a/opus/application/apps/tools/file_utils.py
+++ b/opus/application/apps/tools/file_utils.py
@@ -267,7 +267,12 @@ def get_displayed_browse_products(opus_id, version_name='Current'):
     """
     browse_products = get_pds_products(opus_id,
                                        product_types=settings.DISPLAYED_BROWSE_PRODUCTS)
+    # Get full size browse products
+    browse_products_full = get_pds_products(opus_id,
+                                       product_types=settings.FULL_SIZE_BROWSE_PRODUCTS)
+
     selected_browse_products = browse_products[opus_id].get(version_name, [])
+    selected_browse_products_full = browse_products_full[opus_id].get(version_name, [])
     # When there is no preview image, we return settings.THUMBNAIL_NOT_FOUND
     if len(selected_browse_products) == 0:
         return [(settings.THUMBNAIL_NOT_FOUND, settings.THUMBNAIL_NOT_FOUND)]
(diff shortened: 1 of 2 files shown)


rfrenchseti (changes requested):
Since `get_pds_products` takes a list of product types, isn't there some way to pass in medium and full at the same time and get back the results already paired up? It seems wasteful to have to search through the results to find the matches.

opus/application/apps/tools/file_utils.py
```diff
@@ -275,9 +276,20 @@ def get_displayed_browse_products(opus_id, version_name='Current'):
     # One opus id could have multiple previews, for example:
     # co-rss-occ-2008-039-rev058c-x43-i
     # co-uvis-occ-2005-232-alpsco-i
+    # Get paired medium and full browse urls in res (mediem url, full url)
```
rfrenchseti: mediem -> medium

opus/application/apps/tools/file_utils.py
```diff
@@ -267,6 +267,7 @@ def get_displayed_browse_products(opus_id, version_name='Current'):
     """
```
rfrenchseti, line 268: We should say something in the comment about the return being a list of tuples (medium, full)

opus/application/settings.py
```diff
@@ -334,7 +334,8 @@
 }
 
 # Browse products displayed in OPUS detail tab
-DISPLAYED_BROWSE_PRODUCTS = ['browse_medium','diagram_medium']
+DISPLAYED_BROWSE_PRODUCTS = ['browse_medium','diagram_medium',
```
rfrenchseti: Should have spaces after commas

rfrenchseti (changes requested):

rfrenchseti (approved):

Issue #313, opened 2018-04-23
Multiple form types for a single widget


rfrenchseti:
The "widgets" that allow searching can be expanded to more than just check boxes and range searches. Each data field could actually have multiple ways of search on it (for example, both range and check boxes for Cassini rev_no). In addition:

From Mark:

When we get around to dealing with asteroid and comet targets, which we are sort of committed to dealing with as part of our PDART program, we will have to handle target names in multiple ways too--we can't have a checkbox for every asteroid, but we also don't want to get rid of the checkboxes for planets and moons, so the UI will have to provide both options.

This was actually part of the thinking behind OPUS2--that there could be a variety of UI widgets inside any of those constraint boxes. 

[Old content, see #354, #581]
For example, I would love to see options to select ranges for things like incidence, emission, phase, latitude and longitude using sliders. Wouldn't it be cool if you could see a map of Enceladus and then draw a box around the region you want to search for?]

rfrenchseti:
Phases of implementation:
1) Allow sub-categories in checkbox fields. Allow multiple levels of sub-category, like "Other/Asteroid/A-C/A/Asteroid AAA111".
2) Allow checkbox fields to be hidden. There is already a mult field for this, but I'm not sure it's paid attention to when displaying the widget.
3) Support arbitrary list of aliases for mult values.
3) Add API support for string searching on checkbox fields. This will require a new type of search term in the URL.
4) Add API support string completion on checkbox fields.
5) Add UI support for adding a string-search version of the widget for checkbox fields in addition to the normal widget.

rfrenchseti:
Phase 1, 2, and 3 are finished (although there is no way to search on alias strings).

The details of #4 are yet to be determined.

PR #1269, opened 2022-10-13
313 Multiple form types for a single widget

juzen2003:
- The is the pull request to track issue: #313 
- Implementation done:
    - phase 1: Allow sub-categories in checkbox fields. Allow multiple levels of sub-category.
        - Now ```grouping``` field can store directory tree like main_cat/sub_cat1/sub_cat2  

- Steps to test (phase 1):
1. In config_targets.py, replace the normal PLANET_GROUP_MAPPING to the commented out one and re-import the test database.
2. Intended target name widget will have multiple level of sub categories shown. Each level can have a mix of checkboxes and sub-category.

### DIFF
--- a/opus/application/apps/ui/views.py
+++ b/opus/application/apps/ui/views.py
@@ -261,20 +261,50 @@ def api_get_widget(request, **kwargs):
     """
 
     def _update_form_with_grouping(form, form_vals, glabel, gvalue):
-        form += ("\n\n"
-                 +'<div class="mult_group_label_container'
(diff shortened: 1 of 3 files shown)


rfrenchseti (changes requested):
The green hinting numbers are wrong. I'm not quite sure how they're being calculated now but they don't make sense. The numbers at each level should include ALL of the checkboxes at all levels below it.

rfrenchseti (approved):

PR #1327, opened 2023-11-20
Rename volumeid to bundleid

rfrenchseti:
- Fixes #N/A
- Were any Django, import pipeline, or table_schema files modified? Y
  - If YES:
    - Database used (or new import): New import
    - FLAKE8 run on modified code: Y
    - All Django tests pass: Y
    - Code coverage run and still at 100%: Y
- Were any JavaScript or CSS files modified? Y
  - If YES:
    - JSHINT run on all affected files: Y
    - Tested on Chrome, Firefox, Safari, Edge, and iOS: NA
      (Remember to test all browser sizes)
    - Tested for race conditions (with delayed API calls if applicable): NA
- Was the documentation reviewed for necessary changes or additions? Y
  - About
  - Getting Started
  - FAQ
  - API Guide
  - Tooltips
  - Import/database schema

Description of changes:

This is the first set of changes required to move OPUS to supporting PDS4. It primarily renames `volume` to `bundle` everywhere to agree with the convention that everything should use PDS4 terminology going forward. The actual data imported has not changed. The UI adds new fields for PDS4, which are currently unpopulated, and renames some existing fields to be more general.

- Renamed volumeid to bundleid, volume to bundle everywhere. Note this means the API return values have changed! The previous `volumeid` return slug is now called `bundleid`.
- Kept `volumeid` as a backup slug name for `bundleid` to allow backwards compatibility
- Renamed UI "Volume ID" field to "Bundle/Volume ID"
- Renamed UI "Data Set ID" to "PDS3 Data Set ID"
- Renamed UI "Product ID" to "PDS3 Product ID"
- Added new field "Primary Product LID", which is currently None for all observations
- Update tooltips, help files, and help menu as necessary

Known problems:

- The descriptions of OPUS product types have not changed. Some of these, such as `rms_index` may need to be changed in the future to cover whatever we provide for PDS4 bundles.

PR #1346, opened 2024-03-11
Add PDS4 support and uranus_occs_earthbased

### LINKED ISSUE #353: PDS4: Implement import of PDS4 data products
We will need to convert the import pipeline to import from PDS4 products instead of (or in addition to) PDS3. This includes the ability to do nightly incremental updates.

### LINKED ISSUE #1132: Add Uranus Earth-based occultations to OPUS
When available, we need to add the Uranus occultations (PDS4 data set) to OPUS.

### LINKED ISSUE #1345: Check CORSS_8001 DSN descriptions
The current descriptions for the DSNs used by CORSS_8001 may be wrong. For example, the documentation says that all X-band measurements were made with 70m dishes, but we say they were 34m dishes. All DSNs need to be checked after carefully reading the documentation.

### LINKED ISSUE #1354: Sub-solar fields should be NULL for occultations
Currently we treat sub-solar (elevation, opening angle) to mean sub-photon-source for occultations. However, this is confusing to people who actually expect it to mean sub-SOLAR in all cases. The easiest fix is to set these fields to NULL for occultations, since we already have incidence angle to handle this constraint. The tooltips will need to be updated to remove comments about occultations, and the Profile Constraints tab will have to replace the solar fields with incidence.

rfrenchseti:
- Fixes #353
- Fixes #1132 
- Fixes #1345
- Fixes #1354 
  - If YES:
  - About
  - Getting Started
  - FAQ
  - API Guide
  - Tooltips
  - Import/database schema

Description of changes:

- Add support for PDS4 throughout.
  - Major restructuring of `obs_` file and class hierarchy to have PDS3- and PDS4-specific subclasses. For example, there is now `ObsBase` with methods applicable to both, and subclasses `ObsBasePDS3` and `ObsBasePDS4` with methods specific to the particular PDS version.
  - Add PDS version info to bundle table.
  - Try both PDS3 and PDS4 when converting to/from filespecs and volume/bundle names.
  - Add reader for PDS4 index files without an associated label. Infer column data types from contents.
  - Rename `--override-pds-data-dir` to `--override-pds3-data-dir` and add `--override-pds4-data-dir`.
  - Rename `opus_secrets.py` variable `PDS_DATA_DIR` to `PDS3_DATA_DIR` and add `PDS4_DATA_DIR`.
  - Change the name of the Help menu to "Bundles/Volumes Available..." (this was missed in #1327)
  - Disable "Index" file links on the Detail tab for PDS4 products since they are not yet supported by ViewMaster.
- Add support for `uranus_occs_earthbased` bundleset. Add new stars to target table.
- EBROCC:
  - Change OPUS IDs to conform to new standard
  - Change EBROCC opus types to conform to new standard
- Add new ring geometry "Ascending Node" longitude fields to pair with existing "J2000" fields. For metadata read from `ring_summary` tables, the AN is computed from J2000. For occultations, the J2000 is computed from AN. This fixes a long-standing bug in the way occultations are imported and also clears up longstanding confusion about what ring "longitude" means.
- Change Observed Resolution to be NULL for occultations.
- Rename `obs_volume_vg28xx` files to match volume name standard.
- Standardize names of ground-based telescopes.
- Rename DSN instruments to DSS.
- Fix result count return for `api_get_data` to return the proper value when in "cart" mode.
- Remove deprecated `profile_index` from opus_products (main change in PdsFile).
- Clean up instances of `from X import *`
- Update `import_for_tests.sh` to abort immediately after an import step fails and fix `main_opus_import.py` to actually exit with an error code when an import fails.
- Remove duplicate log messages from PdsFile. This eliminates the interminable list of ViewSet errors when importing NH.

Known problems:

- In apps/ui/views.py, re-enable Index file links when ViewMaster supports it
- Support PDS4 index labels when available.
- Support index files with XPaths when available.
- Turn on shelve support when available in Pds4File.
- Code marked with `TODOPDS4` will need to be addressed eventually but does not cause problems now.
- When `rms-pdsfile` pull 25 is merged, change `opus_setup_environment.sh`.


opus/import/config_targets.py
```diff
@@ -761,167 +761,199 @@
...
+    'UCAC2_22296665':         (270.68744680, -23.64381530), # u1052
+    'UCAC2_22517254':         (261.49121500, -23.29306250), # u28
+    'UCAC2_22563790':         (283.26958000, -23.26173840), # u84
+    'UCAC2_22564036':         (283.39508950, -23.25030480), # u83
+    'UCAC2_22734194':         (255.59000530, -22.80714590), # u25
+    'UCAC2_22735323':         (256.37848620, -22.87389030), # u23
+    'UCAC2_22794421':         (287.39833590, -22.91141370), # u103
+    'UCAC2_22794648':         (287.52555270, -22.89765390), # u102a/b
+    'UCAC2_23016546':         (292.54515500, -22.30772340), # u9539
+    'UCAC2_23509999':         (299.02695120, -21.33800060), # 134
```
juzen2003: `# u134`

opus/import/do_import.py
```diff
@@ -400,7 +398,7 @@ def update_mult_table(table_name, field_name, table_column, val, label, aliases=
 
     if 'mult_options' in table_column:
         import_util.log_nonrepeating_error(
-            f'Attempting to add value "{val}" to preprogrammed mult table '+
+            f'Unable to add value "{val}" to preprogrammed mult table '+
```
juzen2003: We can remove the `+` here to make it consistent with other multilines string interpolation.

opus/import/import_util.py
```diff
@@ -133,40 +137,53 @@ def yield_import_bundle_ids(arguments):
...
+                if (not bundle_pdsfile.is_bundle_dir and
+                    not bundle_pdsfile.is_bundleset_dir):
+                    any_invalid = True
+                    impglobals.LOGGER.log('fatal',
+                        f'Bundle descriptor not a bundle or bundleset: {bundle_desc}')
                 if not bundle_pdsfile.exists:
                     any_invalid = True
                     impglobals.LOGGER.log('fatal',
-                               f'Volume descriptor not found: {bundle_desc}')
+                            f'Bundle descriptor not found: {bundle_desc}')
```
juzen2003: Remove extra tab to make it consistent with line 158

juzen2003 (changes requested):
Looks good to me, just couple minor code style cleanup needed.

Issue #1405, opened 2025-04-06
QR codes differ on Linux and Windows


rfrenchseti:
Recent updates to Python packages and/or Linux programs and/or Windows programs has caused the QR codes for full OPUS search URLs to be slightly different in compression, but identical in graphical representation. This is irrelevant except it's causing the automated tests to fail.

fyellin:
Yikes.  So when you look at the value `zlib.ZLIB_RUNTIME_VERSION` on Linux and Windows, you get the values 1.2.11 (Linux) and 1.2.13 (Windows).  This means that the PNG algorithm is using slightly different compression algorithms, and explains why the Windows string is always a few bytes shorter than the Linux one.

At some point, we're going to have to separately (1) compare the pages with the images removed, and then (2) compare the images by converting them back into bitmaps and seeing the results.  And at some point, Linux and MacOS will certainly be updated to use the newer zlib library.

PR #1407, opened 2025-04-10
Turn off QR test on Windows

rfrenchseti:
- Fixes #N/A
- Is a new database import required? N
  - If YES:
  - About
  - Getting Started
  - FAQ
  - API Guide
  - Tooltips
  - Import/database schema

Description of changes:

- Because Windows seems to be producing slightly different binary for otherwise-identical QR codes (see #1405), the tests fail on Windows. Until we can figure out what's going on, I have disabled the single QR code test when running on Windows.

Known problems:

### DIFF
--- a/opus/application/test_api/test_help_api.py
+++ b/opus/application/test_api/test_help_api.py
@@ -4,6 +4,7 @@ import logging
 import os
 import platform
 import requests
+import unittest
 from unittest import TestCase
 
 from django.core.cache import cache
--- a/opus/application/test_api/test_help_api.py
+++ b/opus/application/test_api/test_help_api.py
@@ -93,9 +93,15 @@ class ApiHelpTests(TestCase, ApiTestHelper):
         url = '/__help/citing.html'
         self._run_status_equal(url, 200)
 
-    def test__api_help_citing_qr(self):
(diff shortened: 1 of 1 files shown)


fyellin:
I don't think I have access to the machines you are using.  Just out of curiosity, could you compare the following versions on the Linux machine and the Windows machine you are using?

```
% python3 --version
Python 3.12.0
fy@Franks-14-Mac-Book-Pro-4 chambers3 % python3
Python 3.12.0 (v3.12.0:0fb18b02c8, Oct  2 2023, 09:45:56) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import zlib
>>> zlib.__version__
'1.0'
>>> import PIL
>>> PIL.__version__
'10.2.0'
>>> 
```

rfrenchseti:
It's not really that easy because we're using the GitHub runner environment under Windows and it keeps its own Python versions. However, the tests automatically do a `pip freeze` before they start running and you can see the result in the action log. I already verified the PyPI versions where the same between Linux and Windows. However, `zlib` is a Python built-in library, so it's not included in the report. It should just track the Python version. When I run it on the system Python 3.12 on Windows I get `1.0`.

PR #1410, opened 2025-04-12
Overkill on Windows BUG

### LINKED ISSUE #1405: QR codes differ on Linux and Windows

fyellin:
- Fixes #1405 
- Is a new database import required? N
  - If YES:

Possible overkill.  If an HTML page has a dynamic PNG file embedded, we pull out the PNG bytes and compare them as images.  Perhaps a bit overkill.

I also changed PIL version directly, in order to get the tests to pass.

### DIFF
--- a/opus/application/test_api/api_test_helper.py
+++ b/opus/application/test_api/api_test_helper.py
@@ -142,8 +142,8 @@ class ApiTestHelper:
         resp = self._clean_string(str(response.content))
         if embedded_dynamic_image:
             # extract the dynamic images and replace with generic tests.
-            expected, expected_images = self._extract_images(expected)
-            resp, resp_images = self._extract_images(resp)
+            expected, expected_images = self.__extract_images(expected)
+            resp, resp_images = self.__extract_images(resp)
         else:
             expected_images = resp_images = []
         print('Got:')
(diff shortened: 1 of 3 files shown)


opus/application/test_api/api_test_helper.py
```diff
@@ -1,12 +1,15 @@
...
 import difflib
 from io import BytesIO
 import json
 import os
 import re
 import tarfile
 import zipfile
+from PIL import Image
 
+import numpy as np
```
rfrenchseti: numpy is not in requirements.txt. It is brought in second-hand from other modules. It needs to be in the primary requirements.

opus/application/test_api/api_test_helper.py
```diff
@@ -137,13 +140,23 @@ def _run_html_equal_file(self, url, exp_file):
...
+            expected, expected_images = self.__extract_images(expected)
+            resp, resp_images = self.__extract_images(resp)
+        else:
+            expected_images = resp_images = []
         print('Got:')
         print(resp)
         print('Expected:')
         print(expected)
         if resp != expected:
             self._print_clean_diffs(resp, expected)
```
rfrenchseti, line 153: This prints the diffs even if the images end up being the same. It seems like we wouldn't want to do that.

opus/application/test_api/api_test_helper.py
```diff
@@ -307,3 +320,34 @@ def _run_archive_file_equal(self, url, expected,
...
+        if image1 == image2:
+            # Shortcut. If both byte strings are the same, they must be identical images.
+            return
+        image1 = Image.open(BytesIO(image1)).convert('RGB')
+        image2 = Image.open(BytesIO(image2)).convert('RGB')
+        # Images must have the same size, and the same RGB bytes.
+        self.assertEqual(image1.size, image2.size)
+        array1, array2 = np.asarray(image1), np.asarray(image2)
+        self.assertTrue(np.array_equal(array1, array2))
+
```
rfrenchseti: Remove excess blank lines

rfrenchseti (changes requested):

opus/application/test_api/api_test_helper.py
fyellin: A couple of questions. 

Some projects have a way of specifying requirements needed only for testing. Does this have that? If yes, I'll add numpy there. If not, I'll get rid of numpy and compare the two images bit by bit. Slower, but fast enough for a test. 

The print statements have always been there. I didn't add them. But you only see the output of the print statements in the log when the test fails. They turned out to be incredibly useful for me.

rfrenchseti: Because OPUS isn't designed as an installable module, there's currently only one set of requirements. In our other projects that you can do "pip install" on, pyproject.toml has the installation requirements and requirements.txt has all the requirements including testing, doc generation, linting, etc.

Regarding the print statements, I meant the diffs when the strings differ. Normally if the strings differ, the test fails. In this case, if the strings differ, the test can still pass. I thought we should maybe only print the differences if the test is going to fail. But it's not a big deal.

fyellin: I will also add.  This print("got").... sequence of code appears 11 times in the file. pytest only shows stdout if the test fails, so I'm assuming it's harmless.

fyellin:
Just double checking that everything is okay with this PR.

rfrenchseti (approved):

PR #1437, opened 2026-04-01
Import for cassini iss fring mosaics rsfrench2025 and b ring

### LINKED ISSUE #1416: Add support for the F Ring bundle
Add import pipeline support for the new F ring bundle.

This will probably include adding a new "F Ring Reprojected Image Constraints" table and possibly creating a "Reprojected Image" main table for reprojected grid parameters.

### LINKED ISSUE #1417: Add support for the MattH B Ring Spokes bundle
Add import pipeline support for the new B Ring Spokes bundle.

This will probably include adding a new "B Ring Reprojected Image Constraints" table and possibly creating a "Reprojected Image" main table for reprojected grid parameters.

juzen2003:
- Fixes #1416
- Fixes #1417 
- Need to work with https://github.com/SETI/rms-pdsfile/pull/57 and https://github.com/SETI/rms-pdsfile/pull/89
- Is a new database import required? Y
  - If YES:
- Was the documentation reviewed for necessary changes or additions? Y/NA
  - About
  - Getting Started
  - FAQ
  - API Guide
  - Tooltips
  - Import/database schema

Description of changes:
- Add `obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py` to define `ObsBundleCassiniIssFringMosaicsRSFrench2025` class for all f-ring bundle fields  
  - Set greater pixel size to 18000, and lesser pixel size to 401
  - Set intensity levels to N/A
  - Set shutter mode, compression type, data conversion type, gain mode, image number, target description to N/A
  - Set the unit of longitudinal resolution to generic_angle
- Add a new entry in bundle info list for f-ring bundle (`config_bundle_info.py`)
- Add new opus product type entries for f-ring bundles (`internal_def_product_types.json`)
- Change the max length of "primary_lid", "category", and "short_name" fields to work with the f-ring bundle. (`obs_files.json`, `obs_pds.json`, and `search/models.py`)
- Add tests results for `co-iss-n1479208692` (f ring), `co-iss-fring-mosaic-iss_00ari_spkmovper001_prime`, and `co-iss-fring-mosaic-iss_183ri_spokemov001_prime`
- Update `api_product_types_empty_search.json` to include newly added opus product types for f-ring bundle.
- Update `test__api_meta_range_endpoints_greaterpixelsize_COISS` (`test_metadata_api.py`)
- Update `test__api_meta_range_endpoints_lesserpixelsize_COISS` (`test_metadata_api.py`)
- Add f-ring bundle to `api_help_bundles.html`
- Update `test__api_meta_mults` to include the f-ring bundles counts under Cassini category (`test_metadata_api.py`)
- Add test results for  `co-iss-n1479210132` (f ring and b ring) and `co-iss-n1481265970` (b ring)

Note:
- Steps to import with shelves files:
  - Copy shelves files form these directories `_infoshelf-*` and `checksums-*` in Dropbox under `Pds4FileTest/pds4-holdings`
  - We are ready to import.

Known problems:

## Summary by CodeRabbit

* **New Features**
  * Added Cassini ISS F‑Ring mosaics and reprojected‑image products, with background‑subtracted variants, documentation, and multi‑resolution browse thumbnails.

* **Chores**
  * Expanded schema/field size limits to support longer product identifiers.
  * Introduced a new "Mosaic" observation subtype and updated observation‑type mappings.
  * Added import bundle/config entries, import script adjustments, and many new/static API test fixtures for the new products.


opus/application/test_api/responses/api_product_types_empty_search.json
```diff
@@ -97,6 +97,27 @@
         "version_number": 999999,
         "version_name": "Current"
     },
+    {
+       "category": "Cassini ISS F Ring Reprojected Image",
```
rfrenchseti: For consistency I think this needs to be "Image**s**".

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_files.json
```diff
@@ -0,0 +1,134 @@
+{
+    "data": {
+        "co-iss-fring-mosaic-iss_00ari_spkmovper001_prime": {
+            "coiss_f_ring_mosaic": [
+                "https://opus.pds-rings.seti.org/pds4-holdings/bundles/cassini_iss_fring_mosaics_rsfrench2025/cassini_iss_fring_mosaics_rsfrench2025/data_mosaic/iss_00ari_spkmovper001_prime/iss_00ari_spkmovper001_prime_mosaic.img",
+                "https://opus.pds-rings.seti.org/pds4-holdings/bundles/cassini_iss_fring_mosaics_rsfrench2025/cassini_iss_fring_mosaics_rsfrench2025/data_mosaic/iss_00ari_spkmovper001_prime/iss_00ari_spkmovper001_prime_mosaic.lblx"
+            ],
+            "coiss_f_ring_mosaic_metadata": [
+                "https://opus.pds-rings.seti.org/pds4-holdings/bundles/cassini_iss_fring_mosaics_rsfrench2025/cassini_iss_fring_mosaics_rsfrench2025/data_mosaic/iss_00ari_spkmovper001_prime/iss_00ari_spkmovper001_prime_mosaic_metadata_params.tab",
```
rfrenchseti, line 10: Need to add the mosaic .lblx file for the metadata

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+    },
+    "PDS Constraints": {
+        "bundleid": "cassini_iss_fring_mosaics_rsfrench2025",
+        "datasetid": null,
+        "productid": null,
+        "primarylid": "urn:nasa:pds:cassini_iss_fring_mosaics_rsfrench2025:data_mosaic:iss_00ari_spkmovper001_prime_mosaic",
+        "productcreationtime": "2025-09-26T00:00:00.000",
+        "primaryfilespec": "cassini_iss_fring_mosaics_rsfrench2025/data_mosaic/iss_00ari_spkmovper001_prime/iss_00ari_spkmovper001_prime_mosaic.lblx",
+        "opusid": "co-iss-fring-mosaic-iss_00ari_spkmovper001_prime",
+        "note": ""
```
rfrenchseti: The note field could be derived from the index file. There's a "note" column with 0 or more semicolon-separated flags. We could create text for each flag, and then append them together to make this text note. See the user guide for examples (the new user guide will have better descriptions). For example: M1 = "Multiple contiguous observations of the same inertial longitude range"

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+    },
+    "Image Constraints": {
+        "duration": "52562",
+        "greaterpixelsize": null,
```
rfrenchseti: Update to 18000x401

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "lesserpixelsize": null,
+        "levels": "4096",
+        "imagetype": "Frame"
```
rfrenchseti: As discussed 4096 doesn't make sense here - set to None

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "occdir": null,
+        "bodyoccflag": null,
+        "occtimesampling": null,
+        "occdataquality": null,
+        "occdepth1": null,
+        "occdepth2": null,
+        "occwlband": null,
+        "occsource": null,
+        "occreceiverhost": null,
+        "RINGGEOringradius1": "138891.78",
```
rfrenchseti, line 60: Looking into this more, these are actually the correct values. Don't need the new "Core Radius" column.

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "RINGGEOringradius2": "140934.463",
+        "RINGGEOsolarringopeningangle1": null,
```
rfrenchseti, line 62: The solar ring opening angle can be computed from the incidence angle. For Saturn it's 90-north_based_incidence

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "RINGGEOsolarringopeningangle2": null,
+        "RINGGEOobserverringopeningangle1": null,
```
rfrenchseti, line 64: The observer ring opening angle can be computed from the emission angle. For Saturn it's 90-north_based_emission

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+    "Ring Geometry Constraints": {
+        "RINGGEOringradius1": "138891.78",
+        "RINGGEOringradius2": "140934.463",
+        "RINGGEOascendinglongitude1": "15.464",
+        "RINGGEOascendinglongitude2": "50.901",
+        "RINGGEOsubsolarringascendinglong1": null,
+        "RINGGEOsubsolarringascendinglong2": null,
+        "RINGGEOsubobserverringascendinglong1": null,
+        "RINGGEOsubobserverringascendinglong2": null,
+        "RINGGEOJ2000longitude1": "146.054",
```
rfrenchseti: This is in the wrong place. The inertial_longitude from the index file is the "ascending node" version, and the J2000 version should be computed from it.

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "RINGGEOrangetoringintercept2": null,
+        "RINGGEOringcenterdistance1": null,
+        "RINGGEOringcenterdistance2": null,
+        "RINGGEOresolution1": "27.02",
+        "RINGGEOresolution2": "48.833",
+        "RINGGEOprojectedradialresolution1": "27.02",
+        "RINGGEOprojectedradialresolution2": "48.833",
+        "RINGGEOprojectedlongitudinalresolutionangle1": null,
+        "RINGGEOprojectedlongitudinalresolutionangle2": null,
+        "RINGGEOprojectedlongitudinalresolution1": "0.04657",
```
rfrenchseti: Should be angle version, not distance version

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "RINGGEOprojectedlongitudinalresolution2": "0.05101",
+        "RINGGEOphase1": "84.199",
+        "RINGGEOphase2": "84.906",
+        "RINGGEOincidence1": "66.603",
+        "RINGGEOincidence2": "66.603",
+        "RINGGEOemission1": "77.196",
+        "RINGGEOemission2": "77.441",
+        "RINGGEOnorthbasedincidence1": null,
```
rfrenchseti: Need to compute north-based incidence based on time of observation (before or after equinox)

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "RINGGEOnorthbasedincidence2": null,
+        "RINGGEOnorthbasedemission1": null,
```
rfrenchseti: Need to compute north-based emission based on time (before or after equinox)

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "CASSINIprimeinst": "ISS",
+        "CASSINIisprime": "Yes",
+        "CASSINIsequenceid": null,
+        "CASSINIspacecraftclockcount1": "1479201490.118",
+        "CASSINIspacecraftclockcount2": "1479254052.118",
+        "CASSINIert1": null,
+        "CASSINIert2": null
+    },
+    "Cassini ISS Constraints": {
+        "COISScamera": "Narrow Angle",
```
rfrenchseti, line 169: How are you detecting the camera? Some mosaics are NAC, some are WAC

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "COISSfilter": "CLEAR",
+        "COISSshuttermode": "NACONLY",
```
rfrenchseti: Same for here - NACONLY vs WACONLY

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_00ari_spkmovper001_prime_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "COISSshutterstate": "Enabled",
+        "COISScompressiontype": "Lossless",
```
rfrenchseti: None of these items are known - compression type, conversion type, gain mode, image number, target desc, imageoberservationtype.

opus/application/test_api/responses/results_co_iss_n1479208692_files.json
```diff
@@ -0,0 +1,206 @@
+{
+    "data": {
+        "co-iss-n1479208692": {
+            "coiss_f_ring_reproj_img": [
+                "https://opus.pds-rings.seti.org/pds4-holdings/bundles/cassini_iss_fring_mosaics_rsfrench2025/cassini_iss_fring_mosaics_rsfrench2025/data_reproj_img/iss_00ari_spkmovper001_prime/1479208692n_reproj_img.img",
+                "https://opus.pds-rings.seti.org/pds4-holdings/bundles/cassini_iss_fring_mosaics_rsfrench2025/cassini_iss_fring_mosaics_rsfrench2025/data_reproj_img/iss_00ari_spkmovper001_prime/1479208692n_reproj_img.lblx"
+            ],
+            "coiss_f_ring_reproj_img_spice_pointing": [
```
rfrenchseti, line 8: This should also include the .lblx file

opus/application/test_api/responses/results_co_iss_n1479208692_files.json
```diff
@@ -0,0 +1,206 @@
...
+                "https://opus.pds-rings.seti.org/pds4-holdings/bundles/cassini_iss_fring_mosaics_rsfrench2025/cassini_iss_fring_mosaics_rsfrench2025/data_reproj_img/iss_00ari_spkmovper001_prime/1479208692n_reproj_suppl.txt"
+            ],
+            "coiss_f_ring_reproj_img_metadata": [
+                "https://opus.pds-rings.seti.org/pds4-holdings/bundles/cassini_iss_fring_mosaics_rsfrench2025/cassini_iss_fring_mosaics_rsfrench2025/data_reproj_img/iss_00ari_spkmovper001_prime/1479208692n_reproj_img_metadata_params.tab"
```
rfrenchseti, line 14: And this should include the .lblx file

opus/application/test_api/responses/results_co_iss_n1479208692_metadata.json
```diff
@@ -0,0 +1,244 @@
...
+        "RINGGEOprojectedlongitudinalresolutionangle2": null,
+        "RINGGEOprojectedlongitudinalresolution1": null,
+        "RINGGEOprojectedlongitudinalresolution2": null,
+        "RINGGEOphase1": "84.618",
+        "RINGGEOphase2": "85.095",
+        "RINGGEOincidence1": "66.598",
+        "RINGGEOincidence2": "66.599",
+        "RINGGEOemission1": "77.021",
+        "RINGGEOemission2": "77.39",
+        "RINGGEOnorthbasedincidence1": "113.401",
```
rfrenchseti, line 170: One way to test your mosaic metadata is to look at one of the ISS reprojected images taken at the same time (and thus included as part of the mosaic). Things like phase, incidence, emission, and north-based versions of the same should be roughly the same between the single image and the mosaic. Also solar opening angle, opening angle to observer, longitude (AN and J2000 versions).

opus/import/table_schemas/internal_def_product_types.json
```diff
@@ -1028,5 +1028,120 @@
         "pi_dict_context": "OPUS_PRODUCT_TYPE",
         "pi_dict_name": "couvis_solar_occ_browse",
         "definition": "Plots of signal, transparency, and normal optical depth for a single occultation; ingress and egress portions are given in separate files. Associated labels (*.xml) describe the contents of the detailed browse images."
+    },
+    {
+        "pi_dict_context": "OPUS_PRODUCT_TYPE",
+        "pi_dict_name": "coiss_f_ring_mosaic",
+        "definition": "Mosaic image files (*.img) and associated labels (*.lblx) for Cassini ISS F-ring mosaics."
```
rfrenchseti: I don't think there should be a dash in "F ring"

opus/import/table_schemas/internal_def_product_types.json
```diff
@@ -1028,5 +1028,120 @@
...
+    },
+    {
+        "pi_dict_context": "OPUS_PRODUCT_TYPE",
+        "pi_dict_name": "coiss_f_ring_mosaic_metadata",
+        "definition": "Mosaic metadata tables (*.tab) for Cassini ISS F-ring mosaics."
```
rfrenchseti: And associated labels

opus/import/table_schemas/internal_def_product_types.json
```diff
@@ -1028,5 +1028,120 @@
...
+    },
+    {
+        "pi_dict_context": "OPUS_PRODUCT_TYPE",
+        "pi_dict_name": "coiss_f_ring_mosaic_bkg_sub",
+        "definition": "Background-subtracted mosaic image files (*.img) and associated labels (*.lblx) for Cassini ISS F-ring mosaics."
+    },
+    {
+        "pi_dict_context": "OPUS_PRODUCT_TYPE",
+        "pi_dict_name": "coiss_f_ring_mosaic_bkg_sub_metadata",
+        "definition": "Background-subtracted mosaic metadata tables (*.tab) for Cassini ISS F-ring mosaics."
```
rfrenchseti: And associated labels

opus/import/table_schemas/internal_def_product_types.json
rfrenchseti: Based on our discussion, also say "Labels also include definitions for mosaic metadata tables, which can be downloaded separately."

opus/import/table_schemas/internal_def_product_types.json
rfrenchseti: Based on our discussion, also say "Labels also include definitions for mosaic image files, which can be downloaded separately."

opus/import/table_schemas/internal_def_product_types.json
```diff
@@ -1028,5 +1028,120 @@
...
+    },
+    {
+        "pi_dict_context": "OPUS_PRODUCT_TYPE",
+        "pi_dict_name": "coiss_f_ring_global_reproj_img_index",
+        "definition": "Global reprojected image index files (global_reproj_img_index.tab and .lblx) for Cassini ISS F-ring reprojected frames."
+    },
+    {
+        "pi_dict_context": "OPUS_PRODUCT_TYPE",
+        "pi_dict_name": "coiss_f_ring_reproj_img",
+        "definition": "Reprojected image files (*.img) and associated labels (*.lblx) for Cassini ISS F-ring reprojected frames."
```
rfrenchseti: frames -> images
Add "Labels also include definitions for reprojected image metadata tables and spice pointing, which can be downloaded separately."
Do the same for the spice pointing and img_metadata items below.

opus/import/table_schemas/obs_ring_geometry.json
```diff
@@ -1033,7 +1033,7 @@
         "pi_disp_order": 250,
         "pi_display": 1,
         "pi_display_results": 1,
-        "pi_form_type": "RANGE%.5f:distance_resolution",
+        "pi_form_type": "RANGE%.5f:generic_angle",
```
rfrenchseti: This is wrong! There are two fields for longitude resolution - one for distance and one for angle. This is the distance one (see the definition above).

opus/import/_import_all_internal.sh
```diff
@@ -19,6 +19,8 @@ python main_opus_import.py --do-all-import --override-db-schema $1 $3 uranus_occ
 if [ $? -ne 0 ]; then exit -1; fi
 python main_opus_import.py --do-all-import --override-db-schema $1 $3 cassini_uvis_solarocc_beckerjarmak2023
 if [ $? -ne 0 ]; then exit -1; fi
+python main_opus_import.py --do-all-import --override-db-schema $1 $3 cassini_iss_fring_mosaics_rsfrench2025
```
rfrenchseti: For right now let's just comment this out. I want to be able to merge these changes in before the bundle is actually on production so we can keep the code base stable, but this will cause import to crash.

opus/import/do_import.py
```diff
@@ -1575,11 +1575,15 @@ def get_opus_products_rows_for_filespec(pds_version, filespec, obs_general_id,
...
                 # shelves/info files exist, if not, we skip the file.
                 if pds_version == 3 and file.shelf_exists_if_expected() is False:
-                    # TODOPDS4 ^^^
-                    import_util.log_nonrepeating_warning(
-                        'Missing corresponding ' +
-                        f'shelves/info for {file.abspath}')
-                    continue
+
+                    # For cross pds4 products, don't skip the import if shelves file
+                    # doesn't exists.
```
rfrenchseti: exists -> exist

opus/import/import_for_tests.sh
```diff
@@ -31,6 +31,7 @@ python main_opus_import.py --do-all-import VG_28xx $1
 python main_opus_import.py --do-all-import COCIRS_0406,COCIRS_1003 $1
 python main_opus_import.py --do-all-import uranus_occ_u0_kao_91cm $1
 python main_opus_import.py --do-all-import cassini_uvis_solarocc_beckerjarmak2023 $1
+python main_opus_import.py --do-all-import cassini_iss_fring_mosaics_rsfrench2025 $1
```
rfrenchseti, line 34: Before final merge we need to put the mosaic shelve files and index in Dropbox so tests pass

opus/import/main_opus_import.py
```diff
@@ -420,7 +420,7 @@ def _new_warning_handler(message, category, filename, lineno, file, line):
         Pds3File.use_shelves_only()
         # Pds4File.use_shelves_only()    # TODOPDS4
     Pds3File.require_shelves(True)
-    # Pds4File.require_shelves(True)   # TODOPDS4
+    Pds4File.require_shelves(True)   # TODOPDS4
```
rfrenchseti: Is this still TODO?

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+################################################################################
+
+import import_util
+import opus_support
+
+from obs_cassini_common_pds4 import ObsCassiniCommonPDS4
+from obs_volume_coiss_12xxx import _COISS_FILTER_WAVELENGTHS
+
+
+class ObsBundleCassiniIssFringMosaicsRSFrench2025(ObsCassiniCommonPDS4):
```
rfrenchseti: This should probably be `ObsBundleCassiniISSFRingMosaicsRSFrench2025`

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+    @property
+    def primary_filespec(self):
+        rel = self._index_col('file_spec')
+        if rel is None:
+            return None
+        return self.bundle + '/' + rel.strip()
+
+    @property
+    def camera(self):
+        return self._index_col('min_image_name')[-1].upper()
```
rfrenchseti: Ah I see how you detect the camera. Good thinking. But we don't know the shutter mode (NACONLY or WACONLY) because it could be BOTSIM. So have to leave shutter mode NULL.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+
+    def primary_filespec_from_index_row(self, row, convert_lbl=False,
+                                        add_phase_from_inst=False):
+        rel = row.get('file_spec')
+        if rel is None:
+            return None
+        return self.bundle + '/' + str(rel).strip()
+
+    def _some_index_col(self, col, idx=None):
```
rfrenchseti: I think you're doing this because this:

```
    def field_obs_mission_cassini_obs_name(self):
        return self._some_index_col('OBSERVATION_ID')
```

is defined in `obs_cassini_common.py`. But putting this hack here is not the right way to fix this. The above definition using `OBSERVATION_ID` is PDS3-specific, and thus should be moved to the `obs_cassini_common_pds3.py` file, with an appropriate version written for PDS4 in `obs_cassini_common_pds4.py`. The other places that reference that column name explicitly, like `_rev_no` and `_prime_inst`, should be modified to call this field function instead, so they always get the PDS3- or PDS4-specific version.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+                self._metadata['index_row'], 'cassini:observation_id', idx=idx)
+            if v is not None and isinstance(v, str):
+                return v.strip()
+            return v
+        return super()._some_index_col(col, idx=idx)
+
+
+    ################################
+    ### OVERRIDE FROM ObsProfilePDS4
+    ### (non-occultation; match ObsProfilePDS3 pattern)
```
rfrenchseti: Remove stupid AI comment

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+    ################################
+
+    def field_obs_profile_occ_type(self):
+        return self._create_mult(None)
+
+    def field_obs_profile_occ_dir(self):
```
rfrenchseti: I think this entire section can be deleted. We don't want to override things that are defined at a higher class hierarchy level with reasonable defaults.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+        fs = self._index_col('file_spec')
+        if fs is not None and 'data_reproj_img' in fs.lower():
+            return self._create_mult('IMG')
+        return self._create_mult('MOS')
+
+    def field_obs_general_time1(self):
+        return self._time_from_index(column='pds:start_date_time')
+
+    def field_obs_general_time2(self):
+        return self._time2_from_some_index(self.field_obs_general_time1(),
```
rfrenchseti: Should be `_time2_from_index`. No need to specify column name for either because it's the default.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+                                           column='pds:stop_date_time')
+
+
+    ################################
+    ### OVERRIDE FROM ObsPdsPDS4 ###
+    ################################
+
+    def field_obs_pds_product_creation_time(self):
```
rfrenchseti: Remove this. I need to change the index file to have `pds:creation_date_time` as it expected in the PDS4 default code.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+        return self._time_from_index(column='product_creation_date')
+
+    def field_obs_pds_note(self):
+        return self._index_col('notes')
```
rfrenchseti: See earlier discussion of notes.

rfrenchseti: It would be nice if one of the chosen test mosaics actually had some notes

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+
+
+    #####################################
+    ### OVERRIDE FROM ObsRingGeometry ###
+    ### (geometry lives in primary index)
```
rfrenchseti: Remove stupid AI comment

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+    #####################################
+
+    def field_obs_ring_geometry_ring_radius1(self):
+        return self._index_col('rings:minimum_ring_radius')
+
+    def field_obs_ring_geometry_ring_radius2(self):
+        return self._index_col('rings:maximum_ring_radius')
+
+    def field_obs_ring_geometry_j2000_longitude1(self):
+        return self._index_col('rings:minimum_inertial_ring_longitude')
```
rfrenchseti: This is ascending node

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+        return self._index_col('rings:maximum_radial_resolution')
+
+    def field_obs_ring_geometry_projected_radial_resolution1(self):
+        return self.field_obs_ring_geometry_resolution1()
+
+    def field_obs_ring_geometry_projected_radial_resolution2(self):
+        return self.field_obs_ring_geometry_resolution2()
+
+    def field_obs_ring_geometry_projected_long_resolution1(self):
+        return self._index_col('rings:minimum_longitudinal_resolution')
```
rfrenchseti, line 238: This is angular

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+
+
+    ######################################
+    ### OVERRIDE FROM ObsCassiniCommon ###
+    ######################################
+
+    def field_obs_mission_cassini_mission_phase_name(self):
+        return self._create_mult(self._cassini_normalize_mission_phase_name())
+
+    def field_obs_mission_cassini_spacecraft_clock_count1(self):
```
rfrenchseti: This deserves to be in `obs_cassini_common_pds4.py`

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+        return 401.
+
+
+    ###################################
+    ### OVERRIDE FROM ObsWavelength ###
+    ###################################
+
+    def field_obs_wavelength_wavelength1(self):
+        filter1, filter2 = 'CL1', 'CL2'
+        central_wl, fwhm, _effective = _COISS_FILTER_WAVELENGTHS[
```
rfrenchseti: We really need to increase the minimum line length for this repo. This is a crazy line break.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+    def field_obs_wavelength_wave_res2(self):
+        return self.field_obs_wavelength_wave_res1()
+
+    def field_obs_wavelength_wave_no_res1(self):
+        return self._wave_no_res_from_full_bandwidth()
+
+    def field_obs_wavelength_wave_no_res2(self):
+        return self.field_obs_wavelength_wave_no_res1()
+
+    def field_obs_wavelength_polarization_type(self):
```
rfrenchseti: Not needed - default

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+        return self._create_mult('NONE')
+
+
+    ##############################################
+    ### FIELD METHODS FOR obs_instrument_coiss #
```
rfrenchseti: Fix ###

rfrenchseti: Fix ###

Now that we have two volumes/bundles for COISS, we should have an `obs_cassini_iss_common.py`, `obs_cassini_iss_common_pds3.py`, and `obs_cassini_iss_common_pds4.py`. Both `obs_volume_coiss_12xxx.py` and this file should use it. The main common file should define all of the fields as None. The PDS3 version should have all of the stuff from `coiss_12xxx` (and delete it from there). The PDS4 version should try to fill in all of the fields using standard PDS4 class names. See any reprojected image label file for a full set of PDS4 Cassini ISS attributes.

Then remove the default items from this section and leave the ones necessary to override, since we don't actually provide any of these items in the mosaic index file.

scripts/server/import_and_deploy/_opus_import_volumes.sh
```diff
@@ -38,6 +38,7 @@ done
 for VOLUME in \
   EBROCC \
   uranus_occs_earthbased \
+  cassini_iss_fring_mosaics_rsfrench2025 \
```
rfrenchseti: Comment out for now

rfrenchseti (changes requested):

opus/application/test_api/responses/api_product_types_empty_search.json
```diff
@@ -1595,6 +1672,90 @@
...
+        "version_number": 999999
+    },
+    {
+        "category": "browse",
+        "description": "Browse Reprojected Image (full)",
+        "product_type": "coiss_f_ring_browse_reproj_img_full",
+        "version_name": "Current",
+        "version_number": 999999
+},
     {
```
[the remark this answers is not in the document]
juzen2003: Fixed

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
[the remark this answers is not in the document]
juzen2003: Leave as-is, _index_col('min_image_name') won't be None

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+        filter1, filter2 = 'CL1', 'CL2'
+        central_wl, fwhm, _effective = _COISS_FILTER_WAVELENGTHS[
+            (self.camera, filter1, filter2)]
+        return (central_wl - fwhm / 2) / 1000.
+
+    def field_obs_wavelength_wavelength2(self):
+        filter1, filter2 = 'CL1', 'CL2'
+        central_wl, fwhm, _effective = _COISS_FILTER_WAVELENGTHS[
+            (self.camera, filter1, filter2)]
+        return (central_wl + fwhm / 2) / 1000.
```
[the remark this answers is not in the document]
juzen2003: Leave as-is

opus/import/main_opus_import.py
```diff
@@ -418,9 +418,9 @@ def _new_warning_handler(message, category, filename, lineno, file, line):
 
     if not impglobals.ARGUMENTS.dont_use_shelves_only:
         Pds3File.use_shelves_only()
-        # Pds4File.use_shelves_only()    # TODOPDS4
+        Pds4File.use_shelves_only()    # TODOPDS4
     Pds3File.require_shelves(True)
-    # Pds4File.require_shelves(True)   # TODOPDS4
+    Pds4File.require_shelves(True)   # TODOPDS4
```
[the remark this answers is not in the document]
juzen2003: Leave as-is, `require_shelves` and `use_shelves_only` set different flags. `require_shelves` sets a flag to raise an error when info shelf or link shelf files are missing. `use_shelves_only` sets a flag to read through shelves files, it uses shelves files rather than refer to the actual holdings directory when checking if a path is a directory or a path exists. They serves different purpose.

juzen2003:
Addressed all the comments.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
juzen2003: We have to either override these in `obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py` or modify ones in `obs_profile_pds4.py` to allow None (like occ_type or occ_dir) instead raising errors. So I kept ones in higher class the same, and override them here to allow None.

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_036rf_fmovie002_vims_files.json
```diff
@@ -0,0 +1,138 @@
...
+                "https://opus.pds-rings.seti.org/pds4-holdings/bundles/cassini_iss_fring_mosaics_rsfrench2025/cassini_iss_fring_mosaics_rsfrench2025/data_mosaic_bkg_sub/iss_036rf_fmovie002_vims/iss_036rf_fmovie002_vims_mosaic_bkg_sub.lblx"
+            ],
+            "coiss_f_ring_mosaic_bkg_sub_metadata": [
+                "https://opus.pds-rings.seti.org/pds4-holdings/bundles/cassini_iss_fring_mosaics_rsfrench2025/cassini_iss_fring_mosaics_rsfrench2025/data_mosaic_bkg_sub/iss_036rf_fmovie002_vims/iss_036rf_fmovie002_vims_mosaic_bkg_sub.lblx",
+                "https://opus.pds-rings.seti.org/pds4-holdings/bundles/cassini_iss_fring_mosaics_rsfrench2025/cassini_iss_fring_mosaics_rsfrench2025/data_mosaic_bkg_sub/iss_036rf_fmovie002_vims/iss_036rf_fmovie002_vims_mosaic_bkg_sub_metadata_params.tab",
+                "https://opus.pds-rings.seti.org/pds4-holdings/bundles/cassini_iss_fring_mosaics_rsfrench2025/cassini_iss_fring_mosaics_rsfrench2025/data_mosaic_bkg_sub/iss_036rf_fmovie002_vims/iss_036rf_fmovie002_vims_mosaic_bkg_sub_metadata_src_imgs.tab"
```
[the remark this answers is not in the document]
juzen2003: This is the correct .lblx for those .tab files, the output is expected. Leave as-is.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,332 @@
...
+
+    ################################
+    ### OVERRIDE FROM ObsPdsPDS4 ###
+    ################################
+
+    def field_obs_pds_note(self):
+        raw = self._index_col('notes')
+        if raw is not None: raw = raw.strip()
+
+        if not raw: return None
```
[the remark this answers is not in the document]
juzen2003: Fixed.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,332 @@
...
+    ### OVERRIDE FROM ObsRingGeometry ###
+    #####################################
+
+    def field_obs_ring_geometry_ring_radius1(self):
+        # minimum_core_radius
+        return self._index_col('rings:minimum_ring_radius')
+
+    def field_obs_ring_geometry_ring_radius2(self):
+        # maximum_core_radius
+        return self._index_col('rings:maximum_ring_radius')
```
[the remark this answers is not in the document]
juzen2003: We will change to use core_radius if they are available in the index file in the future. Leave as-is.

opus/import/obs_cassini_common_pds4.py
```diff
@@ -8,10 +8,114 @@
...
+        return self._create_mult(self._index_col('cassini:instrument_mode_id'))
+
+    def field_obs_instrument_coiss_target_desc(self):
+        target_desc = self._index_col('cassini:pds3_​target_​desc')
+        if target_desc is not None:
+            target_desc = target_desc.upper()
+            coiss_target_desc_mapping = self._coiss_target_desc_mapping()
+            if target_desc in coiss_target_desc_mapping:
+                target_desc = coiss_target_desc_mapping[target_desc]
+        return self._create_mult(target_desc)
```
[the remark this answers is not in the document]
juzen2003: Fixed.

opus/import/obs_volume_coiss_12xxx.py
```diff
@@ -296,32 +131,10 @@ def field_obs_type_image_lesser_pixel_size(self):
...
-        self._log_nonrepeating_warning('Ignoring unknown COISS filter combination '+
-                                      f'{key[0]}/{key[1]}/{key[2]}')
-        return None, None, None
-
     def field_obs_wavelength_wavelength1(self):
         camera = self._index_col('INSTRUMENT_ID')[3]
         filter1, filter2 = self._index_col('FILTER_NAME')
-        central_wl, fwhm, effective_wl = self._coiss_wavelength_helper(
-                                                    camera, filter1, filter2)
+        central_wl, fwhm, effective_wl = self._coiss_wavelength_helper(camera, filter1, filter2)
```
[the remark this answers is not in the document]
juzen2003: Fixed.

scripts/server/import_and_deploy/_opus_import_volumes.sh
```diff
@@ -38,6 +38,7 @@ done
 for VOLUME in \
   EBROCC \
   uranus_occs_earthbased \
+#   cassini_iss_fring_mosaics_rsfrench2025 \
   cassini_uvis_solarocc_beckerjarmak2023 \
```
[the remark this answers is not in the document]
juzen2003: Fixed.

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_036rf_fmovie002_vims_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "rightasc2": null,
+        "declination1": null,
+        "declination2": null
+    },
+    "PDS Constraints": {
+        "bundleid": "cassini_iss_fring_mosaics_rsfrench2025",
+        "datasetid": null,
+        "productid": null,
+        "primarylid": "urn:nasa:pds:cassini_iss_fring_mosaics_rsfrench2025:data_mosaic:iss_036rf_fmovie002_vims_mosaic",
+        "productcreationtime": null,
```
rfrenchseti: There is a product creation time field in the index file

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_036rf_fmovie002_vims_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "primaryfilespec": "cassini_iss_fring_mosaics_rsfrench2025/data_mosaic/iss_036rf_fmovie002_vims/iss_036rf_fmovie002_vims_mosaic.lblx",
+        "opusid": "co-iss-fring-mosaic-iss_036rf_fmovie002_vims",
+        "note": "Some source images have corrupted or missing data. Some areas may be overexposed."
```
rfrenchseti, line 27: Cool I love this note.

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_036rf_fmovie002_vims_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+    },
+    "Image Constraints": {
+        "duration": "48119",
+        "greaterpixelsize": "18000",
+        "lesserpixelsize": "401",
+        "levels": null,
+        "imagetype": "Frame"
```
rfrenchseti: I think "Frame" doesn't really make sense for a Mosaic because it's not a single image. Maybe we need a new imagetype for "Mosaic".

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_036rf_fmovie002_vims_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "RINGGEOrangetoringintercept2": null,
+        "RINGGEOringcenterdistance1": null,
+        "RINGGEOringcenterdistance2": null,
+        "RINGGEOresolution1": null,
+        "RINGGEOresolution2": null,
+        "RINGGEOprojectedradialresolution1": "10.175",
+        "RINGGEOprojectedradialresolution2": "10.558",
+        "RINGGEOprojectedlongitudinalresolutionangle1": "0.00502",
+        "RINGGEOprojectedlongitudinalresolutionangle2": "0.00537",
+        "RINGGEOprojectedlongitudinalresolution1": null,
```
rfrenchseti: Actually we can fill this in. The F ring core is at 140,221.3 km, for a diameter of 280,442.6 km. This gives a circumference of pi*D = 881,036 km. Thus the longitudinal resolution is 881,036 * angular_res / 360, which should be around 12 km.

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_036rf_fmovie002_vims_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "RINGGEOphase2": "135.708",
+        "RINGGEOincidence1": "75.694",
+        "RINGGEOincidence2": "75.694",
+        "RINGGEOemission1": "143.631",
+        "RINGGEOemission2": "145.698",
+        "RINGGEOnorthbasedincidence1": "104.306",
+        "RINGGEOnorthbasedincidence2": "104.306",
+        "RINGGEOnorthbasedemission1": "34.302",
+        "RINGGEOnorthbasedemission2": "36.369",
+        "RINGGEOsolarringelevation1": null,
```
rfrenchseti: solar ring elevation = 90 - north-based incidence (for Saturn)
observer ring elevation = 90 - north-based emission

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_036rf_fmovie002_vims_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "RINGGEOringcenterphase2": null,
+        "RINGGEOringcenterincidence1": null,
+        "RINGGEOringcenterincidence2": null,
+        "RINGGEOringcenteremission1": null,
+        "RINGGEOringcenteremission2": null,
+        "RINGGEOringcenternorthbasedincidence1": null,
+        "RINGGEOringcenternorthbasedincidence2": null,
+        "RINGGEOringcenternorthbasedemission1": null,
+        "RINGGEOringcenternorthbasedemission2": null,
+        "RINGGEOsolarringopeningangle1": "-14.306",
```
rfrenchseti: I think these are supposed to be null and the elevation ones above are supposed to be set with these values.

opus/application/test_api/responses/results_co_iss_fring_mosaic_iss_036rf_fmovie002_vims_metadata.json
```diff
@@ -0,0 +1,183 @@
...
+        "RINGGEOringdiameterpixels": null,
+        "RINGGEOringpolepositionangle": null,
+        "RINGGEOcenterxcoordinate": null,
+        "RINGGEOcenterycoordinate": null,
+        "RINGGEOringintercepttime1": null,
+        "RINGGEOringintercepttime2": null
+    },
+    "Cassini Mission Constraints": {
+        "CASSINIobsname": "ISS_036RF_FMOVIE002_VIMS",
+        "CASSINIactivityname": null,
```
rfrenchseti: Most of these fields should be derived from the obsname

opus/import/table_schemas/internal_def_product_types.json
```diff
@@ -1028,5 +1028,120 @@
...
+    },
+    {
+        "pi_dict_context": "OPUS_PRODUCT_TYPE",
+        "pi_dict_name": "coiss_f_ring_global_reproj_img_index",
+        "definition": "Global reprojected image index files (global_reproj_img_index.tab and .lblx) for Cassini ISS F ring reprojected images."
+    },
+    {
+        "pi_dict_context": "OPUS_PRODUCT_TYPE",
+        "pi_dict_name": "coiss_f_ring_reproj_img",
+        "definition": "Reprojected image files (*.img) and associated labels (*.lblx) for Cassini ISS F ring reprojected images. Labels also include definitions for reprojected image metadata tables and spice pointing, which can be downloaded separately."
```
rfrenchseti: Capitalize SPICE

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
rfrenchseti: I think it makes more sense to have `obs_profile_pds4.py` default everything to None and then only override them when we want to supply actual values. It made more sense for `obs_profile_pds4.py` to error previously, because all we had was occultation datasets. But now they will be in the minority. Does that make sense?

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,334 @@
...
+        return self._create_mult('SAT')
+
+    def _target_name(self):
+        return [('S RINGS', 'Saturn Rings')]
+
+    def field_obs_general_quantity(self):
+        return self._create_mult('REFLECT')
+
+    def field_obs_general_observation_type(self):
+        # Mosaic / bkg_sub tables: MOS; per-frame reprojected images: IMG
```
rfrenchseti: We don't have separate observations for reproj img so this comment and test don't need to be here.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,334 @@
...
+        raw = self._index_col('notes')
+        if raw is not None:
+            raw = raw.strip()
+
+        if not raw:
+            return None
+
+        raw_list = raw.split(';')
+        note_list = [NOTE_MAPPING[note] for note in raw_list]
+        return ' '.join(note_list)
```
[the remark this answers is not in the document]
rfrenchseti: It's true perhaps this should raise a more intelligent exception instead of just failing on a KeyError

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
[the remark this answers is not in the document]
rfrenchseti: These are the correct fields to use, even with `core_radius` in the index file.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,334 @@
...
+    def field_obs_ring_geometry_projected_long_resolution_angle1(self):
+        return self._index_col('rings:minimum_longitudinal_resolution')
+
+    def field_obs_ring_geometry_projected_long_resolution_angle2(self):
+        return self._index_col('rings:maximum_longitudinal_resolution')
+
+    # Equinox: 2009-08-11T01:40:08.914
+    # After this time, north side of the ring is lit.
+    # Before this time, south side of the ring is lit.
+    def _is_ring_north_side_lit(self):
```
rfrenchseti: Don't we have some variant of this elsewhere in the import pipeline? We should combine these into a common location.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,334 @@
...
+    def field_obs_mission_cassini_mission_phase_name(self):
+        return self._create_mult(self._cassini_normalize_mission_phase_name())
+
+
+    ###################################
+    ### OVERRIDE FROM ObsTypeImage ###
+    ###################################
+
+    def field_obs_type_image_image_type_id(self):
+        return self._create_mult('FRAM')
```
rfrenchseti: See earlier comment about making this a new Mosaic image type.

opus/import/obs_cassini_common_pds4.py
```diff
@@ -8,10 +8,114 @@
...
+    def field_obs_mission_cassini_spacecraft_clock_count1(self):
+        raw = self._index_col('cassini:spacecraft_clock_start_count')
+        if raw is None:
+            return None
+        try:
+            return opus_support.parse_cassini_sclk(str(raw).strip())
+        except Exception as e:
+            self._log_nonrepeating_error(
+                f'Unable to parse Cassini SCLK "{raw}": {e}')
+            return None
```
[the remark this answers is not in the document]
rfrenchseti: I don't think this is worth doing

scripts/server/import_and_deploy/_opus_import_volumes.sh
```diff
@@ -38,6 +38,7 @@ done
 for VOLUME in \
   EBROCC \
   uranus_occs_earthbased \
+# cassini_iss_fring_mosaics_rsfrench2025 (temporarily disabled)
```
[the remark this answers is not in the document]
rfrenchseti: Don't worry about this.

rfrenchseti (changes requested):

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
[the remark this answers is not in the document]
juzen2003: Option 2 used.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,316 @@
+################################################################################
+# obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
+#
+# Defines the ObsBundleCassiniISSFRingMosaicsRSFrench2025 class, which encapsulates
+# fields in the common, obs_mission_cassini, and obs_instrument_coiss tables for
+# the PDS4 bundleset "cassini_iss_fring_mosaics_rsfrench2025". This class
+# supports derived data from the cassini_iss_fring_mosaics_rsfrench2025 bundle.
+################################################################################
+
+from import_util import cached_tai_from_iso
```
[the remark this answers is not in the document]
juzen2003: Fixed.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,316 @@
...
+    'R': 'Follows one co-rotating longitude range with different inertial longitudes.',
+}
+
+class ObsBundleCassiniISSFRingMosaicsRSFrench2025(ObsCassiniCommonPDS4):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+
+    #######################
+    ### HELPER FUNTIONS ###
```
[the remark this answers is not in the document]
juzen2003: Fixed.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,316 @@
...
+    #######################
+
+    def _camera(self):
+        return self._index_col('min_image_name')[-1].upper()
+
+    def _circumference(self):
+        f_ring_core = 140221.3
+        return 3.14 * 2 * f_ring_core
```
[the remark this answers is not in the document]
juzen2003: Fixed.

opus/import/obs_bundle_cassini_uvis_solarocc_beckerjarmak2023.py
```diff
@@ -64,6 +62,42 @@ def field_obs_profile_temporal_sampling(self):
...
 
+    def field_obs_profile_occ_dir(self):
+        occ_dir = self._index_col('rings:ring_profile_direction')
+        if occ_dir is None:
+            occ_dir = self._index_col('rings:time_series_direction')
+        if occ_dir is None:
+            self._log_nonrepeating_error(
+                'rings:ring_profile_direction and rings:time_series_direction" '
+                'missing')
+            return None
```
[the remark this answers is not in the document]
juzen2003: Fixed.

opus/import/obs_bundle_uranus_occs_earthbased.py
```diff
@@ -133,7 +133,57 @@ def _target_name(self):
     ### OVERRIDE FROM ObsProfile ###
     ################################
 
-    # Occultation type is set as 'STE' in obs_profile_pds4
+    def field_obs_profile_occ_type(self):
+        occ_type = self._index_col('rings:occultation_type')
+        if occ_type == 'stellar':
+            return self._create_mult('STE')
+        self._log_nonrepeating_error(
+            f'Unknown rings:occultation:type "{occ_type}"')
```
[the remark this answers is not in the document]
juzen2003: Fixed.

opus/import/obs_bundle_uranus_occs_earthbased.py
```diff
@@ -133,7 +133,57 @@ def _target_name(self):
...
+        if occ_dir is None:
+            self._log_nonrepeating_error(
+                'rings:ring_profile_direction and rings:time_series_direction" '
+                'missing')
+            return None
+        occ_dir = occ_dir.upper()
+        if occ_dir in ('INGRESS', 'EGRESS', 'BOTH'):
+            return self._create_mult(occ_dir[0])
+        self._log_nonrepeating_error(f'Unknown profile direction "{occ_dir}"')
+        return self._create_mult(None)
```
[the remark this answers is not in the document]
juzen2003: Fixed.

opus/import/obs_bundle_uranus_occs_earthbased.py
```diff
@@ -133,7 +133,57 @@ def _target_name(self):
...
+        return ret
+
+    def field_obs_profile_wl_band(self):
+        wl_range = self._index_col('pds:wavelength_range')
+        if wl_range.upper() == 'INFRARED':
+            return self._create_mult('IR')
+        if wl_range.upper() == 'VISIBLE':
+            return self._create_mult('VI')
+        self._log_nonrepeating_error(f'Unknown pds:wavelength_range "{wl_range}"')
+        return None
```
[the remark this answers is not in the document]
juzen2003: Fixed.

opus/import/obs_cassini_common_pds4.py
```diff
@@ -8,10 +8,115 @@
...
         super().__init__(*args, **kwargs)
+
+
+    ##############################################################
+    ### OVERRIDE FOR obs_mission_cassini FROM ObsCassiniCommon ###
+    ##############################################################
+
+    def field_obs_mission_cassini_obs_name(self):
+        # Strip leading/trailing whitespace from the label value
+        return self._some_index_col('cassini:observation_id').strip()
```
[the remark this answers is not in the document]
juzen2003: Fixed.

opus/import/obs_cassini_common_pds4.py
```diff
@@ -8,10 +8,115 @@
...
+
+        filter1 = self._index_col('cassini:filter_name_1')
+        filter2 = self._index_col('cassini:filter_name_2')
+
+        new_filter = self._combined_filter(camera, filter1, filter2)
+        return self._create_mult_keep_case(new_filter)
+
+    def field_obs_instrument_coiss_camera(self):
+        camera = self._index_col('cassini:shutter_mode_id')[0]
+        assert camera in ('N', 'W')
```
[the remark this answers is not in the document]
juzen2003: Fixed.

opus/import/obs_cassini_common.py
```diff
@@ -109,12 +126,186 @@
...
     # HELPER FUNCTIONS USED BY CASSINI INSTRUMENTS
     ################################################################################
 
+    # Equinox: 2009-08-11T01:40:08.914
+    # After this time, north side of the ring is lit.
+    # Before this time, south side of the ring is lit.
+    def _is_ring_north_side_lit(self):
+        start_time = self.field_obs_general_time1()
+        equinox_time = cached_tai_from_iso('2009-08-11T01:40:08.914')
+        return start_time > equinox_time
```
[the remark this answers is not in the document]
juzen2003: Fixed.

opus/import/obs_cassini_common.py
```diff
@@ -198,6 +389,65 @@ def _cassini_normalize_mission_phase_name(self):
...
+            elif filter2.find('P') != -1:
+                new_filter = filter1 + '+' + filter2
+            else:
+                if (((wl1 is None or wl2 is None or wl1 == wl2) and
+                     filter1 > filter2) or
+                    wl1 > wl2):
+                    # Place filters in wavelength order
+                    # If wavelengths are the same, make it name order
+                    filter1, filter2 = filter2, filter1
+                new_filter = filter1 + '+' + filter2
```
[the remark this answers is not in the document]
juzen2003: Fixed.

juzen2003:
All comments addressed.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+    ######################################
+    ### OVERRIDE FROM ObsCassiniCommon ###
+    ######################################
+
+    def field_obs_mission_cassini_mission_phase_name(self):
+        return self._create_mult(self._cassini_normalize_mission_phase_name())
+
+
+    ###################################
+    ### OVERRIDE FROM ObsTypeImage ###
```
rfrenchseti, line 237: Fix comment style

opus/import/obs_bundle_uranus_occs_earthbased.py
```diff
@@ -133,7 +133,61 @@ def _target_name(self):
...
+        self._log_nonrepeating_error(f'Unknown profile direction "{occ_dir}"')
+        return self._create_mult(None)
+
+    def field_obs_profile_body_occ_flag(self):
+        return self._create_mult(self._index_col('rings:planetary_occultation_flag'))
+
+    def field_obs_profile_quality_score(self):
+        return self._create_mult(self._index_col('rings:data_quality_score'))
+
+    # TODO: Investigate further and fix if necessary. For the moment, assume BeckerJarmak optical
```
rfrenchseti: All of this BeckerJarmak stuff doesn't make sense here - this is the Uranus occs file

opus/import/obs_cassini_common.py
```diff
@@ -77,6 +77,23 @@
     'UP': 'UP (Upstream of the wake)'
 }
 
+# These mappings are for the TARGET_DESC field to clean them up
```
rfrenchseti, line 80: This stuff is *probably* PDS3-only. Add a TODO comment here to look at it when we transition to using PDS4 bundles for the main Cassini observations. Then we'd move it to a PDS3-specific class.

opus/import/obs_bundle_cassini_uvis_solarocc_beckerjarmak2023.py
```diff
@@ -64,6 +62,41 @@ def field_obs_profile_temporal_sampling(self):
...
+        self._log_nonrepeating_error(f'Unknown profile direction "{occ_dir}"')
+        return self._create_mult(None)
+
+    def field_obs_profile_body_occ_flag(self):
+        return self._create_mult(self._index_col('rings:planetary_occultation_flag'))
+
+    def field_obs_profile_quality_score(self):
+        return self._create_mult(self._index_col('rings:data_quality_score'))
+
+    # TODO: Investigate further and fix if necessary. For the moment, assume BeckerJarmak optical
```
rfrenchseti: Likewise remove the Uranus stuff from here

opus/import/obs_cassini_common_pds4.py
```diff
@@ -8,10 +8,135 @@
...
+    def field_obs_instrument_coiss_target_desc(self):
+        target_desc = self._index_col('cassini:pds3_target_desc')
+        if target_desc is not None:
+            target_desc = target_desc.upper()
+            coiss_target_desc_mapping = self._coiss_target_desc_mapping()
+            if target_desc in coiss_target_desc_mapping:
+                target_desc = coiss_target_desc_mapping[target_desc]
+        return self._create_mult(target_desc)
+
+    def _coiss_camera_letter_pds4(self):
```
rfrenchseti: This is a real hack. Maybe this is only needed for the Mosaics and should be moved to that file? For the real ISS bundles (that we aren't importing yet) we should get the camera from the LID.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+
+
+    #######################
+    ### HELPER FUNCTIONS ###
+    #######################
+
+    def _camera(self):
+        return self._index_col('min_image_name')[-1].upper()
+
+    def _circumference(self):
```
rfrenchseti: This should be a global constant:

```
_F_RING_CIRCUMFERENCE = 2 * math.pi * 140221.3
```

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+    def field_obs_ring_geometry_ascending_longitude1(self):
+        # minimum_longitude_ascending_node
+        return self._index_col('rings:minimum_inertial_ring_longitude')
+
+    def field_obs_ring_geometry_ascending_longitude2(self):
+        # maximum_longitude_ascending_node
+        return self._index_col('rings:maximum_inertial_ring_longitude')
+
+    # Phase angle: The angle between the point where incoming source photons
+    # hit the ring , to the direction where outgoing photons to the observer
```
rfrenchseti: Extra space before ,

Not sure why you're defining phase angle at all here, though.

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+
+    def field_obs_ring_geometry_j2000_longitude2(self):
+        if (self.field_obs_ring_geometry_ascending_longitude1() == 0 and
+            self.field_obs_ring_geometry_ascending_longitude2() == 360):
+            return 360
+        return self._ascending_to_j2000(
+            self.field_obs_ring_geometry_ascending_longitude2())
+
+    def field_obs_ring_geometry_ascending_longitude1(self):
+        # minimum_longitude_ascending_node
```
rfrenchseti: This comment and the one below seem redundant with the function name

opus/import/obs_bundle_cassini_iss_fring_mosaics_rsfrench2025.py
```diff
@@ -0,0 +1,325 @@
...
+    # hit the ring , to the direction where outgoing photons to the observer
+    def field_obs_ring_geometry_phase1(self):
+        return self._index_col('rings:minimum_phase_angle')
+
+    def field_obs_ring_geometry_phase2(self):
+        return self._index_col('rings:maximum_phase_angle')
+
+    # Source: star, observer: COISS
+    # Incidence angle: the angle between the point where incoming source photons
+    # hit the ring, to the north pole of the planet we're looking at (normal vector
```
rfrenchseti: The lit side is not the north side! It's the south pole when the lit side is the south side of the rings.

But not sure why any of these comments are here - these are just restating the definitions of these fields.

opus/import/obs_cassini_common_pds4.py
```diff
@@ -8,10 +8,135 @@
...
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+
+
+    ##############################################################
+    ### OVERRIDE FOR obs_mission_cassini FROM ObsCassiniCommon ###
+    ##############################################################
+
+    def field_obs_mission_cassini_obs_name(self):
+        # Strip leading/trailing whitespace from the label value
```
rfrenchseti, line 27: We shouldn't need to do this here. It should be taken care of where the table is read in the first place.

rfrenchseti (changes requested):

juzen2003:
All comments addressed.

opus/import/import_util.py
```diff
@@ -317,6 +317,11 @@ def safe_pdstable_read_pds3(filename):
 
     return rows, label
 
+def _strip_if_str(value):
```
rfrenchseti: So this is another workaround for a problem that shouldn't be here in the first place. The point is that when the index file is *read* there should not be any strings with trailing space in it. PdsTable does this correctly, but we are only using it for PDS3 index files. The issue is we are using just plain CSV reading for PDS4, but I'm giving you a fixed-width table, and the csv reading isn't stripping the fields. This will eventually be fixed properly when we use PdsTable to read PDS4 index files, but for now the correct fix is to change the *reading of the index file*.

opus/import/obs_bundle_cassini_uvis_solarocc_beckerjarmak2023.py
```diff
@@ -82,19 +82,12 @@ def field_obs_profile_body_occ_flag(self):
...
         return self._create_mult(self._index_col('rings:data_quality_score'))
 
-    # TODO: Investigate further and fix if necessary. For the moment, assume BeckerJarmak optical
-    # depth parameter can be interpreted the same as the Uranus Occs opacity parameter.
-    # Also, the -999 value in Becker/Jarmak data needs to be handled correctly.
     def field_obs_profile_optical_depth1(self):
-        ret = self._index_col('rings:lowest_detectable_opacity') # Uranus Occs
-        if ret is None:
-            ret = self._index_col('rings:lowest_detectable_normal_optical_depth') # BeckerJarmak. Note co-uvis-occ-2016-269-sun-i has -999.
+        ret = self._index_col('rings:lowest_detectable_normal_optical_depth')
```
rfrenchseti, line 87: So the fix is good but the comment about the -999 has gone away. But I think this is a very interesting comment. In the current OPUS the -999 is showing up in the search results, which is wrong. I presume when the new index file is created it will have -999 marked as a bad value and PdsTable will mask it out, but I'm going to create an issue about it so we don't forget it. But this is an example of paying attention to details in a way that AI will never do for you.

rfrenchseti (changes requested):

opus/import/obs_bundle_cassini_uvis_solarocc_beckerjarmak2023.py
juzen2003: Comments are put back

opus/import/import_util.py
juzen2003: .strip() is done in `safe_pdstable_read` function now

rfrenchseti (approved):

rfrenchseti (approved):

